PR #10, opened 2018-06-28
Implementation of HPACK encoding and header tables

AlanQuatermain:
Implementation of HPACK routines, including integer compression, header compression and indexed header tables, and huffman string compression.

Motivation:

This is the beginning of removing the dependency on `nghttp2`.

Modifications:

I've added a new library target, `NIOHPACK`, which contains all the necessary pieces to implement HPACK encoding as specified in RFC 7541. There are unit tests exercising all the code and some additional edge cases around integer encoding, along with some performance tests for the Huffman encode/decode process. The latter currently takes longer to decode than to encode, which is worrying, but this is only really noticeable when encoding > 16KB strings as a rule; therefore I've decided not to pull out the optimization hammer just yet. Included in the unit tests are the inputs and outputs from all the examples in RFC 7541 and RFC 7540.

I've tried to match the coding style from the repo, but some of my own idioms may have crept in here & there. Please let me know if you see anything wrong in that regard.

Currently the static header table and Huffman encoding tables are implemented as Swift `let`s. For the static header table that's ok (only 62 entries), but the Huffman table is huge. I'm debating whether it's worth having that in a plain C file somewhere, but I'm not sure what the tradeoffs are vs. performance. Right now it seems to only affect compilation time, but I may be wrong.

Result:

In general, nothing will change: nothing in NIOHTTP2 is using this library yet. That part is what I'm working on now.

PR #11, opened 2018-08-06
Correctly discard only inactive streams.

Lukasa:
Motivation:

Long-lived connections that have many inactive streams would cause
crashes. Crashes are, in my professional opinion, less than ideal
when they occur under expected usage cases.

Modifications:

- Fixed an inverted boolean check that would discard only *active*
  streams.
- Added tests.

Result:

Fewer crashes, better software.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2Parser.swift
+++ b/Sources/NIOHTTP2/HTTP2Parser.swift
@@ -156,16 +156,27 @@ public final class HTTP2Parser: ChannelInboundHandler, ChannelOutboundHandler {
     private let mode: ParserMode
     private let initialSettings: [HTTP2Setting]
     private let reentrancyManager = ReentrancyManager()
+    private let maxCachedClosedStreams: Int
 
-    public init(mode: ParserMode, initialSettings: [HTTP2Setting] = nioDefaultSettings) {
+    /// Create a `HTTP2Parser`.
+    ///
+    /// - parameters:
+    ///     - mode: The mode for the parser to operate in: server or client.
+    ///     - initialSettings: The initial settings used for the connection, to be sent in the
+    ///         connection preamble.
+    ///     - maxCachedClosedStreams: The maximum number of streams for which metadata will be preserved
+    ///         to handle delayed frames (e.g. DATA frames that were already in flight after stream reset
+    ///         or GOAWAY).
+    public init(mode: ParserMode, initialSettings: [HTTP2Setting] = nioDefaultSettings, maxCachedClosedStreams: Int = 1024) {
         self.mode = mode
         self.initialSettings = initialSettings
+        self.maxCachedClosedStreams = maxCachedClosedStreams
     }
 
     public func channelActive(ctx: ChannelHandlerContext) {
         self.session = NGHTTP2Session(mode: self.mode,
                                       allocator: ctx.channel.allocator,
-                                      maxCachedStreamIDs: 1024,  // TODO(cory): Make configurable
+                                      maxCachedStreamIDs: self.maxCachedClosedStreams,
                                       frameReceivedHandler: { ctx.fireChannelRead(self.wrapInboundOut($0)) },
                                       sendFunction: { ctx.write(self.wrapOutboundOut($0), promise: $1) },
                                       userEventFunction: { ctx.fireUserInboundEventTriggered($0) })
--- a/Sources/NIOHTTP2/NGHTTP2Session.swift
+++ b/Sources/NIOHTTP2/NGHTTP2Session.swift
@@ -285,7 +285,7 @@ fileprivate struct StreamManager {
     // Discard old and unnecessary streams.
     private mutating func purgeOldStreams() {
         while self.streamMap.count >= maxSize {
-            let lowestStreamID = self.streamMap.filter { $0.value.active }.keys.sorted().first { $0 != 0 && $0 != Int32.max }!
+            let lowestStreamID = self.streamMap.filter { !$0.value.active }.keys.sorted().first { $0 != 0 && $0 != Int32.max }!
             self.streamMap.removeValue(forKey: lowestStreamID)
         }
     }
--- a/Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
+++ b/Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
@@ -183,9 +183,9 @@ class SimpleClientServerTests: XCTestCase {
     }
 
     /// Establish a basic HTTP/2 connection.
-    func basicHTTP2Connection() throws {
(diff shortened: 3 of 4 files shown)


Sources/NIOHTTP2/HTTP2Parser.swift
```diff
@@ -156,16 +156,27 @@ public final class HTTP2Parser: ChannelInboundHandler, ChannelOutboundHandler {
...
         self.mode = mode
         self.initialSettings = initialSettings
+        self.maxCachedClosedStreams = maxCachedClosedStreams
     }
 
     public func channelActive(ctx: ChannelHandlerContext) {
         self.session = NGHTTP2Session(mode: self.mode,
                                       allocator: ctx.channel.allocator,
-                                      maxCachedStreamIDs: 1024,  // TODO(cory): Make configurable
+                                      maxCachedStreamIDs: self.maxCachedClosedStreams,  // TODO(cory): Make configurable
```
weissi: well, it's now configurable so the comment can go right? Or did you mean `ChannelOption` kind of configurable?

Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
```diff
@@ -1230,4 +1230,90 @@ class SimpleClientServerTests: XCTestCase {
         XCTAssertNoThrow(try self.clientChannel.finish())
         XCTAssertNoThrow(try self.serverChannel.finish())
     }
+
+    func testManyConcurrentInactiveStreams() throws {
+        let maxCachedClosedStreams = 1024
+
+        // Begin by getting the connection up.
+        try self.basicHTTP2Connection(maxCachedClosedStreams: 1024)
```
weissi: use the `maxCachedClosedStreams` variable here?

Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
```diff
@@ -1230,4 +1230,90 @@ class SimpleClientServerTests: XCTestCase {
...
+        XCTAssertNoThrow(try self.clientChannel.finish())
+        XCTAssertNoThrow(try self.serverChannel.finish())
+    }
+
+    func testDontRemoveActiveStreams() throws {
+        // This is a bit of a regression test, not a generally useful one.
+        let maxCachedClosedStreams = 1024
+
+        // Begin by getting the connection up.
+        try self.basicHTTP2Connection(maxCachedClosedStreams: 1024)
```
weissi: and here

Sources/NIOHTTP2/NGHTTP2Session.swift
```diff
@@ -285,7 +285,7 @@ fileprivate struct StreamManager {
     // Discard old and unnecessary streams.
     private mutating func purgeOldStreams() {
         while self.streamMap.count >= maxSize {
-            let lowestStreamID = self.streamMap.filter { $0.value.active }.keys.sorted().first { $0 != 0 && $0 != Int32.max }!
+            let lowestStreamID = self.streamMap.filter { !$0.value.active }.keys.sorted().first { $0 != 0 && $0 != Int32.max }!
```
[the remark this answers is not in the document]
MrMage: I assume the filtered array (and, by extension, `self.streamMap`) will now never be empty?

Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
Lukasa: Fixed.

Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
Lukasa: Fixed.

Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
```diff
@@ -1230,4 +1230,90 @@ class SimpleClientServerTests: XCTestCase {
         XCTAssertNoThrow(try self.clientChannel.finish())
         XCTAssertNoThrow(try self.serverChannel.finish())
     }
+
+    func testManyConcurrentInactiveStreams() throws {
+        let maxCachedClosedStreams = 1024
+
+        // Begin by getting the connection up.
+        try self.basicHTTP2Connection(maxCachedClosedStreams: maxCachedClosedStreams)
```
MrMage: I was about to mention the following (first part is already fixed): As you are declaring `maxCachedClosedStreams` above, should you replace `1024` with `maxCachedClosedStreams` here? Also, would it make sense to pick a lower value than 1024 for faster test runtime?

Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
```diff
@@ -1230,4 +1230,90 @@ class SimpleClientServerTests: XCTestCase {
...
+
+            let serverStreamID = try self.assertFramesRoundTrip(frames: [reqFrame, reqBodyFrame], sender: self.clientChannel, receiver: self.serverChannel).first!.streamID
+
+            // Let's send a quick response back.
+            var respFrame = HTTP2Frame(streamID: serverStreamID, payload: .headers(responseHeaders))
+            respFrame.endStream = true
+            try self.assertFramesRoundTrip(frames: [respFrame], sender: self.serverChannel, receiver: self.clientChannel)
+        }
+
+        // Ok, now we're going to open *two* streams. In the old, broken code, the opening of the second
```
MrMage: Mention this PR to more easily reference what the broken code was?

Sources/NIOHTTP2/NGHTTP2Session.swift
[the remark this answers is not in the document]
Lukasa: Not quite, but I wanted to separate the other issue out into another patch.

The TL;DR there is that if the sum of both sides SETTINGS_MAX_CONCURRENT_STREAMS is higher than `maxSize`, and there are actually more than `maxSides` valid open active streams, then we'll explode again. That can be fixed by changing the loop conditional above to move the `filter` out of the loop so that we only care if there are too many *closed* streams.

This will all get a bit better when we land #10 and the follow-on framing logic such that I can start work on replacing the nghttp2 state machine. At that stage we'll have a much more complete insight into the connection state and won't have to use this somewhat insane purging construction.

Lukasa:
No need to apologise @MrMage, more code review is always valuable. I've addressed both your comments.

Lukasa:
@weissi Want to re-review?

bmhatfield:
(I'm just a bystander, but) Bump: @weissi this PR fixes an issue discovered in https://github.com/grpc/grpc-swift/pull/281 - just adding a friendly bump for a re-review here :-)

PR #12, opened 2018-08-16
don't crash when client sends bad magic

weissi:
Motivation:

We shouldn't just crash if the client sends bad magic.

Changes:

Send an error through the pipeline instead of crashing on bad client
magic.

Result:

fewer crashes


Lukasa:
This is also missing the linux tests.

PR #16, opened 2018-08-16
More Swift-y types for HTTP/2 frame, flags, and settings.

AlanQuatermain:
Motivation:

Prior to bringing in my HTTP/2 encoder/decoder, I think my implementations of HTTP/2 frame flags as `OptionSet`s and settings as an enum (i.e. existentials) make for a nicer API, and provide somewhat more information for the compiler to work with when optimizing.

Modifications:

- `HTTP2Frame.FrameFlags` is now an `OptionSet`, and includes the CACHE_DIGEST `reset` and `complete` flag values.
- `HTTP2Frame.FramePayload` now supports all current & proposed HTTP/2 frame types, can return their on-wire identifier, and can return the set of `HTTP2Frame.FrameFlags` that the frame allows.
- `HTTP2Setting` is now an enum with associated values, and includes the SETTINGS_ACCEPT_CACHE_DIGEST and SETTINGS_ENABLE_CONNECT_PROTOCOL values.
- `HTTP2Setting` already includes code to encode/decode itself, though it's not yet used.
- Added `NIOHTTP2Errors.InvalidSettings` error, used when bad settings are encountered by decoder.
- NIOHTTP2 now depends on NIOHPACK.
- `HTTP2Stream` now maintains a `HPACKEncoder` and `HPACKDecoder`.
- Updated `NGHTTP2Session.swift` to handle (i.e. `fatalError("not implemented")`) the additional frame types CONTINUATION, BLOCKED, ORIGIN, and CACHE_DIGEST which were added to `HTTP2Frame.FramePayload`.
- Updated tests to use the new `HTTP2Setting` type.

Result:

Not a lot changes, beyond the in-memory layout of some things. All these types still have nghttp2 conversion methods, and all tests still run successfully.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -30,7 +30,7 @@ let package = Package(
         .target(name: "NIOHTTP2Server",
             dependencies: ["NIOHTTP2"]),
         .target(name: "NIOHTTP2",
-            dependencies: ["NIO", "NIOHTTP1", "NIOTLS", "CNIONghttp2", "NIOHPACK"]),
+            dependencies: ["NIO", "NIOHTTP1", "NIOTLS", "CNIONghttp2"]),
         .target(name: "NIOHPACK",
             dependencies: ["NIO", "NIOConcurrencyHelpers", "NIOHTTP1"]),
         .testTarget(name: "NIOHTTP2Tests",
--- a/Package.swift
+++ b/Package.swift
@@ -30,7 +30,7 @@ let package = Package(
         .target(name: "NIOHTTP2Server",
             dependencies: ["NIOHTTP2"]),
         .target(name: "NIOHTTP2",
-            dependencies: ["NIO", "NIOHTTP1", "NIOTLS", "CNIONghttp2"]),
+            dependencies: ["NIO", "NIOHTTP1", "NIOTLS", "CNIONghttp2", "NIOHPACK"]),
         .target(name: "NIOHPACK",
             dependencies: ["NIO", "NIOConcurrencyHelpers", "NIOHTTP1"]),
         .testTarget(name: "NIOHTTP2Tests",
--- a/Sources/NIOHTTP2/HTTP2Error.swift
+++ b/Sources/NIOHTTP2/HTTP2Error.swift
@@ -59,24 +59,6 @@ public enum NIOHTTP2Errors {
             self.nghttp2ErrorCode = nghttp2ErrorCode
         }
     }
-
-    /// Received/decoded data was invalid.
-    public struct InvalidSettings: NIOHTTP2Error {
-        /// The network identifier of the setting being read.
-        public var settingCode: UInt16
-        
-        /// The offending value.
-        public var value: Int32
-
-        /// The error code associated with the error.
-        public var errorCode: HTTP2ErrorCode
-
-        public init(setting: UInt16, value: Int32, errorCode: HTTP2ErrorCode) {
-            self.settingCode = setting
-            self.value = value
-            self.errorCode = errorCode
-        }
-    }
 }
 
 
--- a/Sources/NIOHTTP2/HTTP2Error.swift
+++ b/Sources/NIOHTTP2/HTTP2Error.swift
@@ -59,6 +59,24 @@ public enum NIOHTTP2Errors {
             self.nghttp2ErrorCode = nghttp2ErrorCode
         }
     }
+
+    /// Received/decoded data was invalid.
+    public struct InvalidSettings: NIOHTTP2Error {
+        /// The network identifier of the setting being read.
+        public var settingCode: UInt16
+        
+        /// The offending value.
+        public var value: Int32
+
+        /// The error code associated with the error.
+        public var errorCode: HTTP2ErrorCode
+
+        public init(setting: UInt16, value: Int32, errorCode: HTTP2ErrorCode) {
+            self.settingCode = setting
+            self.value = value
+            self.errorCode = errorCode
+        }
+    }
 }
 
 
--- a/Sources/NIOHTTP2/HTTP2Frame.swift
+++ b/Sources/NIOHTTP2/HTTP2Frame.swift
@@ -109,7 +109,7 @@ public struct HTTP2Frame {
         }
     }
     
-    public struct FrameFlags : OptionSet {
+    public struct FrameFlags: OptionSet {
         public typealias RawValue = UInt8
         
         public private(set) var rawValue: UInt8
--- a/Sources/NIOHTTP2/HTTP2Frame.swift
+++ b/Sources/NIOHTTP2/HTTP2Frame.swift
@@ -108,7 +108,7 @@ public struct HTTP2Frame {
         ///
         /// For more information on server push in HTTP/2, see
         /// [RFC 7540 § 8.2](https://httpwg.org/specs/rfc7540.html#rfc.section.8.2).
-        case pushPromise(HTTPHeaders)
+        case pushPromise
         
         /// A PING frame, used to measure round-trip time between endpoints.
         ///
--- a/Sources/NIOHTTP2/HTTP2Frame.swift
+++ b/Sources/NIOHTTP2/HTTP2Frame.swift
@@ -21,88 +21,45 @@ public struct HTTP2Frame {
     /// The payload of this HTTP/2 frame.
     public var payload: FramePayload
 
-    /// The frame flags as an 8-bit integer. To set/unset well-defined flags, consider using the
-    /// other properties on this object (e.g. `endStream`).
-    public var flags: UInt8
+    /// The frame flags.
+    public var flags: FrameFlags
 
     /// The frame stream ID as a 32-bit integer.
     public var streamID: HTTP2StreamID
 
+    private func _hasFlag(_ flag: FrameFlags) -> Bool {
+        return self.flags.contains(flag)
+            }
+    
+    private mutating func _setFlagIfValid(_ flag: FrameFlags) {
+        if self.payload.allowedFlags.contains(flag) {
+            self.flags.formUnion(flag)
+        }
+            }
+
     // Whether the END_STREAM flag bit is set.
     public var endStream: Bool {
-        get {
-            switch self.payload {
-            case .data, .headers:
-                return (self.flags & UInt8(NGHTTP2_FLAG_END_STREAM.rawValue) != 0)
-            default:
-                return false
-            }
+        get { return self._hasFlag(.endStream) }
+        set { self._setFlagIfValid(.endStream) }
         }
-        set {
-            switch self.payload {
-            case .data, .headers:
-                self.flags |= UInt8(NGHTTP2_FLAG_END_STREAM.rawValue)
-            default:
-                break
-            }
-        }
-    }
 
     // Whether the PADDED flag bit is set.
     public var padded: Bool {
-        get {
-            switch self.payload {
-            case .data, .headers, .pushPromise:
-                return (self.flags & UInt8(NGHTTP2_FLAG_PADDED.rawValue) != 0)
-            default:
-                return false
-            }
+        get { return self._hasFlag(.padded) }
+        set { self._setFlagIfValid(.padded) }
         }
-        set {
-            switch self.payload {
-            case .data, .headers, .pushPromise:
-                self.flags |= UInt8(NGHTTP2_FLAG_PADDED.rawValue)
-            default:
-                break
-            }
-        }
-    }
 
     // Whether the PRIORITY flag bit is set.
     public var priority: Bool {
-        get {
-            if case .headers = self.payload {
-                 return (self.flags & UInt8(NGHTTP2_FLAG_PRIORITY.rawValue) != 0)
-            } else {
-                return false
+        get { return self._hasFlag(.priority) }
+        set { self._setFlagIfValid(.priority) }
             }
-        }
-        set {
-            if case .headers = self.payload {
-                self.flags |= UInt8(NGHTTP2_FLAG_PRIORITY.rawValue)
-            }
-        }
-    }
 
     // Whether the ACK flag bit is set.
     public var ack: Bool {
-        get {
-            switch self.payload {
-            case .settings, .ping:
-                return (self.flags & UInt8(NGHTTP2_FLAG_ACK.rawValue) != 0)
-            default:
-                return false
+        get { return self._hasFlag(.ack) }
+        set { self._setFlagIfValid(.ack) }
             }
-        }
-        set {
-            switch self.payload {
-            case .settings, .ping:
-                self.flags |= UInt8(NGHTTP2_FLAG_ACK.rawValue)
-            default:
-                break
-            }
-        }
-    }
 
     public enum FramePayload {
         case data(IOData)
@@ -114,15 +71,86 @@ public struct HTTP2Frame {
         case ping(HTTP2PingData)
         case goAway(lastStreamID: HTTP2StreamID, errorCode: HTTP2ErrorCode, opaqueData: ByteBuffer?)
         case windowUpdate(windowSizeIncrement: Int)
-        case alternativeService
+        case continuation(HTTPHeaders)
+        case alternativeService(origin: String?, field: ByteBuffer?)
+        case blocked
+        case origin([String])
+        case cacheDigest(origin: String?, digest: ByteBuffer?)
+        
+        var code: UInt8 {
+            switch self {
+            case .data:                 return 0x0
+            case .headers:              return 0x1
+            case .priority:             return 0x2
+            case .rstStream:            return 0x3
+            case .settings:             return 0x4
+            case .pushPromise:          return 0x5
+            case .ping:                 return 0x6
+            case .goAway:               return 0x7
+            case .windowUpdate:         return 0x8
+            case .continuation:         return 0x9
+            case .alternativeService:   return 0xa
+            case .blocked:              return 0xb
+            case .origin:               return 0xc
+            case .cacheDigest:          return 0xd
+            }
+        }
+        
+        var allowedFlags: FrameFlags {
+            switch self {
+            case .data:
+                return [.padded, .endStream]
+            case .headers:
+                return [.endStream, .endHeaders, .padded, .priority]
+            case .pushPromise:
+                return [.endHeaders, .padded]
+            case .continuation:
+                return .endHeaders
+            case .cacheDigest:
+                return [.reset, .complete]
+                
+            case .settings, .ping:
+                return .ack
+                
+            case .priority, .rstStream, .goAway, .windowUpdate,
+                 .alternativeService, .blocked, .origin:
+                return []
+            }
+        }
+    }
+    
+    public struct FrameFlags : OptionSet {
+        public typealias RawValue = UInt8
+        
+        public private(set) var rawValue: UInt8
+        
+        public init(rawValue: UInt8) {
+            self.rawValue = rawValue
+        }
+        
+        public static let endStream     = FrameFlags(rawValue: 0x01)
+        public static let ack           = FrameFlags(rawValue: 0x01)
+        public static let reset         = FrameFlags(rawValue: 0x01)
+        public static let complete      = FrameFlags(rawValue: 0x02)
+        public static let endHeaders    = FrameFlags(rawValue: 0x04)
+        public static let padded        = FrameFlags(rawValue: 0x08)
+        public static let priority      = FrameFlags(rawValue: 0x20)
+        
+        // useful for test cases
+        internal static var allFlags: FrameFlags = [.endStream, .endHeaders, .padded, .priority]
     }
 }
 
 
 internal extension HTTP2Frame {
+    internal init(streamID: HTTP2StreamID, flags: HTTP2Frame.FrameFlags, payload: HTTP2Frame.FramePayload) {
+        self.streamID = streamID
+        self.flags = flags.intersection(payload.allowedFlags)
+        self.payload = payload
+    }
     internal init(streamID: HTTP2StreamID, flags: UInt8, payload: HTTP2Frame.FramePayload) {
         self.streamID = streamID
-        self.flags = flags
+        self.flags = FrameFlags(rawValue: flags).intersection(payload.allowedFlags)
         self.payload = payload
     }
 }
@@ -131,7 +159,7 @@ public extension HTTP2Frame {
     /// Constructs a frame header for a given stream ID. All flags are unset.
     public init(streamID: HTTP2StreamID, payload: HTTP2Frame.FramePayload) {
         self.streamID = streamID
-        self.flags = 0
+        self.flags = []
         self.payload = payload
     }
 }
--- a/Sources/NIOHTTP2/HTTP2Frame.swift
+++ b/Sources/NIOHTTP2/HTTP2Frame.swift
@@ -23,43 +23,43 @@ public struct HTTP2Frame {
 
     /// The frame flags.
     public var flags: FrameFlags
-
+    
     /// The frame stream ID as a 32-bit integer.
     public var streamID: HTTP2StreamID
-
+    
     private func _hasFlag(_ flag: FrameFlags) -> Bool {
         return self.flags.contains(flag)
-            }
+    }
     
     private mutating func _setFlagIfValid(_ flag: FrameFlags) {
         if self.payload.allowedFlags.contains(flag) {
             self.flags.formUnion(flag)
         }
-            }
-
+    }
+    
     // Whether the END_STREAM flag bit is set.
     public var endStream: Bool {
         get { return self._hasFlag(.endStream) }
         set { self._setFlagIfValid(.endStream) }
-        }
-
+    }
+    
     // Whether the PADDED flag bit is set.
     public var padded: Bool {
         get { return self._hasFlag(.padded) }
         set { self._setFlagIfValid(.padded) }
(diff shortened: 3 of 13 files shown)


Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -114,15 +71,86 @@ public struct HTTP2Frame {
...
+        public static let endStream     = FrameFlags(rawValue: 0x01)
+        public static let ack           = FrameFlags(rawValue: 0x01)
+        public static let reset         = FrameFlags(rawValue: 0x01)
+        public static let complete      = FrameFlags(rawValue: 0x02)
+        public static let endHeaders    = FrameFlags(rawValue: 0x04)
+        public static let padded        = FrameFlags(rawValue: 0x08)
+        public static let priority      = FrameFlags(rawValue: 0x20)
+        
+        // useful for test cases
+        internal static var allFlags: FrameFlags = [.endStream, .endHeaders, .padded, .priority]
```
ianpartridge: Should this contain `complete`?

AlanQuatermain: In theory, yes, but in the implementation it's the same value as `.ack`, so there's not really much value in including it. I could go either way, though.

AlanQuatermain:
I've more code to come focussed on the actual encode/decode process, but @Lukasa suggested I make smaller pull requests 😉 

In here I'm just altering some types to be more swift-y, since once the nghttp2 code is gone we'll be working with them directly much more often thus a nicer API is beneficial. Once this is approved I'll be able to start bringing in my encode/decode flows, part of which you can see in the `HTTP2Setting` implementation.

One issue I'm running into, though, has to do with the placement of the HPACK encoders/decoders in the `HTTPStream` type. It looks as though I'd need to take the `StreamManager` (or something like it) from `NGHTTP2Session.swift` and bring it into `HTTP2Parser.swift`, and implement the stream-location and decoding directly within that type. I'm not clear on the behavior of things like the ReentrancyManager though, for instance whether it would be necessary at all once we move away from NGHTTP2. I'm leaning towards renaming `HTTP2Parser` as `NGHTTP2Parser` and implementing a fresh one, with the ability to configure either of them into the channel handler stack. I'd appreciate some discussion around the best way to handle this ahead of time, to try and avoid some of the churn that occurred on #10.

Sources/NIOHTTP2/HTTP2Frame.swift
ianpartridge: I must be misunderstanding. `.ack` is `0x01` and `.complete` is `0x02`?

AlanQuatermain: Er. Yes. Er. Oops. Good catch 😅

AlanQuatermain: Le fixed 😊

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -114,15 +71,86 @@ public struct HTTP2Frame {
         case ping(HTTP2PingData)
         case goAway(lastStreamID: HTTP2StreamID, errorCode: HTTP2ErrorCode, opaqueData: ByteBuffer?)
         case windowUpdate(windowSizeIncrement: Int)
-        case alternativeService
+        case continuation(HTTPHeaders)
```
Lukasa: We don't want `continuation`.

CONTINUATION frames should be completely invisible from the perspective of this object. This is because they don't contain lists of headers, they just contain arbitrary bytes that are smashed together with the previous HEADERS frame. The mental model for this is that CONTINUATION exists literally and only to allow HEADERS to escape from the max frame size limit that applies to other frames.

In this case, we should rightly pretend that this does not exist.

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -114,15 +71,86 @@ public struct HTTP2Frame {
...
+        case alternativeService(origin: String?, field: ByteBuffer?)
+        case blocked
```
Lukasa: I don't think BLOCKED ever got specced.

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -114,15 +71,86 @@ public struct HTTP2Frame {
...
+                return .ack
+                
+            case .priority, .rstStream, .goAway, .windowUpdate,
+                 .alternativeService, .blocked, .origin:
+                return []
+            }
+        }
+    }
+    
+    public struct FrameFlags : OptionSet {
```
Lukasa: Nit: no whitespace before colon.

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -114,15 +71,86 @@ public struct HTTP2Frame {
         case ping(HTTP2PingData)
         case goAway(lastStreamID: HTTP2StreamID, errorCode: HTTP2ErrorCode, opaqueData: ByteBuffer?)
         case windowUpdate(windowSizeIncrement: Int)
-        case alternativeService
+        case continuation(HTTPHeaders)
+        case alternativeService(origin: String?, field: ByteBuffer?)
+        case blocked
+        case origin([String])
+        case cacheDigest(origin: String?, digest: ByteBuffer?)
```
Lukasa: For now let's not put in CACHE_DIGEST as it hasn't made it to a full RFC yet.

Sources/NIOHTTP2/HTTP2Settings.swift
```diff
@@ -12,101 +12,232 @@
...
+import NIO
 
-/// A HTTP/2 settings parameter that allows representing both known and unknown HTTP/2
-/// settings parameters.
-public struct HTTP2SettingsParameter {
-    internal let networkRepresentation: UInt16
-
-    /// Create a `HTTP2SettingsParameter` that is not known to NIO.
+/// A single setting for HTTP/2, a combination of parameter identifier and its value.
+public enum HTTP2Setting {
```
Lukasa: Let's not use an `enum` for this.

I'm already not wild about the use of an `enum` for `FramePayload`, but I don't really want to compound the mistake. A Swift `enum` is entirely non-extensible without performing a breaking release of the software. That means any new specced setting will force a brand new semver major release of this library. That's just not a good experience for anyone involved. It's better to leave this as a struct for now.

Incidentally, the struct may want to be implemented by *wrapping* a `private enum`, but that's a separate note.

Sources/NIOHTTP2/HTTP2Settings.swift
```diff
@@ -12,101 +12,232 @@
...
-    }
-
-    /// Corresponds to SETTINGS_HEADER_TABLE_SIZE
-    public static let headerTableSize = HTTP2SettingsParameter(1)
+    /// Allows the sender to inform the remote endpoint of the maximum size of
+    /// the header compression table used to decode header blocks, in octets. The
+    /// encoder can select any size equal to or less than this value by using
+    /// signaling specific to the header compression format inside a header block.
+    /// The initial value is 4,096 octets.
+    case headerTableSize(Int32)
```
Lukasa: Why are these fields `Int32`? They should be `Int` (for programmer convenience).

Sources/NIOHTTP2/HTTP2Settings.swift
```diff
@@ -12,101 +12,232 @@
...
-extension HTTP2SettingsParameter: Equatable {
-    public static func ==(lhs: HTTP2SettingsParameter, rhs: HTTP2SettingsParameter) -> Bool {
-        return lhs.networkRepresentation == rhs.networkRepresentation
+    /// SETTINGS_ENABLE_CONNECT_PROTOCOL (0x8)
+    ///
+    /// Upon receipt of this parameter with a value of `true`, a client MAY use
+    /// the Extended CONNECT method when creating new streams, for example to
+    /// bootstrap a WebSocket connection. Receipt of this parameter by a server
+    /// does not have any impact.
+    case enableConnectProtocol(Bool)
```
Lukasa: Let's not add settings without an RFC number yet.

Sources/NIOHTTP2/HTTP2Settings.swift
```diff
@@ -12,101 +12,232 @@
...
+            self.init(settings_id: 0x8, value: v ? 1 : 0)
+        }
     }
 }
 
-extension HTTP2Setting: Equatable {
-    public static func ==(lhs: HTTP2Setting, rhs: HTTP2Setting) -> Bool {
-        return lhs.parameter == rhs.parameter && lhs._value == rhs._value
+extension HTTP2Setting {
+    // nullable *and* throws? Invalid data causes an error, but unknown setting types return 'nil' quietly.
```
Lukasa: We should probably have an escape hatch for unknown settings types, rather than returning `nil` and throwing them away.

Sources/NIOHTTP2/HTTP2Stream.swift
```diff
@@ -108,6 +110,14 @@ final class HTTP2Stream {
...
     /// safe to prune.
     var active: Bool = false
+    
+    /// An HPACK decoder for this stream, handling integer and Huffman decoding and the management of a dynamic
+    /// header table.
+    var decoder: HPACKDecoder
+    
+    /// An HPACK encoder, handling integer and huffman encoding, header packing, and dynamic header table
+    /// management.
+    var encoder: HPACKEncoder
```
Lukasa: No need to add these here now, it's just a bunch of memory allocation that the user doesn't need yet.

Sources/NIOHTTP2/HTTP2Stream.swift
```diff
@@ -119,6 +129,8 @@ final class HTTP2Stream {
         self.outboundHeaderStateMachine = HTTP2HeadersStateMachine(mode: mode)
         self.streamID = streamID
         self.dataProvider = HTTP2DataProvider()
+        self.decoder = HPACKDecoder(allocator: ByteBufferAllocator())
+        self.encoder = HPACKEncoder(allocator: ByteBufferAllocator())
```
Lukasa: As a side note, these allocators should come from the `Channel`.

Lukasa:
> One issue I'm running into, though, has to do with the placement of the HPACK encoders/decoders in the HTTPStream type. It looks as though I'd need to take the StreamManager (or something like it) from NGHTTP2Session.swift and bring it into HTTP2Parser.swift, and implement the stream-location and decoding directly within that type. I'm not clear on the behavior of things like the ReentrancyManager though, for instance whether it would be necessary at all once we move away from NGHTTP2. I'm leaning towards renaming HTTP2Parser as NGHTTP2Parser and implementing a fresh one, with the ability to configure either of them into the channel handler stack. I'd appreciate some discussion around the best way to handle this ahead of time, to try and avoid some of the churn that occurred on #10.

Yeah, you shouldn't try to retrofit your HPACK encoders/decoders onto the `HTTP2Parser` as it exists today: that object is very definitely nghttp2 specific. Rather than rename it, create a new one with a temporary bad name (e.g. `SwiftHTTP2Parser` or `BetterHTTP2Parser`) and work on it alongside the current one: we can then re-use the testing for both before we eventually remove the nghttp2-based one and replace it with the Swift one.

Reentrancy manager exists to defend against the fact that nghttp2 uses callbacks to notify us of events. This is rarely a good idea in a parser because it leads to a number of tricky edge-cases:

1. You need to know whether the parser is safe to call re-entrantly. nghttp2 doesn't make this clear, so we assume it is not.
2. You need to ensure that if you call out to code you don't control from these callbacks, that you ensure you will never re-enter in an uncontrolled way.
3. You need to ensure that you don't screw up feeding data into the parser.

All of this is managed by reentrancy manager essentially turning all recursive re-entrant calls into `HTTP2Parser` into serial calls instead. For our new parser, we'll want to avoid this risk altogether by changing the parser to avoid using callbacks and instead using return values to trigger callouts. If we use that design, we do not need the re-entrancy manager.

Sources/NIOHTTP2/HTTP2Settings.swift
AlanQuatermain: Wow. 🤦‍♂️ That's… amazingly-not-forward-thinking design of the compiler there. As an API designer by trade, Swift's enums provide a good half of the elegance of Swift for API design. Having them designed in a way that makes them essentially unusable across library boundaries is just… wow. I literally cannot describe just how frustrating and disappointing that is; I read this message just as I was going to bed last night & I couldn't get to sleep for an hour afterwards. It's like "look at all these great things! Oh, you're not working on static binaries? Yeah, you can't use these, sorry."

This basically leaves us with the only (reasonable) option being the rather convoluted model used by `ChannelOption` in SwiftNIO; I guess now I know why it has that design. Well, I know how to do that, at least, so I'll put something together using that approach, because there *has* to be something better than 'magic numbers with other magic numbers in a struct'.

Sources/NIOHTTP2/HTTP2Settings.swift
AlanQuatermain: With the non-enum design, that won't be a problem any more: I'll have an 'unknown setting' type to return.

Sources/NIOHTTP2/HTTP2Stream.swift
AlanQuatermain: Ok.

Sources/NIOHTTP2/HTTP2Settings.swift
AlanQuatermain: This actual setting is the entire reason I'm working on HTTP/2 😉

I'm trying to get conforming reference implementations in place to nudge the RFC editor to actually look at the damned thing, since they've been sitting on it for a couple of months now with no action beyond a single question and answer right when it was first put into their queue. The same goes for much of the other items, too: most everything else on HTTP/2 the http-bis WG has completed & submitted to the RFC editors has basically sat dormant with no discussion while those of us who have real actual needs to use it now are holding our breath for it to be published. Hence muggins here trying to come up with a way for our iOS apps to tunnel a WebSocket through an HTTP/2 Adobe Traffic Server proxy to a back-end running a Netty or Akka HTTP/2 server: literally the one use-case that stops us enabling HTTP/2 across the organization. Sigh.

Ultimately I want my branch to be somewhere that I can put together experimental implementations of in-progress standards, to better provide feedback to the IETF & co. WebSocket bootstrapping, origin, and cache digests, however, are things I need to provide an implementation for asap, though, and actually form the primary focus of my work on the codecs and framers. Can you suggest a good way to handle this? Maintaining two parallel branches with this much churn on the implementation details is a real pain, but I guess that's my only option?

Sources/NIOHTTP2/HTTP2Settings.swift
ianpartridge: Agreed. Take a look at https://github.com/apple/swift-evolution/blob/master/proposals/0192-non-exhaustive-enums.md which aims to solve this for enums in the stdlib and overlays, but not for enums in libraries like SwiftNIO.

AlanQuatermain: Oh FFS, if they only care about fixing it for stdlib & Apple's own frameworks that's just adding insult to injury. Not to mention it severely reducing the usefulness of the Swift 5.0 stable ABI from an third-party library perspective.

Sources/NIOHTTP2/HTTP2Settings.swift
AlanQuatermain: Hm, actually, it has an RFC number assigned: RFC8441. It's just waiting on IANA registration of the "websocket" value for the `:protocol` header: https://www.rfc-editor.org/auth48/rfc8441

Sources/NIOHTTP2/HTTP2Settings.swift
AlanQuatermain: @Lukasa Given the problems with enums, would it not make more sense to define frame payloads as `protocol`/`struct` types? For example, see https://github.com/AlanQuatermain/H2Swift/blob/1e389199710ace9f6f29795892cbdd0f8dc37232/Sources/H2Swift/Frames/Frame.swift#L193 and https://github.com/AlanQuatermain/H2Swift/blob/master/Sources/H2Swift/Frames/DataFrame.swift

AlanQuatermain: …for the payload types themselves, at least.

Lukasa: For frame payloads I basically propose we leave it alone for now as it's already there, and we take a separate spin at re-designing them later in the process. They should be easily-enough redesigned in isolation and it'll help focus the mind.

Sources/NIOHTTP2/HTTP2Settings.swift
Lukasa: This may be a good time to think about what, if anything, the extensibility model should be here. We'd really like to be able to support extensions without requiring that core NIO adopt them directly. In the short term I'm ok with putting this in, so long as we're confident that the various IANA-assigned code points aren't going to change. Given that this has an RFC number we're probably safe.

Package.swift
```diff
@@ -30,7 +30,7 @@ let package = Package(
         .target(name: "NIOHTTP2Server",
             dependencies: ["NIOHTTP2"]),
         .target(name: "NIOHTTP2",
-            dependencies: ["NIO", "NIOHTTP1", "NIOTLS", "CNIONghttp2"]),
+            dependencies: ["NIO", "NIOHTTP1", "NIOTLS", "CNIONghttp2", "NIOHPACK"]),
```
Lukasa: We can also probably leave this off for now. Good enthusiasm, but let's not make SwiftPM compile this for no reason.

Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -59,6 +59,24 @@ public enum NIOHTTP2Errors {
             self.nghttp2ErrorCode = nghttp2ErrorCode
         }
     }
+
+    /// Received/decoded data was invalid.
+    public struct InvalidSettings: NIOHTTP2Error {
```
Lukasa: I don't think this struct is actually used anywhere, is it?

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -114,15 +71,74 @@ public struct HTTP2Frame {
         case ping(HTTP2PingData)
         case goAway(lastStreamID: HTTP2StreamID, errorCode: HTTP2ErrorCode, opaqueData: ByteBuffer?)
         case windowUpdate(windowSizeIncrement: Int)
-        case alternativeService
+        case alternativeService(origin: String?, field: ByteBuffer?)
+        case origin([String])
+        
+        var code: UInt8 {
```
Lukasa: Can we get a doc comment here?

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -114,15 +71,74 @@ public struct HTTP2Frame {
...
+            case .pushPromise:          return 0x5
+            case .ping:                 return 0x6
+            case .goAway:               return 0x7
+            case .windowUpdate:         return 0x8
+            case .alternativeService:   return 0xa
+            case .origin:               return 0xc
+            }
+        }
+        
+        var allowedFlags: FrameFlags {
```
Lukasa: Can we get a doc comment here too?

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -114,15 +71,74 @@ public struct HTTP2Frame {
...
+            case .data:
+                return [.padded, .endStream]
+            case .headers:
+                return [.endStream, .endHeaders, .padded, .priority]
+            case .pushPromise:
+                return [.endHeaders, .padded]
+                
+            case .settings, .ping:
+                return .ack
+
```
Lukasa: There's some surprise whitespace here: mind cleaning it up?

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -114,15 +71,74 @@ public struct HTTP2Frame {
...
+                return .ack
+                
+            case .priority, .rstStream, .goAway, .windowUpdate,
+                 .alternativeService, .origin:
+                return []
+            }
+        }
+    }
+    
+    public struct FrameFlags: OptionSet {
```
Lukasa: Doc comment here too please.

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -114,15 +71,74 @@ public struct HTTP2Frame {
...
+        
+        public init(rawValue: UInt8) {
+            self.rawValue = rawValue
+        }
+        
+        public static let endStream     = FrameFlags(rawValue: 0x01)
+        public static let ack           = FrameFlags(rawValue: 0x01)
+        public static let endHeaders    = FrameFlags(rawValue: 0x04)
+        public static let padded        = FrameFlags(rawValue: 0x08)
+        public static let priority      = FrameFlags(rawValue: 0x20)
```
Lukasa: Doc comments on these too if you please.

Sources/NIOHTTP2/HTTP2Settings.swift
Lukasa: Actually designing the extension point is a bit tricky, because we have a performance/flexibility trade-off that has to be made.

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -21,108 +21,208 @@ public struct HTTP2Frame {
...
-                return false
-            }
-        }
-        set {
-            if case .headers = self.payload {
-                self.flags |= UInt8(NGHTTP2_FLAG_PRIORITY.rawValue)
-            }
-        }
+        get { return self._hasFlag(.priority) }
+        set { self._setFlagIfValid(.priority) }
```
Lukasa: While we're here, let's ask the question: now that we're using an `OptionSet` for flags, are these computed properties special enough to stay? @AlanQuatermain, what do you think?

Sources/NIOHTTP2/NGHTTP2Session.swift
```diff
@@ -457,7 +457,7 @@ class NGHTTP2Session {
             }
             nioFramePayload = .settings(settings)
         case NGHTTP2_PUSH_PROMISE.rawValue:
-            nioFramePayload = .pushPromise
+            nioFramePayload = .pushPromise(self.headersAccumulation)
```
Lukasa: This seems unrelated.

Sources/NIOHTTP2/HTTP2Frame.swift
AlanQuatermain: There's also the question of what to do if the user wants to set a flag that's invalid for the payload type being used. Silently ignore it? Raise an error? Allow it, but ignore at encode time? Ignore, but raise error when encoding? I've generally gone in the direction of silently ignoring the modification when it's made, in my own implementation.

AlanQuatermain: Hm, that doesn't work with `frame.flags.formUnion(someInvalidFlag)` though. It looks like we might have use for an internal `validateFlags()` method or a `validatedFlags` property that would return the intersection of the set flags and those that are actually allowed on the wire. 

Additionally, allowing any flag value to be set means that a client might request padding by setting that flag. I'm in favor of supporting pad bytes when encoding because it's not difficult, but would have it only be used if the client set that flag itself. We wouldn't set it ourselves unless we find a particular use case in testing that would really benefit from it, but really padding seems to be for the benefit of the recipient—our encoding and in-memory representation doesn't cry out for nicely aligned frame sizes. The question, then, is whether the other end does, or whether the intermediary hops would do so. *\*shrug\**

Lukasa: Padding is super weird, it's hard to deploy correctly. We should support serialising it when asked for, but let's not give it a special place in the API: I don't want to *encourage* it. ;)

Otherwise, I'm fine with adding a `didSet` that silently strips flags that have no meaning for this frame, or alternatively adding one that traps if you set a bad one.

Lukasa: Oh, whoops, I also forgot to say: let's burn the helpers to the ground.

Oh, and actually let's just ignore bad flags, rather than stripping them or trapping. That gives a logical place for extensions to hook in without us having a really rough time, at least early on: they can just set random flag bits if they want to.

AlanQuatermain: Yeah, I don't want to encourage use of padding, since it's of limited value to us. That said, if someone wants to go setting that flag, we should honour it.

Lukasa: This is a post-hoc comment from me, but I think I disagree with my past self:

> Oh, and actually let's just ignore bad flags, rather than stripping them or trapping. That gives a logical place for extensions to hook in without us having a really rough time, at least early on: they can just set random flag bits if they want to.

I think this was a partially thought-out idea. While it's nice for extension hooking, it leads to a substantially less pleasant programming environment for everyone else. Really, the flags are logically part of the payload of a given frame, and we should treat them as such.

This does raise questions about how we're going to handle extensions, but right now we don't have a good extensibility answer and I don't think it's going to land in the 1.0 release (simply no time to solve that issue). So I think we should consider extensions an explicit goal of the 2.0 release, and in the short term build the best API for non-extensible use-cases we can.

This is mostly here as a historical note and to allow @AlanQuatermain to shout me down if he wants ;).

PR #18, opened 2018-09-03
Removes wrong return type documentation

t089:
The documentation previously stated that `createStreamChannel` returns an "`EventLoopFuture` that completes with the new channel", where in reality the method returns `Void` and you instead supply an `EventLoopPromise` that is succeeded with the new activated channel. This PR removes the `returns:` line from the documentation comment.


normanmaurer:

weissi:
thanks @t089 !

PR #23, opened 2018-11-02
Fix tests on nghttp2-1.34.0

kevints:
Motivation:

nghttp2 changed the error code it returns when the server sends a GOAWAY.
It's now `.cancelled` instead of `.refusedStream`.

Modifications:

Don't require a specific error code in the tests, just require that an
error matching the expected stream ID occurred.

Result:

Tests will pass on nghttp2-1.34.0.

### DIFF
--- a/Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
+++ b/Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
@@ -966,8 +966,8 @@ class SimpleClientServerTests: XCTestCase {
         self.interactInMemory(self.clientChannel, self.serverChannel)
         try self.clientChannel.assertReceivedFrame().assertGoAwayFrame(lastStreamID: 0, errorCode: UInt32(http2ErrorCode: HTTP2ErrorCode.inadequateSecurity), opaqueData: nil)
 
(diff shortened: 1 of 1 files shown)


Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
```diff
@@ -1170,7 +1170,7 @@ class SimpleClientServerTests: XCTestCase {
         // Now the streams are closed, they should have seen user events.
         XCTAssertEqual(clientHandler.events.count, 1)
         XCTAssertEqual(serverHandler.events.count, 1)
-        XCTAssertEqual(clientHandler.events[0] as? StreamClosedEvent, StreamClosedEvent(streamID: clientStreamID, reason: .refusedStream))
+        XCTAssertEqual((clientHandler.events[0] as? StreamClosedEvent)?.streamID, clientStreamID)
         XCTAssertEqual(serverHandler.events[0] as? StreamClosedEvent, StreamClosedEvent(streamID: serverStreamID, reason: .refusedStream))
```
Lukasa: I think this one needs to be amended as well.

Lukasa:

Lukasa:
Thanks @kevints!

PR #25, opened 2018-11-16
Trailers sent on idle streams deserve to be sent too.

### LINKED ISSUE #24: Trailers sent on streams without pending data never get emitted
If a stream has had all its data frames emitted, such that the HTTP2StreamDataProvider is in the `.pending` state, and the user then submits a trailer frame, that frame will never be emitted. This is because we don't correctly resume the data provider in this case.

Lukasa:
Motivation:

Due to nghttp2's weird data provider model, all trailers go through a
similar send path as data frames. This includes the fact that the data
sender can go "idle". If the data sender is idle, and the user sends
trailers, we don't bother to tell nghttp2 to ask us for more data,
so the trailers sit there forever, feeling sad.

Modifications:

- Correctly restart the data provider when sending data.

Result:

Trailers will be emitted when the stream is idle.
Resolves #24.

### DIFF
--- a/Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
+++ b/Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
@@ -1006,6 +1006,41 @@ class SimpleClientServerTests: XCTestCase {
         XCTAssertNoThrow(try self.serverChannel.finish())
     }
 
+    func testSendingTrailersWhenDataProviderIsIdle() throws {
+        // Begin by getting the connection up.
+        try self.basicHTTP2Connection()
+
+        // We're now going to try to send a request from the client to the server. This request will be one headers frame, one
+        // data frame, and then another headers frame for trailers. The headers frame for trailers will be sent late, to
+        // reproduce https://github.com/apple/swift-nio-http2/issues/24.
+        let headers = HTTPHeaders([(":path", "/"), (":method", "POST"), (":scheme", "https"), (":authority", "localhost")])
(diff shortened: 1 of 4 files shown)


Lukasa:
Note: this issue blocks grpc/grpc-swift#281.

Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
```diff
@@ -1006,6 +1006,41 @@ class SimpleClientServerTests: XCTestCase {
...
+        let clientStreamID = HTTP2StreamID()
+        let reqFrame = HTTP2Frame(streamID: clientStreamID, payload: .headers(headers))
+        let reqBodyFrame = HTTP2Frame(streamID: clientStreamID, payload: .data(.byteBuffer(requestBody)))
+        var trailerFrame = HTTP2Frame(streamID: clientStreamID, payload: .headers(trailers))
+        trailerFrame.flags.insert(.endStream)
+
+        let serverStreamID = try self.assertFramesRoundTrip(frames: [reqFrame, reqBodyFrame], sender: self.clientChannel, receiver: self.serverChannel).first!.streamID
+
+        // Now we can send the next trailers.
+        try self.assertFramesRoundTrip(frames: [trailerFrame], sender: self.clientChannel, receiver: self.serverChannel)
```
normanmaurer: XCTAssertNoThrow ?

MrMage:
I can confirm that this fixes the issue I had been encountering. Thank you so much, Corey!

Lukasa:
I've updated to use `XCTAssertNoThrow`.

PR #28, opened 2018-11-26
Pure-Swift protocol implementation.

Lukasa:
This PR adds a pure-Swift implementation of the HTTP/2 protocol state machine.

The goal here is to get the core of the protocol logic codified. This, combined with @AlanQuatermain's previous HPACK work, and with a future requirement for a frame parser, would give us a complete HTTP/2 implementation entirely in Swift.

This PR is unfortunately very large, as HTTP/2 is a fairly complex protocol. However, it contains roughly the following:

- A stream state machine, which validates that frames may be sent or received
- A connection state machine, which validates that frames may be sent or received, and which owns the rest of the state of the connection
- SETTINGS state management
- Flow control management
- Header validation

### Design

The design tentpoles here are as follows:

1. Make impossible states impossible (or at least very hard) to represent.

    Where it is possible for us to store state *in* the state machine, rather than on it, we do so. This is taken to its logical extreme in the `ConnectionStateMachine`, which has only one stored variable: the state `enum`.

    The result of this change is that states that are associated with missing information (e.g. no settings emitted) literally do not store that information, making it impossible to query about it or accidentally rely on its presence.

2. Fatal errors are only the result of programmer error

    As this is an implementation that processes untrusted bytes, it is important to ensure that it handles protocol errors gracefully. In this case, we do so by ensuring that the only `assert` or `precondition` statements in the code are enforcing programmer error, rather than making any assertions about the behaviour of the protocol itself.

    Our return codes make it easy to handle all errors, and store all the relevant information required to process them.

3. Minimise allocator utilization

    One of the most costly things we can do in this kind of protocol implementation is allocate memory. Almost all computation is cheap in comparison to the cost of allocating memory. For this reason, code here is written defensively, such that the common path requires little-or-no allocation.

    One of the consequences is that we avoid classes and CoW value types wherever possible. It is naturally not possible to entirely avoid them, but this implementation reduces the number of them to manageable amounts (a connection at its busiest stores 3 dictionaries and 2 arrays, and contains no classes). All of our CoW value types have best-guess sizes provided for the common-case, to minimise the amount of delayed allocation that the common path will have to perform.

4. Connection state has value semantics

    Logically, the state of a single HTTP/2 connection is a value. While the connection *itself* has reference semantics, this corresponds to the relevant `ChannelHandler`, not the state itself. The state should be conceptually immutable and copyable.

    This provides very useful properties. It makes diagnostics easier, as mutation is carefully guarded via Swift's exclusion rules. It forces us to be extremely carefully about "outcalls" from the state machine, as they will cause runtime crashes in Swift. An interesting line of investigation is that it opens the door to having a fairly trivial completely serialisable representation of the state of a HTTP/2 connection that could be used to entirely regenerate that state. This is a useful design property to investigate here, as it allows for a number of interesting behaviours down the road, as well as providing a powerful tool for protocols like QUIC where the network 5-tuple does not uniquely identify the connection (i.e. connections are moveable).

5. No I/O in the state machine.

As much as possible the state machine should know *nothing* about how to handle I/O. This allows us to divorce the state machine from the ChannelHandler abstraction that we'll use down the road. It helps us make testing vastly easier, as well as to make it easier to handle evolutions in the ChannelHandler API that may come down the road.

### Omissions

The major omission from this PR is any support for handling protocol extensions. For the moment that was judged a step too far in complexity. However, this implementation should allow easy evolution to handle protocol extensions: the vast majority of what is needed for protocol extensions has been built but kept internal or private to allow prototyping internally.


Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
```diff
@@ -0,0 +1,1283 @@
...
+        case .closed:
+            return .connectionError(underlyingError: NIOHTTP2Errors.IOOnClosedConnection(), type: .protocolError)
+        }
+    }
+
+    /// Called when a PING frame has been received from the network.
+    mutating func receivePing() -> StateMachineResult {
+        // Pings are pretty straightforward: they're basically always allowed. This is a bit weird, but I can find no text in
+        // RFC 7540 that says that receiving PINGs with ACK flags set when no PING ACKs are expected is forbidden. This is
+        // very strange, but we allow it.
```
ktoso: Not sure I understand your concern correctly, but: as far as I remember a PING with ACK https://tools.ietf.org/html/rfc7540#section-6.7 can be seen kind of like a "PONG" in WebSockets, where peers may send those at-will, uni-directional, not expecting anything back, though perhaps utilizing this as a form of heartbeat - related section in WS: https://tools.ietf.org/html/rfc6455#section-5.5.3 I assumed the receiving and ignoring of PING+ACK at any time would serve a similar purpose here.

Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
```diff
@@ -0,0 +1,1307 @@
...
+        /// This is a weird state, unlikely to be encountered in most programs, but it's technically possible.
+        case quiescingPrefaceSent(QuiescingPrefaceSentState)
+
+        /// We have received a preface, and received a GOAWAY, but we haven't sent our preface yet.
+        /// This is a weird state, unlikely to be encountered in most programs, but it's technically possible.
+        case quiescingPrefaceReceived(QuiescingPrefaceReceivedState)
+
+        /// The connection has completed, either cleanly or with an error. In this state, no further activity may
+        /// occur on the connection.
+        case closed
```
weissi: I think we should rename this because the connection isn't actually closed. Maybe `fullyQuiesced` or `becameUnusable` or something

Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
```diff
@@ -0,0 +1,1307 @@
...
+    }
+}
+
+// MARK:- State modifying methods
+//
+// These methods form the implementation of the public API of the HTTP2ConnectionStateMachine. Each of these methods
+// performs a state transition, and can be used to validate that a specific action is acceptable on a connection in this state.
+extension HTTP2ConnectionStateMachine {
+    /// Called when a SETTINGS frame has been received from the remote peer
+    mutating func receiveSettings(settings: HTTP2Settings, flags: HTTP2Frame.FrameFlags) -> StateMachineResult {
```
weissi: double word 'settings', `receiveSettings(_ settings, ...)`

Lukasa:
@MrMage That's absolutely a thing we're going to do. However, this PR does not provide a functional change yet: it's the second of three major tentpoles we need to build. Once we do that we'll be testing with our known implementations to validate that everything is still working nicely.

Lukasa:
Ok, this patch is updated and squashed into a single commit. I believe it's now ready for review, though I appreciate that it's so massive that reviewing it effectively may be hard.

There are some things missing from this patch, covered by TODOs in the code. Those will be addressed after the SwiftNIO 1.12 release, which will give us the ability to get high-performance access to the `HTTPHeaders` struct to implement header validation.

Next up is frame parsing. @AlanQuatermain do you have work done here already, or should I go ahead and charge on with that?

AlanQuatermain:
I’m in progress right now, and hope to have the initial implementation done tomorrow. There will likely be a secondary pass to work on integration between our two parts, but we can cross that bridge once it’s built.

Sources/NIOHTTP2/ConnectionStateMachine/FrameSendingStates/SendingPushPromiseState.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    var mayPush: Bool { get }
+}
+
+extension SendingPushPromiseState {
+    mutating func sendPushPromise(originalStreamID: HTTP2StreamID, childStreamID: HTTP2StreamID, headers: HTTPHeaders) -> StateMachineResult {
+        return self._sendPushPromise(originalStreamID: originalStreamID, childStreamID: childStreamID, headers: headers)
+    }
+
+    fileprivate mutating func _sendPushPromise(originalStreamID: HTTP2StreamID, childStreamID: HTTP2StreamID, headers: HTTPHeaders) -> StateMachineResult {
```
weissi: what makes this function and the one above (which calls it) different?

Lukasa: Nothing, this is just a holdover from when I had an implementation in the override below. Will compress.

weissi:
@Lukasa btw

```
17:15:17 Compile Swift Module 'NIOHPACKTests' (12 sources)
17:15:17 /code/Sources/NIOHTTP2/HTTP2FlowControlWindow.swift:162:1: error: implementation of 'Equatable' cannot be automatically synthesized in an extension
17:15:17 extension HTTP2FlowControlWindow: Equatable { }
17:15:17 ^
17:15:17 /code/Sources/NIOHTTP2/HTTP2FlowControlWindow.swift:164:1: error: implementation of 'Hashable' cannot be automatically synthesized in an extension
17:15:17 extension HTTP2FlowControlWindow: Hashable { }
17:15:17 ^
17:15:17 error: terminated(1): /root/.swift/usr/bin/swift-build-tool -f /code/.build/debug.yaml test output:
```

Swift 4.1 ...

Lukasa:
Yeah. Do we care? We could either require Swift 4.2 or Swift 5.

Sources/NIOHTTP2/ConnectionStateMachine/FrameSendingStates/SendingPushPromiseState.swift
Lukasa: Addressed.

weissi:
> Yeah. Do we care? We could either require Swift 4.2 or Swift 5.

Good point, let's make it 4.2 for now, we can always (before we'll tag 1.0.0) up to 5. But we should get 4.2 & 5 CIs spun up.

Lukasa:
Cool. @tomerd are you open to doing this work?

Lukasa:
Included in #40.

PR #29, opened 2018-12-04
Add support for Ubuntu 16.04's packaged nghttp2.

Lukasa:
Motivation:

We'd done all our development against newer copies of nghttp2, that
behaved a bit differently. We should support Ubuntu's copy, at least
while nghttp2 is relevant to us.

Modifications:

- Shimmed over a version difference.
- Fixed the tests to stop relying on differences.
- Removed the Swift 4.0.3 dockerfiles as we don't support Swift
  4.0 anyway.

Result:

Users can build against Ubuntu 16.04's nghttp2 package.


docker/docker-compose.ubuntu-1604.403.yaml
```diff
@@ -1,14 +0,0 @@
...
-
-services:
-
-  runtime-setup:
-    image: swift-nio-http2:ubuntu-16.04-4.0.3
-    build:
-      args:
-        ubuntu_version : "16.04"
-        swift_version : "4.0.3"
-        install_nghttp2_from_source: "true"
```
weissi: why are we removing the whole docker yamls here?

weissi: ah, we don't support 4.0 :)

Lukasa: Because we don't support Swift 4.0, and never have. This is mentioned in the commit message.

PR #32, opened 2018-12-10
Pure-Swift HTTP2 frame parsers

AlanQuatermain:
Motivation:

Part of the work to remove nghttp2 dependency.

Modifications:

Implemented frame encoder and decoder types to provide serialization/deserialization:

* Encoder/Decoder types are distinct reference types, designed to be picked up & used by stream channels or similar.
* Uses NIOHPACK to implement header encoding/decoding. Currently always encodes using Huffman coding, though that could change once we have a `Channel` (and thus `ChannelOptions`) story here.
* Properly decodes padding bytes and flags, and throws errors where padding is broken (e.g. padding size larger than remaining payload size). Currently ignores `PADDED` flag when encoding frames, and never emits padding. If we want this supported via `ChannelOptions` then it can be added later.
* Throws errors from decoder only for data-level issues, like frames being mangled, wrong frame size. Also throws if a stream-specific frame specifies the root (connection) stream or vice-versa.
* Transparently handles incoming `CONTINUATION` frames. When a `HEADERS` frame is decoded which does not have the `END_HEADERS` flag set, that frame is cached in the decoder. Following `CONTINUATION` frames have their headers decoded and appended to those in the cached headers frame. Once we see an `END_HEADERS` flag, we return the complete headers frame as our result.
  - Note that we will throw an error if we see a `CONTINUATION` frame which was not preceded by either a `HEADERS` or a `PUSH_PROMISE` frame.

Some stop-gap temporary items which will change when this is merged with @Lukasa's changes:

* `HPACKEncoder`/`HPACKDecoder` types are intended to belong to a HTTP2 Stream. Since that implementation is all changing in @Lukasa's parallel set of changes, I've dropped in a simple lookup table here for now. The intent is NOT for this to go live; it exists for the benefit of the unit tests and to stub out the places in the implementation which would need to reach out & get these HPACK coders from some stream-specific source.
* Custom `NIOHTTP2Errors.ConnectionError` type holding a `NIOHTTP2ErrorCode`. @Lukasa's implementation in #28 is preferred, but I needed a placeholder for this implementation.

Some frame payload types have changed to accommodate the manual parser:

* `headers` now contains an `HPACKHeaders` and an optional `StreamPriorityData` rather than a single `HTTPHeaders`.
* `priority` now contains a `StreamPriorityData`.
* `pushPromise` now contains a `HTTP2StreamID` and an `HPACKHeaders`.

The new `HTTP2Frame.StreamPriorityData` struct holds the stream dependency information: exclusivity, dependent stream ID, and weight.

Note that, since the payload types have changed, the live (nghttp2-based) implementation and relevant tests have similarly been updated to use the new encapsulated types.

The API is fairly straightforward. The decoder is implemented as a simple funnel: you pass in bytes, they are (usually) consumed, and eventually a frame is emitted once all required bytes have been read. The only exception is if the current byte buffer does not contain enough bytes to construct a frame header (9 bytes), then nothing will happen.

The encoder API offers a one-shot method containing an optimization. You hand it a frame and a buffer into which the frame's encoded bytes are to be written. The encoded frame header (9 bytes) will be written into that buffer, as will anything else dynamically encoded by this method. Data already contained in a byte buffer will be returned, in order, as an `[IOData]`. Thus a DATA frame will have its 9-byte header encoded into the input buffer and the `IOData` within the frame's `payload` will be returned. Similarly, HEADERS and PUSH_PROMISE frames, which have their headers encoded to a separate buffer by NIOHPACK, will have the resultant buffers returned via the `[IOData]` return value.

The purpose of this trick is to elide some additional copying that isn't strictly necessary; the data is already encoded and can be sent in multiple chunks as-is, without manually copying them into a single monolithic buffer.

Result:

There's now a nice pure-Swift frame encoder/decoder available to use 😊


Lukasa:
Thanks a bunch for this @AlanQuatermain, I’ll give this a full review tomorrow.

AlanQuatermain:
…aaaand there's a whole host of merge issues now. Sigh.

AlanQuatermain:
Okay, the new tests have been updated and it's all happy again. Yay!

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+    /// the buffer:
+    ///
+    /// - If the input buffer contains more bytes beyond the current frame, those bytes will
+    ///   remain in the buffer; you may continue calling `decode(bytes:)` until the buffer is
+    ///   empty, or until its size does not change.
+    /// - If a given buffer does not contain a complete frame header (9 bytes), this method
+    ///   will take no action, and the buffer will remain unchanged when the method returns.
+    /// - If the buffer does not contain an entire frame including payload, it will be cached
+    ///   internally by the decoder, and the next call to `decode(bytes:)` will append any
+    ///   new bytes to that cache in an effort to complete that frame's payload.
```
Lukasa: I think it might be nicer to have some explicit transfer of ownership of the `ByteBuffer`, probably via a different method to this one. My concern is that because we're passing the `ByteBuffer` `inout`, it's not clear that we may be transferring ownership of that buffer to this object, and it's also not immediately clear when exactly that happens.

In fact, my original assumption was that if this method returns `nil` then then `ByteBuffer` is owned by this object, but that's not actually true. Instead, it's if `bytes.readableBytes == 0` at the end of the call.

It might be nicer to have a method to call that unconditionally owns the `ByteBuffer`, and then the user can simply call this method until it returns `nil`. It also makes state management in the connection object a bit easier, as we have a single cumulation buffer when we're single-stepping the state machine.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+    ///
+    /// - Parameter bytes: Raw bytes received, ready to decode. On exit, any consumed bytes
+    ///                    will no longer be in the buffer.
+    /// - Returns: A decoded frame type or `nil` if more bytes are needed to complete the frame.
+    /// - Throws: Errors specified by RFC 7540 § 6 for mis-sized frames or incorrect
+    ///           stream/connection association.
+    func decode(bytes: inout ByteBuffer) throws -> HTTP2Frame? {
+        switch self.state {
+        case .idle:
+            return try readPotentialFrame(from: &bytes)
```
Lukasa: Can we use the explicit `self` keyword where possible please? :smile:

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+        case .accumulatingData(let header, var currentBytes):
+            let required = header.length - currentBytes.readableBytes
+            if bytes.readableBytes >= required {
+                let view = bytes.viewBytes(at: bytes.readerIndex, length: required)
+                currentBytes.write(bytes: view)
+                bytes.moveReaderIndex(forwardBy: required)
+                return try readFrame(withHeader: header, from: &currentBytes)
```
Lukasa: Explicit `self` here too, please. :smile:

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+    ///         implementation will silently collect and drop all bytes associated with frames
+    ///         whose type is unrecognized; it will not throw an error or otherwise provide
+    ///         those bytes for processing elsewhere.
+    ///
+    /// - Parameter bytes: Raw bytes received, ready to decode. On exit, any consumed bytes
+    ///                    will no longer be in the buffer.
+    /// - Returns: A decoded frame type or `nil` if more bytes are needed to complete the frame.
+    /// - Throws: Errors specified by RFC 7540 § 6 for mis-sized frames or incorrect
+    ///           stream/connection association.
+    func decode(bytes: inout ByteBuffer) throws -> HTTP2Frame? {
```
Lukasa: I appreciate that this is a wrinkle in this decoder, but I think we should consider whether we want to manually fragment `DATA` frames.

It is quite possible for `DATA` frames to be *very* large, and currently this decoder will make all the bytes in that frame contiguous in a single chunk of memory before forwarding it on. In many cases this may require a truly substantial block of memory, which isn't really ideal. However, this decoder could essentially manually "fragment" `DATA` frames by synthesising smaller ones out of whatever bytes it has, and ensuring that the flags are only persisted on the last of the synthesised frames. So long as it only polices frame length on the true header, instead of the synthesised ones, it'll be safe to do this.

Is that ok?

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+                return try readFrame(withHeader: header, from: &currentBytes)
+            } else {
+                currentBytes.write(buffer: &bytes)  // consume entire input buffer
+                return nil  // no frame (yet)
+            }
+        }
+    }
+    
+    private func readPotentialFrame(from bytes: inout ByteBuffer) throws -> HTTP2Frame? {
+        guard let header = bytes.readFrameHeader() else {
```
Lukasa: Do we want to add `assert` statements to validate our assumptions about what state we're in?

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+            // start accumulating bytes and update our state
+            var buffer = allocator.buffer(capacity: header.length)
+            buffer.write(buffer: &bytes)    // consumes the bytes from the input buffer
+            self.state = .accumulatingData(header, buffer)
+            return nil
+        }
+    }
+    
+    private func readFrame(withHeader header: FrameHeader, from bytes: inout ByteBuffer) throws -> HTTP2Frame? {
+        assert(bytes.readableBytes >= header.length, "Buffer should contain at least \(header.length) bytes.")
```
Lukasa: Do we want to add state assertions here too?

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+            
+            payload = .origin(origins)
+            
+        default:
+            // RFC 7540 § 4.1 https://httpwg.org/specs/rfc7540.html#FrameHeader
+            //    "Implementations MUST ignore and discard any frame that has a type that is unknown."
+            self.state = .idle
+            bytes.moveReaderIndex(forwardBy: header.length)
+            return nil
+        }
```
Lukasa: This is *quite* a `switch` statement! Mind pulling the case bodies out into helper functions?

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+//
+//  Created by Jim Dovey on 11/14/18.
+//
+
+import NIO
+import NIOHPACK
+
+/// A temporary item used to hold onto the HPACK encoder/decoder. Something in the real stream state engine will
+/// take its place soon.
+class HTTP2StreamData {
```
Lukasa: HPACK encoders and decoders are not per-stream, they are per-connection. We don't need this: the `FrameDecoder` can simply own a `HPACKDecoder`, and the `FrameEncoder` can simply own a `HPACKEncoder`.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+                                                             weight: bytes.readInteger()!)
+                bytesToRead -= 5
+            } else {
+                priorityData = nil
+            }
+            
+            // slice the input buffer if necessary
+            let headerByteSize = bytesToRead - Int(padding)
+            let headers: HPACKHeaders
+            if bytes.readableBytes == headerByteSize {
```
Lukasa: It may be easier in all of these blocks just to perform the unconditional slice operation, rather than branch on the length. This branch will so rarely be taken that it's almost certainly not going to help.

Lukasa: (This note applies to all of these blocks).

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+            let padding: UInt8
+            if flags.contains(.padded) {
+                padding = bytes.readInteger()!
+                bytesToRead -= 1
+            } else {
+                padding = 0
+            }
+            
+            let raw: UInt32 = bytes.readInteger()!
+            let promisedStreamID = HTTP2StreamID(knownID: Int32(raw & ~0x8000_0000))
```
Lukasa: This streamID is also not allowed to be `.rootStream`.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+            
+            let debugData: ByteBuffer?
+            let extraLen = header.length - 8
+            if extraLen > 0 {
+                debugData = bytes.readSlice(length: extraLen)
+            } else {
+                debugData = nil
+            }
+            
+            payload = .goAway(lastStreamID: HTTP2StreamID(knownID: Int32(raw & ~0x8000_000)),
```
Lukasa: We have a lot of silently ignoring the top bit of the stream ID, when I think strictly speaking setting it is supposed to be an error. Should we have an extension on `HTTPStreamID` that adds a throwing initializer that validates the stream ID doesn't have the top bit set, and throws if it does?

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+        }
+        
+        self.state = .idle
+        let frame = HTTP2Frame(streamID: streamID, flags: flags, payload: payload)
+        
+        switch payload {
+        case .headers where !flags.contains(.endHeaders), .pushPromise where !flags.contains(.endHeaders):
+            // don't emit the frame, store it and wait for CONTINUATION frames to arrive
+            self.currentHeaderFrame = frame
+            return nil
```
Lukasa: We may as well put this logic in the relevant branch of the `switch` statement, rather than essentially `switch` on the same data twice.

Lukasa: Also, all the other cases should be checking that this value is `nil`, and throwing if it isn't. That will let us detect violations of RFC 7540.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+        
+        let stream = HTTP2StreamData.getStream(for: streamID, allocator: self.allocator)
+        
+        // this just contains a header block fragment and flags
+        let headers: HPACKHeaders
+        if bytes.readableBytes == length {
+            headers = try stream.decoder.decodeHeaders(from: &bytes)
+        } else {
+            var slice = bytes.readSlice(length: length)!
+            headers = try stream.decoder.decodeHeaders(from: &slice)
```
Lukasa: This logic is not quite right. A header block is decompressed as a single unit, even if it's split across multiple CONTINUATION frames. Essentially, you can think of CONTINUATION frames as a special jumbo frame abstraction solely for HEADERS.

The TL;DR here is that you have to buffer the frame parts all together and then decompress them as one.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+    /// copy it into another buffer, with the corresponding allocation overhead.
+    ///
+    /// - Parameters:
+    ///   - frame: The frame to encode.
+    ///   - buf: Destination buffer for the encoded frame.
+    /// - Returns: An array containing zero or more additional buffers to send, in
+    ///            order. These may contain data frames' payload bytes, encoded
+    ///            header fragments, etc.
+    /// - Throws: Errors returned from HPACK encoder.
+    func encode(frame: HTTP2Frame, to buf: inout ByteBuffer) throws -> [IOData] {
```
Lukasa: Will the returned blobs ever have more than two elements? If not, it's better to return these as a tuple than as an array, to avoid the heap allocation to store the returned elements.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+//      |                   Frame Payload (0...)                      ...
+//      +---------------------------------------------------------------+
+        
+        // skip 24-bit length for now, we'll fill that in later
+        buf.moveWriterIndex(forwardBy: 3)
+        
+        // 8-bit type
+        buf.write(integer: frame.payload.code)
+        // 8-bit flags -- we don't use padding when encoding in NIO, though
+        buf.write(integer: frame.flags.subtracting(.padded).rawValue)
```
Lukasa: If we don't support padding (which is reasonable at this time), we should throw rather than silently suppress it.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+        // 32-bit stream identifier -- ensuring the top bit is empty
+        buf.write(integer: frame.streamID.safeNetworkStreamID ?? 0)
```
Lukasa: In this context just pull out the network stream ID and force-unwrap it: we're going to move away from this insane `class`-based representation when we move away from nghttp2.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+        
+        func writePayloadSize(_ size: Int) {
+            var bytes: (UInt8, UInt8, UInt8)
+            bytes.0 = UInt8((size & 0xff_00_00) >> 16)
```
Lukasa: Any reason not to just issue `buf.set(integer:)` calls for each of these, instead of having the temporary tuple?

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+        buf.moveWriterIndex(forwardBy: 3)
+        
+        // 8-bit type
+        buf.write(integer: frame.payload.code)
+        // 8-bit flags -- we don't use padding when encoding in NIO, though
+        buf.write(integer: frame.flags.subtracting(.padded).rawValue)
+        // 32-bit stream identifier -- ensuring the top bit is empty
+        buf.write(integer: frame.streamID.safeNetworkStreamID ?? 0)
+        
+        func writePayloadSize(_ size: Int) {
```
Lukasa: This probably belongs as a private extension on `ByteBuffer`.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+        // frame payload follows, which depends on the frame type itself
+        switch frame.payload {
+        case .data(let data):
+            writePayloadSize(data.readableBytes)
+            return [data]
+            
+        case .headers(let headers, let priority):
+            var payloadSize = 0
+            if let priority = priority {
+                // TODO: throw the proper error here, or determine if we can assume it's valid
```
Lukasa: If the error is about the nil-coalescing, I've noted that above. Otherwise, any stream ID is valid here.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+                payloadSize += 5
+            }
+            
+            // HPACK-encoded data. Find the stream and its associated encoder/decoder
+            let stream = HTTP2StreamData.getStream(for: frame.streamID, allocator: self.allocator)
+            try stream.encoder.beginEncoding(allocator: self.allocator)
+            try stream.encoder.append(headers: headers)
+            let encoded = try stream.encoder.endEncoding()
+            
+            payloadSize += encoded.readableBytes
```
Lukasa: You probably don't need to do this math in each of these cases. Instead you should be able to backfill by checking the number of `readableBytes` and subtracting the space you left at the start for the frame header.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+            
+        case .settings(let settings):
+            writePayloadSize(settings.count * 6)
+            for setting in settings {
+                buf.write(integer: setting.parameter.networkRepresentation)
+                buf.write(integer: setting._value)
+            }
+            
+        case .pushPromise(let streamID, let headers):
+            let streamVal: UInt32 = UInt32(streamID.networkStreamID ?? 0) & ~0x8000_0000
```
Lukasa: The bitwise-AND shouldn't be necessary: this value should never be invalid, or `HTTP2StreamID` isn't doing its job.

Sources/NIOHTTP2/HTTP2StreamID.swift
```diff
@@ -84,12 +88,27 @@ public class HTTP2StreamID {
         precondition(self.actualStreamID == -1)
         self.actualStreamID = newID
     }
+    
+    /// Obtains a stream-id value with its most significant bit properly cleared, ready to
+    /// be sent on the wire.
+    internal var safeNetworkStreamID: UInt32? {
+        guard self.actualStreamID >= 0 else { return nil }
+        return UInt32(self.actualStreamID) & ~HTTP2StreamID.reservedBitMask
```
Lukasa: We should clear this on creation, rather than on emission. We should trap if users attempt to create a stream ID too large.

Lukasa:
I'll note that I haven't yet reviewed the tests, I'll do that when I think we have the functional code ready to rock.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,738 @@
...
+    private var streamDecoders: Dictionary<HTTP2StreamID, HPACKDecoder> = [:]
+    
+    private enum ParserState {
+        case idle
+        case accumulatingData(FrameHeader, ByteBuffer)
+    }
+    
+    private var state: ParserState = .idle
+    private var allocator: ByteBufferAllocator
+    private var currentHeaderFrame: HTTP2Frame? = nil
```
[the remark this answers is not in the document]
AlanQuatermain: My problem with that was that there's then a sort of sub-state issue here. We currently have 'idle' and 'collecting frame payload' states. Now there would be additional 'have some headers, but now idle' and 'have some headers, now collecting another frame's payload' states. As such, it seemed that the issue of there being a complete HEADERS frame sitting around waiting for continuations was outside the purview of the `ParserState` enum, since that relates only to a single frame's ingestion state.

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: Oops 😄

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: I'm not really transferring the ownership of that buffer, I'm explicitly decoding its bytes into some other types. If I start accumulating bytes, I'm creating my own temporary buffer matching the expected payload size and copying the current bytes into that. I'm specifically not making any assumptions about the byte buffer I'm being given, for example that there's no already-read data in the buffer so I'm not going to keep large amounts of memory around if I just claim ownership.

Additionally, the intention here is to handle the case where the networking stack hands you, the caller, a 4KB buffer containing a HEADERS frame, two CONTINUATIONs, and a DATA frame, all in one buffer. If I'm taking ownership of that buffer in any way, I just took ownership of four frames' worth of data in a function that expects to return at most a single frame.

Chances are that the only frames likely to need to be built out of multiple passed-in buffers are DATA frames, and I'm much happier knowing that I'm only on the hook for the bytes of that frame, not whatever else happened to be allocated for the capacity of some other buffer. I'm reticent to go around purging all previously-read bytes from a given buffer because—since I don't own it—I don't know how it's going to be used beyond here, so I want to avoid the whole 'unnecessarily reallocating a large buffer' issue.

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: Hm, I hadn't thought of that. That's a neat idea, I'll look into it. Perhaps for another update though?

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: Sounds like a good idea.

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: Well, I *suppose*…

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: Oh. Weird, I didn't realize header compression could be re-used across different streams.

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: Good catch

Sources/NIOHTTP2/HTTP2FrameParser.swift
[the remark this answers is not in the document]
AlanQuatermain: Actually, given that I need to accumulate all header fragments into a single buffer, this is now manageable — we'd have `idle`, `accumulating frame` and `accumulating header bytes` states.

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: It looked like `set(integer:)` was doing a bunch of things based on the size of the integer, which seemed a bit much for a single byte. Writing these as a `ContiguousCollection` looked much faster, and as part of every frame header it seemed a logical place to think about that sort of optimization. Of course, I'm not entirely clear how the `set(integer:)` implementation will compile down and optimize, but it ultimately goes through some endianness logic and a memory copy even for a single byte, so…

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: In my original variant I pulled out `_UInt24` and used that along with a `ByteBuffer` extension. Since I was only ever using it from one method in this implementation it seemed overkill to add it to `ByteBuffer`. Plus my OCD-ish tendencies wouldn't let me go without ensuring there's a parallel `write()` implementation, etc.

*twitches*

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: Ok, I'll assume it's valid. I'll leave it up to the creator of the `HTTP2Frame` instance to make sure it's not passing an idle/closed stream ID here.

Sources/NIOHTTP2/HTTP2FrameParser.swift
Lukasa: > Additionally, the intention here is to handle the case where the networking stack hands you, the caller, a 4KB buffer containing a HEADERS frame, two CONTINUATIONs, and a DATA frame, all in one buffer. If I'm taking ownership of that buffer in any way, I just took ownership of four frames' worth of data in a function that expects to return at most a single frame.

My concern is that *an* entity needs to take ownership of this buffer and manage accumulating within in. If it's not the frame decoder, it has to be the `ConnectionStateMachine`, but one way or another an entity is going to need to own this. This is because the HTTP/2 implementation needs to be a single-step implementation (that is, we parse and operate on one frame at a time) for the sake of correctness.

Given that someone, somewhere needs to take ownership of the buffer my inclination is to have the frame parser do it. I'm willing to be told that you'd prefer that the `ConnectionStateMachine` owns it, but I think that design doesn't really make much sense.

Sources/NIOHTTP2/HTTP2FrameParser.swift
Lukasa: Good catch. I still think I want an initialiser here, one that expressly indicates that this is a received stream ID and so we should ignore the top bit.

Sources/NIOHTTP2/HTTP2FrameParser.swift
Lukasa: Hrm. So the endianness and the memory copy is a bit of a red herring (the endianness swap operation for UInt8 is inlinable and a no-op, so that all just goes away). However, in my quick experiment I noticed that actually doing this as three set ops leads to hilariously more complex assembly than doing it as a single operation on a tuple. There is no good reason for this: I consider that a failure of the optimiser in Swift, which should be able to spot that these are two utterly equivalent operations. I'll investigate further, but for now we can leave this approach in place.

Sources/NIOHTTP2/HTTP2FrameParser.swift
Lukasa: Heh, I recommend suppressing the OCDish tendencies. This is a special case behaviour that we can just use Int with, rather than bring forward _UInt24.

Sources/NIOHTTP2/HTTP2FrameParser.swift
Lukasa: It's totally allowed to pass an idle or closed stream. PRIORITY frames are not sent on a stream, even though they carry a stream ID. Yeah, it's real weird.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+            }
+        case .simulatingDataFrames(let header, let padding, let remaining):
+            let streamID = HTTP2StreamID(knownID: Int32(header.rawStreamID & ~0x8000_0000))
+            let flags = HTTP2Frame.FrameFlags(rawValue: header.flags)
+            if bytes.readableBytes < remaining {
+                // ping out a fake DATA frame containing these bytes, without the END_STREAM flag being set
+                let slice = bytes
+                bytes.clear()   // consuming all the bytes
+                self.state = .simulatingDataFrames(originalHeader: header, padding: padding, remainingBytes: remaining - slice.readableBytes)
+                return HTTP2Frame(streamID: streamID, flags: [], payload: .data(.byteBuffer(slice)))
```
Lukasa: This branch is not quite right.

It's conceivable that the following statement is true: `bytes.readableBytes < remaining && bytes.readableBytes > (remaining - padding)`. If this occurs, this operation will incorrectly include padding bytes in the DATA frame body, which is not right. I recommend adding a test for this case to verify it misbehaves, and then applying a fix.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+            } else {
+                continuationBytes.write(buffer: &bytes)
+                return nil
+            }
+        }
+    }
+    
+    private mutating func emitSyntheticHeadersFrame(_ originalHeader: FrameHeader, bytes: inout ByteBuffer) throws -> HTTP2Frame? {
+        let newHeader = FrameHeader(length: bytes.readableBytes, type: originalHeader.type,
+                                    flags: originalHeader.flags | HTTP2Frame.FrameFlags.endHeaders.rawValue,
```
Lukasa: This looks wrong to me: specifically, I think it mishandles PADDING on HEADERS frames. It keeps the PADDING flag intact, but does not preserve either the PADDING length or the PADDING bytes, which is almost certainly going to cause a HPACK error.

I recommend writing a test for this too before fixing it. :smile:

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+            guard streamID != .rootStream else {
+                // DATA frames MUST be associated with a stream. If a DATA frame is received whose
+                // stream identifier field is 0x0, the recipient MUST respond with a connection error
+                // (Section 5.4.1) of type PROTOCOL_ERROR.
+                throw NIOHTTP2Errors.ConnectionError(code: .protocolError)
+            }
+            
+            var slice = bytes
+            let consumed = slice.readableBytes
+            bytes.clear()       // consumed all the bytes
```
Lukasa: I think this operation is a fancy way to spell `var slice = bytes.readSlice(length: bytes.readableBytes)`

Lukasa: In fact, it's worse than that as `clear` will definitely CoW in this mode.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+                
+                if (flags.contains(.endHeaders)) {
+                    // we're done: send out a frame
+                    return try self.emitSyntheticHeadersFrame(header, bytes: &buffer)
+                } else {
+                    self.state = .accumulatingHeaderBlockFragments(header, buffer)
+                }
+            } else {
+                cbytes = bytes
+                bytes.clear()
```
Lukasa: Don't call `clear` here, it'll trigger a CoW. Use `readSlice` to consume the readable bytes instead.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+                currentBytes.write(buffer: &bytes)  // consume entire input buffer
+                return nil  // no frame (yet)
+            }
+        case .simulatingDataFrames(let header, let padding, let remaining):
+            let streamID = HTTP2StreamID(knownID: Int32(header.rawStreamID & ~0x8000_0000))
+            let flags = HTTP2Frame.FrameFlags(rawValue: header.flags)
+            if bytes.readableBytes < remaining {
+                // ping out a fake DATA frame containing these bytes, without the END_STREAM flag being set
+                let slice = bytes
+                bytes.clear()   // consuming all the bytes
```
Lukasa: Don't call `clear` here, it'll trigger a CoW. Use `readSlice` instead.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+            bytes.clear()       // consumed all the bytes
+            
+            let flags = HTTP2Frame.FrameFlags(rawValue: header.flags)
+            let padding: UInt8
+            if flags.contains(.padded) {
+                padding = slice.readInteger()!
+            } else {
+                padding = 0
+            }
+            self.state = .simulatingDataFrames(originalHeader: header, padding: padding, remainingBytes: header.length - consumed)
```
Lukasa: This also does not correctly handle the case where we have more bytes than the content portion of the DATA frame, but less than the content + padding. Here we'll once again incorrectly emit padding bytes into the DATA stream.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+        let payload: HTTP2Frame.FramePayload
+        do {
+            switch header.type {
+            case 0:
+                payload = try self.parseDataFramePayload(length: header.length, streamID: streamID, flags: flags, bytes: &bytes)
+            case 1:
+                if flags.contains(.endHeaders) {
+                    payload = try self.parseHeadersFramePayload(length: header.length, streamID: streamID, flags: flags, bytes: &bytes)
+                } else {
+                    let slice = bytes.readSlice(length: header.length)!
```
Lukasa: This mishandles the possibility that this frame contains PADDING: we need to strip that off.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+            case 3:
+                payload = try self.parseRstStreamFramePayload(length: header.length, streamID: streamID, bytes: &bytes)
+            case 4:
+                payload = try self.parseSettingsFramePayload(length: header.length, streamID: streamID, flags: flags, bytes: &bytes)
+            case 5:
+                if flags.contains(.endHeaders) {
+                    payload = try self.parsePushPromiseFramePayload(length: header.length, streamID: streamID, flags: flags, bytes: &bytes)
+                } else {
+                    // don't emit the frame, store it and wait for CONTINUATION frames to arrive
+                    let slice = bytes.readSlice(length: header.length)!
```
Lukasa: This mishandles PADDING as the HEADERS section does.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+                // CONTINUATION frame should never be found here -- if they are, they're out of sequence
+                throw NIOHTTP2Errors.ConnectionError(code: .protocolError)
+            case 10:
+                payload = try self.parseAltSvcFramePayload(length: header.length, streamID: streamID, bytes: &bytes)
+            case 12:
+                payload = try self.parseOriginFramePayload(length: header.length, streamID: streamID, bytes: &bytes)
+            default:
+                // RFC 7540 § 4.1 https://httpwg.org/specs/rfc7540.html#FrameHeader
+                //    "Implementations MUST ignore and discard any frame that has a type that is unknown."
+                throw IgnoredFrame()
```
Lukasa: No need to `throw` here, just hoist the catch block up to this case.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+            }
+        } catch _ as IgnoredFrame {
+            self.state = .idle
+            return nil
+        } catch _ as NIOHPACKError {
```
Lukasa: This should be `catch is NIOHPACKError`

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+                // PROTOCOL_ERROR.
+                throw NIOHTTP2Errors.ConnectionError(code: .protocolError)
+            }
+        } else {
+            padding = 0
+            dataLen = length
+        }
+        
+        let buf: ByteBuffer
+        if bytes.readableBytes == dataLen {
```
Lukasa: We don't need this `if` branch, we can just always issue `readSlice`.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+            padding = bytes.readInteger()!
+            bytesToRead -= 1
+            if bytesToRead <= Int(padding) {
+                // Padding that exceeds the size remaining for the header block fragment
+                // MUST be treated as a PROTOCOL_ERROR.
+                throw NIOHTTP2Errors.ConnectionError(code: .protocolError)
+            }
+        } else {
+            padding = 0
+        }
```
Lukasa: This padding logic looks like common code to me: may be best to factor it out.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+            if let priority = priority {
+                var dependencyRaw = UInt32(priority.dependency.networkStreamID ?? 0)
+                if priority.exclusive {
+                    dependencyRaw |= 0x8000_0000
+                }
+                buf.write(integer: dependencyRaw)
+                buf.write(integer: priority.weight)
+            }
+            
+            // HPACK-encoded data. Find the stream and its associated encoder/decoder
```
Lukasa: This comment is slightly out of date now.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+            try self.headerEncoder.beginEncoding(allocator: self.allocator)
```
Lukasa: Hrm. This suggests a possible API change, which is that the encoder could encode *into* a provided buffer. That will let us re-use this buffer here when necessary, which might be quite helpful.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+            
+        case .pushPromise(let streamID, let headers):
+            let streamVal: UInt32 = UInt32(streamID.networkStreamID ?? 0)
+            buf.write(integer: streamVal)
+            
+            try self.headerEncoder.beginEncoding(allocator: self.allocator)
+            try self.headerEncoder.append(headers: headers)
+            let encoded = try self.headerEncoder.endEncoding()
+            
+            writePayloadSize(encoded.readableBytes + 4)
```
Lukasa: Missing the keyword `self` on all of the invocations of this function. :wink:

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+    private var headerDecoder: HPACKDecoder
+    
+    private struct IgnoredFrame : Error {}
+    
+    private enum ParserState {
+        /// We are not in the middle of parsing any frames.
+        case idle
+        
+        /// We are accumulating payload bytes for a single frame.
+        case accumulatingData(FrameHeader, ByteBuffer)
```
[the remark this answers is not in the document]
AlanQuatermain: I prefer to leave these out when it's fairly clear what they are. The others have labels because it's not immediately obvious what the difference is between the various buffers or integer types they contain.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+        
+        /// We are receiving bytes from a DATA frame payload, and are emitting multiple DATA frames,
+        /// one for each chunk of bytes we see here.
+        case simulatingDataFrames(originalHeader: FrameHeader, padding: UInt8, remainingBytes: Int)
+        
+        /// We are accumulating a CONTINUATION frame.
+        case accumulatingContinuationPayload(originalHeader: FrameHeader, headerBytes: ByteBuffer, continuationHeader: FrameHeader, continuationBytes: ByteBuffer)
+        
+        // We are waiting for a new CONTINUATION frame to arrive.
+        case accumulatingHeaderBlockFragments(FrameHeader, ByteBuffer)
```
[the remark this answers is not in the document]
AlanQuatermain: Same response here 😉

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,835 @@
...
+    }
```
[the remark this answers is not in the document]
AlanQuatermain: That makes sense.

AlanQuatermain:
It's just occurred to me that the `state` only applies per stream. Le sigh. That needs to be swapped out again, with another accumulation buffer above the states to accumulate frame headers (because we can't tell which stream to associate with until we read the headers). We can't say 'a CONTINUATION didn't follow a HEADERS, therefore ERROR' just because we received HEADERS on stream 1, DATA on stream 3, and CONTINUATION on stream 1. Argh.

AlanQuatermain:
I suspect this should use a general accumulating/idle state for the encoder as a whole, then a set of per-stream states that all operate with a complete frame header and are accumulating a payload in some form or fashion.

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: It's not a function on `self` 😉

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: This has all been rewritten with a new state machine.

Sources/NIOHTTP2/HTTP2FrameParser.swift
AlanQuatermain: I manually eat the padding and remove that flag in the new state machine.

AlanQuatermain:
Well, I reimplemented a whole bunch of stuff over the last few days, but then git decided that I was secretly rebasing and wiped it all out. Yay.

AlanQuatermain:
Apparently it also cleared Xcode's undo buffer, which before has saved me from these things by allowing me to revert the file. Holy shit, I just lost a week's work.

ianpartridge:
reflog has saved me in the past...

AlanQuatermain:
I've resurrected the commit, but it's mismatched against master now, since this commit came before I tried to rebase. There's some Docker file mismatches in here which I'm hoping won't break things when I cherry-pick it onto the main branch…

AlanQuatermain:
Phew. Tests still run, and it *looks* like it's all survived. Egads that was scary.

Some updates:

* `HTTP2FrameDecoder` now has a much more involved state machine.
* The decoder's API has been modified to better handle the case where a single `ByteBuffer` input happens to contain more than one frame. Now instead of 'buffer in, frame out', we have two separate functions. One is used to feed incoming data into the decoder, and another asks it for the next decoded frame. You can drop in data and then repeatedly call `nextFrame()` until it either throws or returns `nil`.

Note that there's one more thing to be done here: we are currently acting as though everything arrives on a single stream. The state machine can be in 'collecting CONTINUATION frames' state, and if it sees a frame other than a CONTINUATION it'll bail. That should *only* be the case if the other frame is associated with the same **stream ID** as the initial HEADERS/PUSH_PROMISE frame.

Thus I need to carefully consider the best way of handling this case. We may need one state for accumulating frame payloads and a second one for managing the accumulation of multiple frames into a single `HTTP2Frame` instance. The former would ignore stream ID, the latter would be explicitly collecting only frames whose stream IDs match.

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -61,10 +61,26 @@ public struct HPACKHeaders {
...
+                    HPACKHeader(start: $0.name.start, nameLength: $0.name.length, valueStart: $0.value.start, valueLength: $0.value.length)
+                }
+                return _Storage(buffer: buf, headers: hpack)
+            } else {
+                return nil
+            }
+        }
+        
+        if let store = store {
+            // the compiler complains if I just try to assign to self._storage here...
```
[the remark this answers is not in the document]
AlanQuatermain: It says I referenced 'self' before calling 'self.init' or assigning 'self'. Yeah, doesn't make much sense to me either.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,991 @@
...
+    mutating func accumulate(bytes: inout ByteBuffer)
+}
+
+/// Ingests HTTP/2 data and produces frames. You feed data in, and sometimes you'll get a complete frame out.
+struct HTTP2FrameDecoder {
+    private struct IgnoredFrame: Error {}
+    
+    /// The state for a parser that is currently idle.
+    private struct IdleParserState: BytesAccumulating {
+        // This is nil in exactly one instance, with an explicit check
```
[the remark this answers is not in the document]
AlanQuatermain: I'll take a little from column A and a little from column C.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,991 @@
...
+    /// whose details are kept in this state.
+    private struct AccumulatingContinuationPayloadParserState: BytesAccumulating {
+        var headerBlockState: AccumulatingHeaderBlockFragmentsParserState
+        var continuationPayload: ByteBuffer
+        let continuationHeader: FrameHeader
+        
+        init(fromAccumulatingHeaderBlockFragments acc: AccumulatingHeaderBlockFragmentsParserState,
+             continuationHeader: FrameHeader) {
+            self.headerBlockState = acc
+            self.continuationPayload = acc.incomingPayload
```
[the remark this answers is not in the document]
AlanQuatermain: Don't have a CoW, man!



(I couldn't resist)

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,991 @@
...
+            
+        case .accumulatingData(var state):
+            if state.header.type == 0 && state.accumulatedBytes.readableBytes > 0 {
+                // We now have enough bytes to read the expected padding
+                // We should only be here if it's a DATA frame with padding and we couldn't read
+                // the padding before:
+                assert(state.header.flags.contains(.padded))
+                let padding: UInt8 = state.accumulatedBytes.readInteger()!
+                self.state = .simulatingDataFrames(SimulatingDataFramesParserState(fromAccumulatingPayload: state, expectedPadding: padding, remainingBytes: state.header.length - 1))
+                break       // fall out of switch to call self.processNextState() again, which will send the simulated frame
```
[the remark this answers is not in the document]
AlanQuatermain: This is cleared up somewhat by the new layout you provided above, which just falls through normally and returns only if not enough data or an error occurred.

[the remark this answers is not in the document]
AlanQuatermain: Oops, wrong area. Let's try this again...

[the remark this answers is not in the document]
AlanQuatermain: Okay, I've lifted the repeat up into the public `nextFrame()` method, so these `break` statements can all become explicit `return .continue(nil)`

Sources/NIOHTTP2/HTTP2FrameParser.swift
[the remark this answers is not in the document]
Lukasa: ;)

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -0,0 +1,991 @@
...
+    ///
+    /// - returns: A decoded frame, or `nil` if no frame could be decoded.
+    /// - throws: An error if a decoded frame violated the HTTP/2 protocol
+    ///           rules.
+    mutating func nextFrame() throws -> HTTP2Frame? {
+        // Start running through our state machine until
+        return try self.processNextState()
+    }
+    
+    private mutating func processNextState() throws -> HTTP2Frame? {
```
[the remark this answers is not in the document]
Lukasa: Yeah, I spotted that. I have separate concerns about the tail-call, but for now it's fine.

AlanQuatermain:
Pulled in latest apple/master branch and squashed my commits.

Lukasa:
Included in #40.

PR #34, opened 2018-12-12
add missing ruby install step to docker file

tomerd:
motivation: ruby is required for varisou scripts and got dropped in the recent changeset

changes: add ruby installation stanza to dockefile


tomerd:

tomerd:

Lukasa:
@tomerd It's news to me

Incidentally, I don't think this PR should be being tested by 4.1 or 4.2, as it's to the master branch which is now 5.0 only.

Lukasa:
For 5.0, I filed [SR-9481](https://bugs.swift.org/browse/SR-9481).

tomerd:
> Incidentally, I don't think this PR should be being tested by 4.1 or 4.2, as it's to the master branch which is now 5.0 only.

@Lukasa it would not normally, but in this case because the commit exists on both branches GitHub gets confused

PR #36, opened 2018-12-29
Initialize HTTP2Parser handler in the handlerAdded event hook.

sergiocampama:
By initializing the handler in the handlerAdded hook, HTTP2Parser supports
being added dynamically in a channelRead event, after the channel has sent
the channelActive event.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2Parser.swift
+++ b/Sources/NIOHTTP2/HTTP2Parser.swift
@@ -185,15 +185,17 @@ public final class HTTP2Parser: ChannelInboundHandler, ChannelOutboundHandler {
     }
 
     public func channelActive(ctx: ChannelHandlerContext) {
-        self.session = NGHTTP2Session(mode: self.mode,
-                                      allocator: ctx.channel.allocator,
-                                      maxCachedStreamIDs: self.maxCachedClosedStreams,
-                                      frameReceivedHandler: { ctx.fireChannelRead(self.wrapInboundOut($0)) },
-                                      sendFunction: { ctx.write(self.wrapOutboundOut($0), promise: $1) },
-                                      userEventFunction: { ctx.fireUserInboundEventTriggered($0) })
-
-        self.flushPreamble(ctx: ctx)
+        self.initializeSession(ctx: ctx)
         ctx.fireChannelActive()
+    }
+
+    public func handlerAdded(ctx: ChannelHandlerContext) {
+        // Check if the channel is already active when this handler is being added,
+        // in case the handler was added dynamically into the pipeline. If so,
+        // call channelActive to set up the handler.
+        if ctx.channel.isActive {
+            self.initializeSession(ctx: ctx)
+        }
     }
 
     public func handlerRemoved(ctx: ChannelHandlerContext) {
@@ -234,6 +236,19 @@ public final class HTTP2Parser: ChannelInboundHandler, ChannelOutboundHandler {
     public func flush(ctx: ChannelHandlerContext) {
         self.reentrancyManager.markFlushPoint()
         self.reentrancyManager.process(ctx: ctx, self.process)
+    }
+  
+    private func initializeSession(ctx: ChannelHandlerContext) {
+      if self.session == nil {
+          self.session = NGHTTP2Session(mode: self.mode,
+                                        allocator: ctx.channel.allocator,
+                                        maxCachedStreamIDs: self.maxCachedClosedStreams,
+                                        frameReceivedHandler: { ctx.fireChannelRead(self.wrapInboundOut($0)) },
+                                        sendFunction: { ctx.write(self.wrapOutboundOut($0), promise: $1) },
+                                        userEventFunction: { ctx.fireUserInboundEventTriggered($0) })
+      
+          self.flushPreamble(ctx: ctx)
+      }
     }
 
     private func flushPreamble(ctx: ChannelHandlerContext) {
--- a/Tests/NIOHTTP2Tests/TestUtilities.swift
+++ b/Tests/NIOHTTP2Tests/TestUtilities.swift
@@ -79,8 +79,13 @@ extension XCTestCase {
     func assertDoHandshake(client: EmbeddedChannel, server: EmbeddedChannel,
                            clientSettings: [HTTP2Setting] = nioDefaultSettings, serverSettings: [HTTP2Setting] = nioDefaultSettings,
                            file: StaticString = #file, line: UInt = #line) throws {
-        client.pipeline.fireChannelActive()
-        server.pipeline.fireChannelActive()
+        // This connects are not semantically right, but are required in order to activate the
+        // channels.
+        //! FIXME: Replace with registerAlreadyConfigured0 once EmbeddedChannel propagates this
+        //         call to its channelcore.
+        let socket = try SocketAddress(unixDomainSocketPath: "/fake")
+        _ = try client.connect(to: socket).wait()
+        _ = try server.connect(to: socket).wait()
 
         // First the channels need to interact.
         self.interactInMemory(client, server, file: file, line: line)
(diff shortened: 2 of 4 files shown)


sergiocampama:
Sure, I can do that, but for illustrative purposes, let me explain the use case, because I don't follow the "writing to channels before they're active isn't a good call" comment, since I'm not writing to a channel/ It's a server and a handler modifies the pipeline during a `channelRead`.

The architecture I have is this initial pipeline configuration.

`channel.pipeline = [Http1_2Multiplexer(I)]`

`Http1_2Multiplexer` peeks into the first line of the first `channelRead`, and sees whether it's HTTP1 or HTTP2. If it's HTTP1, I modify the pipeline with `NIOHTTP1`'s `configureHTTPServerPipeline`, and after that I append my HTTP1 handler, becoming:

`channel.pipeline = [Http1_2Multiplexer(I), HTTPRequestDecoder(IO), HTTPResponseEncoder(IO), MYHTTP1Handler(I)]`

If the request is HTTP2 instead, it becomes:

`channel.pipeline = [Http1_2Multiplexer(I), HTTP2Parser(IO), HTTP2Multiplexer(IO), HTTP2ToHTTP1ClientCodec(IO), MYHTTP1Handler(I)]`

When I add `HTTP2Parser` to the pipeline, it seems like `channelActive` is not getting called, resulting in the `session` being `nil` and failing at runtime when the first `channelRead` is forwarded from `Http1_2Multiplexer` after modifying the pipeline.

So as I understand it, when the initial request comes, the channel becomes active, I then add HTTP2Parser, and it doesn't get the `channelActive` event. This is well before I write anything to the channel, which is why I'm not following the comment :)

Should I instead trigger a `fireChannelActive` event after I modify the pipeline to ensure that the new pipeline is active?

sergiocampama:
Just checked, if I trigger a `fireChannelActive` in the first `channelRead` from `Http1_2Multiplexer` after I modify the pipeline and before trigger the `fireChannelRead`call, `HTTPParser2` works as expected.

Is this expected? It feels dirty having to trigger the `channelActive` event this way.

Lukasa:
> because I don't follow the "writing to channels before they're active isn't a good call" comment, since I'm not writing to a channel/ 

You aren't, but `HTTP2Parser` *does* in the body of your `handlerAdded` method, as it writes out the HTTP/2 server preamble.

> When I add `HTTP2Parser` to the pipeline, it seems like `channelActive` is not getting called

This is expected behaviour, and you're right that right now `HTTP2Parser` does not tolerate it well.

> Should I instead trigger a `fireChannelActive` event after I modify the pipeline to ensure that the new pipeline is active?

No. What I'm suggesting is that you add the following function to `HTTP2Parser` in this PR:

```swift
func handlerAdded(ctx: ChannelHandlerContext) {
    if ctx.channel.isActive {
        self.channelActive(ctx: ctx)
    }
}
```

Alternatively, if you prefer, both `channelActive` and `handlerAdded` could call into a private function that does what `channelActive` currently does.

sergiocampama:
Gotcha, thanks for the pointers! I failed to notice that the `flushPreamble` was actually writing to the channel. I went on the private method route, that way I don't have the side effect of triggering the `channelActive` event on subsequent handlers, which could create some confusion.

I also needed to change the test server/client handling, as without the `connect`s, the `isActive` state was not getting enabled.

Sources/NIOHTTP2/HTTP2Parser.swift
```diff
@@ -185,17 +185,19 @@ public final class HTTP2Parser: ChannelInboundHandler, ChannelOutboundHandler {
...
     public func channelActive(ctx: ChannelHandlerContext) {
-        self.session = NGHTTP2Session(mode: self.mode,
-                                      allocator: ctx.channel.allocator,
-                                      maxCachedStreamIDs: self.maxCachedClosedStreams,
-                                      frameReceivedHandler: { ctx.fireChannelRead(self.wrapInboundOut($0)) },
-                                      sendFunction: { ctx.write(self.wrapOutboundOut($0), promise: $1) },
-                                      userEventFunction: { ctx.fireUserInboundEventTriggered($0) })
-
-        self.flushPreamble(ctx: ctx)
+        initializeSession(ctx: ctx)
```
Lukasa: Nit: The SwiftNIO style is to prefer to use the explicit `self` keyword, e.g. `self.initializeSession`.

Sources/NIOHTTP2/HTTP2Parser.swift
```diff
@@ -185,17 +185,19 @@ public final class HTTP2Parser: ChannelInboundHandler, ChannelOutboundHandler {
...
         ctx.fireChannelActive()
     }
 
+    public func handlerAdded(ctx: ChannelHandlerContext) {
+        // Check if the channel is already active when this handler is being added,
+        // in case the handler was added dynamically into the pipeline. If so,
+        // call channelActive to set up the handler.
+        if ctx.channel.isActive {
+            initializeSession(ctx: ctx)
```
Lukasa: Same note about explicit `self` here.

Sources/NIOHTTP2/HTTP2Parser.swift
```diff
@@ -235,6 +237,17 @@ public final class HTTP2Parser: ChannelInboundHandler, ChannelOutboundHandler {
         self.reentrancyManager.markFlushPoint()
         self.reentrancyManager.process(ctx: ctx, self.process)
     }
+  
+    private func initializeSession(ctx: ChannelHandlerContext) {
+        self.session = NGHTTP2Session(mode: self.mode,
```
Lukasa: This code should probably refuse to double-initialise, which means it should check whether `session` is `nil` before proceeding.

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -79,8 +79,9 @@ extension XCTestCase {
     func assertDoHandshake(client: EmbeddedChannel, server: EmbeddedChannel,
                            clientSettings: [HTTP2Setting] = nioDefaultSettings, serverSettings: [HTTP2Setting] = nioDefaultSettings,
                            file: StaticString = #file, line: UInt = #line) throws {
-        client.pipeline.fireChannelActive()
-        server.pipeline.fireChannelActive()
+        let socket = try SocketAddress(unixDomainSocketPath: "/fake")
+        _ = try client.connect(to: socket).wait()
+        _ = try server.connect(to: socket).wait()
```
Lukasa: This isn't quite right, semantically speaking. Maybe change this to `server.registerAlreadyConfigured0`.

Sources/NIOHTTP2/HTTP2Parser.swift
sergiocampama: done

Sources/NIOHTTP2/HTTP2Parser.swift
sergiocampama: done

Sources/NIOHTTP2/HTTP2Parser.swift
sergiocampama: done

Tests/NIOHTTP2Tests/TestUtilities.swift
sergiocampama: this doesn't work since `EmbeddedChannel` doesn't forward `registerAlreadyConfigured0` to `channelcore`. also tried a bunch of combinations of `bind` and `connect`, and this one was the only one that worked without having to modify `swift-nio` as well. 

Would you prefer that I add the proxy method in `EmbeddedChannel` in another PR?

Lukasa: Gah, you're right. Ok, let's just put a comment on this to say that it's not right, but is required to activate the embedded channel. We can then update EmbeddedChannel in a separate PR.

Lukasa:
@sergiocampama Ah, do you want to retarget this PR against the branch called `nghttp2-support-branch`? That branch is where new releases of the 0.x cycle will be cut from: the `master` branch is being used for 1.0 development, which will not have nghttp2. I'll forward port this PR to `master`, but the CI is currently busted so I don't want to block this PR on it.

Lukasa:

weissi:

sergiocampama:
fixed the latest test error (missing empty line at the end)

Lukasa:

PR #37, opened 2019-01-09
Forward port #36

Lukasa:
This is a forward-port of #36 to the master branch. We probably can't merge this because we're still having trouble with Swift 5 CI here, but I wanted to open this so we don't lose it on the master branch.


Lukasa:
No longer required.

PR #40, opened 2019-02-19
Pure-Swift HTTP/2 implementation

Lukasa:
This patch brings together #28 and #32, and provides the integration work to move to a pure-Swift HTTP/2 implementation.

This patch should pass all relevant unit tests and is suitable for ongoing work. It also moves to Swift 5 and NIO 2.

Note that with this patch things aren't *perfect*. In particular, we have some missing features and edge cases that need to be addressed. However, this represents the ongoing development model for this repository, so we should aim to get things into master so we can move back to more reviewable patches.

@weissi, I have tagged you to review but fundamentally I don't expect you to do much here: it's just ludicrously too big to be reviewed.


Lukasa:
@weissi What's our plan about the license headers?

Sources/NIOHTTP2/ConnectionStateMachine/FrameReceivingStates/ReceivingDataState.swift
```diff
@@ -0,0 +1,58 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
weissi: this applies to all headers, I think this should be `(c) 2018-2019` and in some cases 2017-2019

Lukasa: Right, but the validator rejects them. So what's the proposed change to the validator?

weissi: @Lukasa the one in [`swift-nio`](https://github.com/apple/swift-nio/blob/master/scripts/sanity.sh#L19) has a `function replace_acceptable_years` whose job it is to replace all acceptable year markers with the string `YEARS`. I think we should do the same for the http2 one

weissi: `2018-2019` needs to be added as an acceptable form too

AlanQuatermain:
An alternative format which I believe works would be something along the lines of (in handy Regex format):

(($FILENAME is p)|P)art of $PROJECT.
$PROJECT is Copyright © 2017–2019 Apple, Inc.

Of course IANAL, so maybe check with Brian on the legal team about that (Tomer knows who that is).

Lukasa:
@weissi Ok, ready for review again.

PR #41, opened 2019-02-21
Add Dockerfile to run h2spec.

Lukasa:
Motivation:

When we get around to it we should aim for our example server to pass
h2spec with flying colours. We aren't there yet, but it's helpful to
be able to run h2spec easily.

Modifications:

Add a docker-compose recipe for running h2spec.

Result:

It'll be easier to run h2spec.


MrMage:
Awesome idea, having a passing h2spec should instill even more confidence in the quality of the implementation 👍

normanmaurer:
+1... We do the same in netty and it really helped us to find problems and fix these :)

Lukasa:
The long term plan will be to have this be in CI, to avoid regressing in the future. Not there yet, sadly. One step at a time!

PR #44, opened 2019-02-25
New HTTP/2 currency types.

Lukasa:
Motivation:

As discussed in our HTTP/2 types manifesto, it is helpful to run more
metadata through the HTTP/2 pipeline. This makes it easier to write
ChannelHandlers that fully participate in the HTTP/2 state machine
while avoiding needing to too tightly couple the handlers.

Modifications:

- Added new HTTP/2 pipeline data types.
- Removed HTTP2Frame.FrameFlags, moved flags into relevant payloads.

Result:

Right now, not much. All tests pass with new payload types. The full
data path rewrite will follow in a separate patch, as this one is large
enough.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2Frame.swift
+++ b/Sources/NIOHTTP2/HTTP2Frame.swift
@@ -16,13 +16,168 @@ import NIO
 import NIOHTTP1
 import NIOHPACK
 
+// This file contains the data types used in the HTTP/2 channel pipeline.
+//
+// The HTTP/2 channel pipeline is a slightly complex entity with a number of channel handlers cooperating together to implement a
+// full HTTP/2 stack. These handlers can be divided into three conceptual groups:
+//
+// 1. Core protocol handlers. These provide core protocol features and conformance, and require a deep understanding of the
+//     behaviour of the state machine in the HTTP2Handler.
+// 2. Non-core handlers. These handlers need the full frame information, but do not themselves participate in protocol
+//     conformance.
+// 3. Per-stream handlers. These handlers are usually part of a child HTTP2StreamChannel of the HTTP2StreamMultiplexer.
+//     These handlers are tied to a single stream, and so gain no utility from being able to see or manipulate the HTTP2Frame
+//     data type.
+//
+// For this reason, there are multiple data types used throughout the various parts of the HTTP/2 channel pipeline structure.
+// Each of these can be thought of as a composition of one of the "less featureful" data types with the information needed to
+// become more featureful.
+//
+// These data types are:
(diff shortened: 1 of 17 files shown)


Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -16,13 +16,158 @@ import NIO
...
+// as makes it very hard to confidently associate a specific state change with a single frame.
+//
+// As we're essentially distributing the core protocol implementation across multiple ChanneHandlers, it reasonable to
+// consider the state machine result to be "data".
+
+
+/// A single HTTP/2 frame, along with the state change it triggered in the state machine.
+public struct HTTP2FrameWithMetadata {
+    /// The received HTTP/2 frame.
+    var frame: HTTP2Frame
```
weissi: shouldn't those `var`s be `public`

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -16,13 +16,158 @@ import NIO
...
+    /// The state change triggered in the connection.
+    var metadata: NIOHTTP2ConnectionStateChange
+}
+
+
+/// A single HTTP/2 frame, along with an optional promise that will be fulfilled with
+/// the state change triggered by parsing this frame.
+public struct HTTP2FrameWithMetadataRequest {
+    /// The HTTP/2 frame to send.
+    var frame: HTTP2Frame
```
weissi: and these guys too?

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -39,7 +184,7 @@ public struct HTTP2Frame {
         /// A DATA frame, containing raw bytes.
         ///
         /// See [RFC 7540 § 6.1](https://httpwg.org/specs/rfc7540.html#rfc.section.6.1).
-        case data(IOData)
+        case data(Data)
```
weissi: I find this reads a bit confusing because (`Foundation.`)`Data` is such a well-known type but here you're referring to `FramePayload.Data`...

Lukasa: Would it be helpful to use the explicit fully qualified type name, or do you want a new name here?

weissi: up to you but I think qualifying is good enough

Lukasa: Resolved.

Lukasa:
Ok, ready for re-review.

PR #50, opened 2019-03-05
[nghttp2] Work around two possible crashes.

MrMage:
Context: Our services received (invalid) HTTP2 streams with a disallowed `connection` header. This caused a `nghttp2 error: Invalid HTTP header field was received: frame type: 1, stream: 23, name: [connection], value: [keep-alive]` to be logged, but also caused our serving process to crash subsequently in `onStreamCloseCallback`. After that was worked around, `doOneWrite` was called with a negative length.

This patch does _not_ fix the underlying issue, it only prevents a hard crash of the process until we can migrate to the non-`nghttp2` implementation.

### DIFF
--- a/Sources/NIOHTTP2/NGHTTP2Session.swift
+++ b/Sources/NIOHTTP2/NGHTTP2Session.swift
@@ -490,7 +490,9 @@ class NGHTTP2Session {
 
     /// Called when the stream `streamID` is closed.
     fileprivate func onStreamCloseCallback(streamID: Int32, errorCode: UInt32) throws {
-        let streamData = self.streamIDManager.getStreamData(for: streamID)!
+        guard let streamData = self.streamIDManager.getStreamData(for: streamID) else {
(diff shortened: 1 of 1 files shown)


Sources/NIOHTTP2/NGHTTP2Session.swift
```diff
@@ -490,7 +490,9 @@ class NGHTTP2Session {
 
     /// Called when the stream `streamID` is closed.
     fileprivate func onStreamCloseCallback(streamID: Int32, errorCode: UInt32) throws {
-        let streamData = self.streamIDManager.getStreamData(for: streamID)!
+		guard let streamData = self.streamIDManager.getStreamData(for: streamID) else {
+			throw NIOHTTP2Errors.NoSuchStream(streamID: HTTP2StreamID(knownID: streamID))
+		}
```
Lukasa: Can you fix up the indentation here?

MrMage: Yikes, sorry!

Lukasa:

PR #51, opened 2019-03-05
Support DATA frames with zero-length application data

glbrntt:
Motivation:

DATA frames with zero-length application data would cause the state
machine to wait for more bytes to continue. This could cause consuming
code to deadlock.

Modifications:

Updated the frame parser to only require more data when there are
remaining bytes in the payload and there are zero readable bytes.

Result:

DATA frames with zero-length application data can be parsed.


Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -405,24 +405,11 @@ struct HTTP2FrameDecoder {
...
-                // we're just eating pad bytes now, maintaining state and emitting nothing
-                if state.payload.readableBytes >= state.remainingByteCount {
-                    // we've got them all, move to idle state with any following bytes
-                    state.payload.moveReaderIndex(forwardBy: state.remainingByteCount)
-                    self.state = .accumulatingFrameHeader(AccumulatingFrameHeaderParserState(unusedBytes: state.payload))
-                } else {
-                    // stay in state and wait for more bytes
-                    return .needMoreData
-                }
-            }
```
Lukasa: Why is this block gone?

glbrntt: It's covered by the if-else block below.

glbrntt: Ah, actually, I think that's not quite the case when the we don't have enough padded bytes as we'd return a frame with empty bytes instead of `.needMoreData`.

Lukasa: Agreed, I think this is a new conditional not covered here at all. We probably need a new if block just after the guard to handle this.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -337,6 +337,14 @@ struct HTTP2FrameDecoder {
                 }
 
                 self.state = .simulatingDataFrames(SimulatingDataFramesParserState(fromIdle: state, header: header, expectedPadding: expectedPadding, remainingBytes: remainingBytes))
+
+                // Emit an empty frame if we only have padding; .simulatingDataFrames will handle eating the padding.
+                if expectedPadding == remainingBytes {
+                    let streamID = HTTP2StreamID(networkID: header.rawStreamID)
+                    let dataPayload = HTTP2Frame.FramePayload.Data(data: .byteBuffer(allocator.buffer(capacity: 0)), endStream: header.flags.contains(.endStream), paddingBytes: nil)
```
Lukasa: Nit: NIO style is to use `self` explicitly: can this be `self.allocator`?

Lukasa:

PR #54, opened 2019-03-06
Mark entire StringRing as readable.

Lukasa:
Motivation:

We use a Bytebuffer as backing storage for a string ringbuffer to avoid
needing to reinvent that data structure ourselves. This used to work great
but NIO 2 recently changed the get* functions to trap if they accessed
memory outside the reader/writer indices.

We were never moving the writer index, instead treating the entire buffer
as valid. This led to preconditions, and sadness.

In an ideal world we'd only mark the bytes that are actually active as
readable to ensure that we have the defensive traps from bytebuffer. In
practice, that's a sufficiently complex thing to do that we're better
served just trusting our code and taking off the guard rails.

In future we should go back to seeing whether a CircularBuffer of strings
will work better, now that Strings are fast. I'll make that Johannes'
problem.

Modifications:

- Mark the entire backing buffer as readable.

Result:

Fewer crashes

### DIFF
--- a/Sources/NIOHPACK/StringRing.swift
+++ b/Sources/NIOHPACK/StringRing.swift
@@ -50,6 +50,7 @@ struct StringRing {
             newBytes.reserveCapacity(capacity)
             _ = newBytes.writeBytes(self._storage.viewBytes(at: r, length: w - r))
             self._storage = newBytes
+            self._storage.markAllReadable()
             self.moveHead(to: 0)
             self.moveTail(to: w - r)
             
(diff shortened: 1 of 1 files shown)


Sources/NIOHPACK/StringRing.swift
```diff
@@ -229,7 +232,7 @@ struct StringRing {
         }
         
         // single write, should be straightforward
-        self._storage.withVeryUnsafeBytes { ptr in
+        self.withVeryUnsafeBytes { ptr in
```
weissi: couldn't you use `withUnsafeReadableBytes` here given that you mark everything as readbale?

Lukasa: I could. Does it matter though?

weissi: no

PR #55, opened 2019-03-08
Remove flushability of streams when stream window size is set to zero.

glbrntt:
Motivation:

When setting the stream window size for a stream which had data but
couldn't write it (because of a zero window size) the stream would be
marked as flushable. Setting the window size to zero again would not
mark it as unflushable. On the following window (non-zero) resize it
would be marked as flushable again. However, since this is preceeded by
an assertion that it wasn't already marked as flushable it would cause a
crash.

Modifications:

- Mark the stream as unflushable if it previously had data which it
  couldn't write.

Result:

Fewer crashes

### DIFF
--- a/Sources/NIOHTTP2/OutboundFlowControlBuffer.swift
+++ b/Sources/NIOHTTP2/OutboundFlowControlBuffer.swift
@@ -103,6 +103,9 @@ internal struct OutboundFlowControlBuffer {
             if $0.hasPendingData && !hadData {
                 assert(!self.flushableStreams.contains($0.streamID))
                 self.flushableStreams.insert($0.streamID)
+            } else if !$0.hasPendingData && hadData {
+                assert(self.flushableStreams.contains($0.streamID))
+                self.flushableStreams.remove($0.streamID)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


Tests/NIOHTTP2Tests/FlowControlHandlerTests.swift
```diff
@@ -440,6 +440,29 @@ class FlowControlHandlerTests: XCTestCase {
...
+        self.channel.createStream(streamOne, initialWindowSize: 0)
+
+        let writePromise = self.makeWritePromise()
+        let writtenBuffer = self.channel.writeDataFrame(streamOne, byteBufferSize: 15, promise: writePromise)
+        // The stream has data to write but the window size is zero, so it won't write.
+        self.channel.flush()
+
+        self.channel.updateStreamWindowSize(streamOne, newWindowSize: 15)
+        self.channel.updateStreamWindowSize(streamOne, newWindowSize: 0)
+        self.channel.updateStreamWindowSize(streamOne, newWindowSize: 15)
```
Lukasa: Can you update this test real quick? It'd be nice to have a quick `flush` and check of `receivedFrames` after `newWindowSize: 0` to prove that the stream really has been marked unflushable.

glbrntt:
@Lukasa I had to build this against NIO commit 83350741083ec8a0b36e67108dd352b2fe1c37e6 since there are breaking changes after that which haven't been resolved in nio-http2 yet.

Lukasa:

PR #61, opened 2019-03-14
Ensure preamble only gets sent once.

glbrntt:
Motivation:

In certain cases it was possible for preamble to be sent once; this
would lead to frames getting sent more than once which in turn would
break state machines.

Modifications:

- Check the state of the connection before sending preamble.
- Include a test to check.

Result:

More correct implementation


Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
```diff
@@ -460,6 +460,16 @@ struct HTTP2ConnectionStateMachine {
             return false
         }
     }
+
+    /// Whether the preamble can be sent.
+    var canSendPreamble: Bool {
```
Lukasa: I think this needs to be more emphatic: `mustSendPreamble`.

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -318,6 +318,8 @@ extension NIOHTTP2Handler {
 extension NIOHTTP2Handler {
     /// Issues the preamble when necessary.
     private func writeAndFlushPreamble(context: ChannelHandlerContext) {
+        guard self.stateMachine.canSendPreamble else { return }
```
Lukasa: NIO style is to avoid this single-line:

```swift
guard self.stateMachine.canSendPreamble else {
    return
}
```

Tests/NIOHTTP2Tests/ConfiguringPipelineTests.swift
```diff
@@ -105,4 +105,37 @@ class ConfiguringPipelineTests: XCTestCase {
...
+        let connectionPromise: EventLoopPromise<Void> = self.serverChannel.eventLoop.makePromise()
+
+        // Register the callback on the promise so we run it as it succeeds (i.e. before fireChannelActive is called).
+        let handlerAdded = connectionPromise.futureResult.flatMap {
+            // Use .server to avoid sending client magic.
+            self.serverChannel.configureHTTP2Pipeline(mode: .server)
+        }
+
+        self.serverChannel.connect(to: try SocketAddress(unixDomainSocketPath: "/fake"), promise: connectionPromise)
+        XCTAssertNoThrow(try connectionPromise.futureResult.and(handlerAdded).wait())
```
Lukasa: This `and` is unnecessary: the `flatMap` will cover it. You can just wait for `handlerAdded`.

Tests/NIOHTTP2Tests/ConfiguringPipelineTests.swift
```diff
@@ -105,4 +105,37 @@ class ConfiguringPipelineTests: XCTestCase {
...
+
+        let initialSettingsData: IOData? = try self.serverChannel.readOutbound()
+        XCTAssertNotNil(initialSettingsData)
+
+        guard case .byteBuffer(let initialSettingsBuffer) = initialSettingsData! else {
```
Lukasa: Rather than force unwrap here, let's do `guard case .some(.byteBuffer(let initialSettingsBuffer)) = initialSettingsData else {`

Lukasa:

PR #69, opened 2019-03-25
Allow changing MAX_FRAME_SIZE.

Lukasa:
Motivation:

SETTINGS_MAX_FRAME_SIZE is changeable: we should allow that change to
occur.

Modifications:

- Added fields for communicating MAX_FRAME_SIZE changes.
- Update the encoders/decoders as those settings change.
- Throw errors from the encoders/decoders if violated.

Result:

Users and remote peers can change MAX_FRAMES_SIZE.


Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -923,6 +929,7 @@ struct HTTP2FrameDecoder {
 struct HTTP2FrameEncoder {
     private let allocator: ByteBufferAllocator
     var headerEncoder: HPACKEncoder
+    var maxFrameSize: UInt32 = 1<<14
```
normanmaurer: I am not against using `<<` but I think most of the times people prefer to have the "real value" here. If you want to use `<<` maybe add some link / comment why this is a good default value.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -227,6 +227,7 @@ struct HTTP2FrameDecoder {
     internal var headerDecoder: HPACKDecoder
     private var state: ParserState
     private var allocator: ByteBufferAllocator
+    internal var maxFrameSize: UInt32 = 1<<14
```
normanmaurer: same as below

Sources/NIOHTTP2/HTTP2FrameParser.swift
Lukasa: Heh, it's a good default value because it's what the spec requires. ;) Will add.

normanmaurer: Thats what I assumed :D So adding a link to the spec is what we need :)

Lukasa:
Updated.

PR #71, opened 2019-03-25
Basic PRIORITY frame validation.

Lukasa:
Motivation:

In the future we'll want to use PRIORITY frames to provide feedback on
how we emit data. Additionally, RFC 7540 forbids certain kinds of
PRIORITY frame, which right now we tolerate receiving. We can design a
minor PRIORITY shim to ensure that we have a code path for PRIORITY data
to be passed around, and then use this shim to enforce RFC 7540's
requirements.

Modifications:

- Added support for PRIORITY frame data inbound.
- Enforce that frames may not establish self-dependency.

Result:

Better policing of RFC 7540 constraints.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2Error.swift
+++ b/Sources/NIOHTTP2/HTTP2Error.swift
@@ -184,10 +184,20 @@ public enum NIOHTTP2Errors {
 
     /// A :status header was received with an invalid value.
     public struct InvalidStatusValue: NIOHTTP2Error {
-        var value: String
+        public var value: String
 
         public init(_ value: String) {
             self.value = value
+        }
+    }
+
+    /// A priority update was received that would create a PRIORITY cycle.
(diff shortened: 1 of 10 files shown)


Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -184,12 +184,22 @@ public enum NIOHTTP2Errors {
...
 
         public init(_ value: String) {
             self.value = value
         }
     }
+
+    /// A priority update was received that would create a PRIORITY cycle.
+    public struct PriorityCycle: NIOHTTP2Error {
+        /// The affected stream ID.
+        public var streamID: HTTP2StreamID
```
normanmaurer: should this be a let ?

Lukasa: No: there is no reason to force fields on structures to be immutable if there is no invariant being broken by their mutation.

PR #81, opened 2019-03-27
Require END_STREAM to be set on trailers.

Lukasa:
Motivation:

RFC 7540 requires that the END_STREAM bit be set on a HEADERS frame that
sends trailers. We should enforce that.

Modifications:

- Require that the END_STREAM bit is set on trailers.

Result:

We reject more incorrect flows.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2Error.swift
+++ b/Sources/NIOHTTP2/HTTP2Error.swift
@@ -200,6 +200,16 @@ public enum NIOHTTP2Errors {
             self.streamID = streamID
         }
     }
+
+    /// An attempt was made to send trailers without setting END_STREAM on them.
+    public struct TrailersWithoutEndStream: NIOHTTP2Error {
+        /// The affected stream ID.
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -200,6 +200,16 @@ public enum NIOHTTP2Errors {
             self.streamID = streamID
         }
     }
+
+    /// An attempt was made to send trailers without setting END_STREAM on them.
+    public struct TrailersWithoutEndStream: NIOHTTP2Error {
```
normanmaurer: Not sure what the Swift convention is but should we have Error as suffix ?

Lukasa: Given that this is namespaced in an enum and can only be found by typing `NIOHTTP2Errors.TrailersWithoutEndStream`, I don't think we need to add the word `Error` as well.

PR #84, opened 2019-03-28
Validate no uppercase characters in header field names.

Lukasa:
Motivation:

RFC 7540 forbids uppercase characters in header field names, but we
currently allow them. We shouldn't.

This is the first in a series of patches that will validate header blocks
when reasonable. As a result, it seems larger than necessary, as it adds
a bunch of scaffolding for extending with further validation code.

Modifications:

- Added header field validation scaffolding.
- Validates that header field names do not contain uppercase chars.
- Allows sending RST_STREAM frame before stream setup.

Result:

Better header field validation.

### DIFF
--- a/Sources/NIOHTTP2/HPACKHeaders+Validation.swift
+++ b/Sources/NIOHTTP2/HPACKHeaders+Validation.swift
@@ -1,0 +1,159 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2019 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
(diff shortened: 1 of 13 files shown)


Sources/NIOHTTP2/HPACKHeaders+Validation.swift
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+}
+
+/// An object that can be used to validate if a given header block is a valid request header block.
+fileprivate struct RequestBlockValidator { }
+
+extension RequestBlockValidator: HeaderBlockValidator {
+    fileprivate mutating func validateNextField(name: HeaderFieldName, value: String) throws {
+        return
```
normanmaurer: just interested... why we need a return here (+ in other "empty" functions)?

Lukasa: We don't, they could have empty bodies.

Lukasa: To clarify: I'm going to be extending these functions as I deal with other parts of RFC 7540 § 8.1.2. For now they are just spaces where code will go in future.

PR #88, opened 2019-04-10
Allow zero length SETTINGS frames to be parsed.

glbrntt:
Zero length SETTINGS frames are permitted by RFC 7540 but the frame
parser currently throws a FRAME_SIZE_ERROR.

- Allow zero length SETTINGS frames to be parsed
- Added a test

Parsing zero length SETTINGS frames don't throw an error.


Lukasa:
Thanks @glbrntt!

PR #89, opened 2019-04-10
Move to String-backed HPACK representations.

Lukasa:
Motivation:

When originally written our HPACK implementation targetted Swift 4.2,
which had performance issues when storing mostly-ASCII data in Strings.
For this reason we had a number of data types that used ByteBuffers as
their backing storage and reified Strings when necessary.

In Swift 5 this is no longer worthwhile: in fact, it imposes performance
penalties above and beyond simply using the Strings. For this reason, we
should move to a simpler model where we simply store Strings again.

Modifications:

- Rewrote all HPACK data structures to use Strings as their backing store.
- Removed now-redundant genericism.
- Removed now-unnecessary StringRing data structure and associated tests.

Result:

Faster performance in almost all cases.


Lukasa:
This change provides a substantial performance boost. I benchmarked using h2load to provide a fairly unrepresentative workload (doing a bunch of GETs and sending back almost no body) which puts a lot of strain on the HPACK implementation. Before this patch:

```
                     min         max         mean         sd        +/- sd
time for request:   359.58ms    508.06ms    400.98ms     23.49ms    62.69%
req/s           :     229.99      260.83      248.95       13.88    64.00%
```

After this patch:

```
                     min         max         mean         sd        +/- sd
time for request:   116.21ms    428.62ms    187.57ms     33.18ms    86.33%
req/s           :     481.62      565.77      532.60       31.92    57.00%
```

That's a more than 2x speedup, as well as a substantial decrease in latency in any case that was using indexed headers.

weissi:
AMAZING

AlanQuatermain:
Fantastic work. All that messing about with byte containers was one of the biggest headaches for me I putting it together (though in all fairness I did learn a lot in the process). Very nice to be able to have a sane String-based implementation now.

Lukasa:
Yup, it was very subtle code and being able to performantly remove it is a big win. It also reduces the surface area for further optimisation, so profiling should do a better job finding expensive operations.

PR #104, opened 2019-04-18
Remove unnecessary state CoWs.

Lukasa:
Motivation:

When benchmarking the HTTP/2 implementation I noticed that we were
performing a large number of CoW operations on the Dictionary that backs
the ConnectionStreamState structure: specifically, once per frame sent or
received.

This is substantial and unnecessary allocation overhead that drags down
the performance of the program, sprays the heap with small allocations, and
generally just leads to nasty performance costs with no gain.

Sadly, this turns out not to be the easiest CoW to get rid of, because
the ConnectionStreamState structure is stored in associated data on the
connection state machine state enumeration. This means that our usual tricks
for avoiding the CoW (such as swap()) no longer work.

The only solution is to add a new state that corresponds to no state at all
so that we can make the connection state go away while we mutate it. This is
all a bit gross, frankly. Unfortunately, the act of making this simple change
leads to a huge diff as we have to tweak basically every state machine
function in order to ensure that it does the right thing.

Modifications:

- Added a function to control CoWing the state machine.
- Used that function basically everywhere.
- Made sure that function had assertions.
- Annoyingly had to increase the visibility of a bunch of structs.

Result:

Fewer CoWs, better performance, sadder code reviewers.

### DIFF
--- a/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
+++ b/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
@@ -63,14 +63,14 @@ import NIOHPACK
 /// frames that just incidentally have a stream ID on them, rather than stream-scoped frames like all the others.
 struct HTTP2ConnectionStateMachine {
     /// The state required for a connection that is currently idle.
-    private struct IdleConnectionState: ConnectionStateWithRole, ConnectionStateWithConfiguration {
+    fileprivate struct IdleConnectionState: ConnectionStateWithRole, ConnectionStateWithConfiguration {
         let role: ConnectionRole
         var headerBlockValidation: ValidationState
         var contentLengthValidation: ValidationState
     }
 
     /// The state required for a connection that has sent a connection preface.
-    private struct PrefaceSentState: ConnectionStateWithRole, ConnectionStateWithConfiguration, MaySendFrames, HasLocalSettings, HasFlowControlWindows {
+    fileprivate struct PrefaceSentState: ConnectionStateWithRole, ConnectionStateWithConfiguration, MaySendFrames, HasLocalSettings, HasFlowControlWindows {
         let role: ConnectionRole
         var headerBlockValidation: ValidationState
         var contentLengthValidation: ValidationState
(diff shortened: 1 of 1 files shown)


Lukasa:
The unfortunate irony here is that the net performance gain of this patch in a more real-world benchmark is pretty small: about 1.5%. The reality is that in that benchmark we just aren't copying too much state around, so the CoWs don't hurt too badly.

Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
```diff
@@ -557,46 +561,67 @@ extension HTTP2ConnectionStateMachine {
 
         switch self.state {
         case .idle(let state):
-            var settingsState = HTTP2SettingsState(localState: true)
-            settingsState.emitSettings(settings)
-            self.state = .prefaceSent(.init(fromIdle: state, localSettings: settingsState))
+            self.avoidingStateMachineCoW { [] newState in
```
weissi: mind removing the left-over `[]` ?

Lukasa: It's there on purpose to avoid me ever accidentally capturing in this block.

weissi: @Lukasa that doesn't work. `[...]` has no effect on things that are not listed

weissi: I complain for years that Swift needs a mode where I can specify an exhaustive capture list, but still not there :(

weissi: the only thing `[` does is to create a by-value copy of all the listed things, optionally with renaming

Lukasa:
Oh god, I hope I don't get retained, I don't need a cow.

Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
Lukasa: Fixed. :grin:

weissi:
> Oh god, I hope I don't get retained, I don't need a cow.

It's fine, you just killed a lot of cows. Enjoy 🥩

weissi:
> more real-world benchmark is pretty small: about 1.5%.

I think in the real world under allocator pressure this could have a much better impact.

Lukasa:
Indeed, I'm still strongly in favour of removing it: it's just not a really heavy hitter.

PR #107, opened 2019-04-18
Avoid unnecessary heap allocation when parsing frame headers.

### LINKED ISSUE #105: Remove unnecessary heap allocation in frame parsing.
While I was debugging #104, I spotted that we have an unnecessary heap allocation in `ByteBuffer.readFrameHeader`:

https://github.com/apple/swift-nio-http2/blob/429862635ade915661dc6cad0a970ed0993b7c56/Sources/NIOHTTP2/HTTP2FrameParser.swift#L1166

This whole method should probably be rewritten with `readWithUnsafeReadableBytes` or by using some `readInteger` calls: we can benchmark the difference between those two implementations.

Lukasa:
Motivation:

As noted in #105, I spotted that we were heap-allocating an array
when parsing the frame header. That heap allocation isn't necessary, we
can just write directly into an Int, which should save us some heap spray.

Modifications:

- Rewrote code to avoid heap allocating array and instead to use Int.
- Elided some bounds checks by using an unsafe pointer, since I was
    already here and it saved about 1% performance in a hot loop.

Result:

Fewer heap allocations on the hot code path.
Resolves #105

### DIFF
--- a/Sources/NIOHTTP2/HTTP2FrameParser.swift
+++ b/Sources/NIOHTTP2/HTTP2FrameParser.swift
@@ -1159,17 +1159,17 @@ fileprivate struct FrameHeader {
 
 fileprivate extension ByteBuffer {
     mutating func readFrameHeader() -> FrameHeader? {
-        guard self.readableBytes >= 9 else {
-            return nil
+        let saveSelf = self
+        guard let lenHigh = self.readInteger(as: UInt16.self),
+            let lenLow = self.readInteger(as: UInt8.self),
+            let type = self.readInteger(as: UInt8.self),
+            let flags = self.readInteger(as: UInt8.self),
+            let rawStreamID = self.readInteger(as: UInt32.self) else {
+                self = saveSelf
+                return nil
         }
 
-        let lenBytes: [UInt8] = self.readBytes(length: 3)!
-        let len = Int(lenBytes[0]) << 16 | Int(lenBytes[1]) << 8 | Int(lenBytes[2])
-        let type: UInt8 = self.readInteger()!
-        let flags: UInt8 = self.readInteger()!
-        let rawStreamID: UInt32 = self.readInteger()!
-
-        return FrameHeader(length: len, type: type, flags: FrameFlags(rawValue: flags), rawStreamID: rawStreamID)
+        return FrameHeader(length: Int(lenHigh) << 8 | Int(lenLow), type: type, flags: FrameFlags(rawValue: flags), rawStreamID: rawStreamID)
     }
 
     mutating func writePayloadSize(_ size: Int, at location: Int) {


Lukasa:
This has a fraction of one percent of a performance improvement in the real-world benchmark I'm running: just not really noticeable. Still worth avoiding the allocation, but just worth keeping it in perspective.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -1163,13 +1163,14 @@ fileprivate extension ByteBuffer {
             return nil
         }
 
-        let lenBytes: [UInt8] = self.readBytes(length: 3)!
```
weissi: I think you should keep the old code but replace

```
let lenBytes: [UInt8] = self.readBytes(length: 3)!
```

with

```
let lenHigh = self.readInteger(as: UInt16.self)
let lenLow = self.readInteger(as: UInt8.self)
let len = UInt32(lenHigh) << 16 | UInt32(lenLow)
```

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -1163,13 +1163,14 @@ fileprivate extension ByteBuffer {
             return nil
```
weissi: I don't really like the pre-calculation up here ^^^ and the `!`s down there. I think this should be replaced with

```swift
let saveSelf = self
guard let lenHigh = self.readInteger(as: UInt16.self),
      let lenLow = self.readInteger(as: UInt8.self),
      let lenLow = self.readInteger(as: UInt8.self),
      let type = self.readInteger(as: UInt8.self),
      let flags = self.readInteger(as: UInt8.self),
      let streamID = self.readInteger(as: UInt32.self) else {
    self = saveSelf
    return nil
}

return FrameHeader(...)
```

Lukasa: I disagree.

Specifically, the form you proposed does repeated and unnecessary bounds checking, incurring a bunch of potentially unnecessary pointer dereferences. Done maliciously it substantially increases the work needed to parse a frame header by forcing a quadratic behaviour when parsing frame headers.

weissi: There are no pointer dereferences.

Lukasa: Sure there are, reading data from the `ByteBuffer` is a pointer dereference.

weissi: the bounds checks are integer comparisons. I agree that we should add a way that can hoist the bounds checks but we don't have one today. Going into unsafe-land is not a good excuse to reduce a constant factor

weissi: For the future, I want something like

```
self.read( integer(UInt16.self) <*> integer(UInt8.self) <*> integer(UInt8.self) <*> integer(UInt8.self) )
```

that does the bounds check only once. But we're not there yet.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -1159,17 +1159,17 @@ fileprivate struct FrameHeader {
...
         }
 
-        let lenBytes: [UInt8] = self.readBytes(length: 3)!
-        let len = Int(lenBytes[0]) << 16 | Int(lenBytes[1]) << 8 | Int(lenBytes[2])
-        let type: UInt8 = self.readInteger()!
-        let flags: UInt8 = self.readInteger()!
-        let rawStreamID: UInt32 = self.readInteger()!
-
-        return FrameHeader(length: len, type: type, flags: FrameFlags(rawValue: flags), rawStreamID: rawStreamID)
+        return FrameHeader(length: Int(lenHigh) << 8 | Int(lenLow), type: type, flags: FrameFlags(rawValue: flags), rawStreamID: rawStreamID)
```
weissi: aren't we overwriting 8 bits here?

Lukasa: No: the UInt16 is converted to an `Int` and shifted left 8 bits. Then the UInt8 is converted to an Int and or'ed with it, fitting into the empty space from the previous left shift.

weissi: @Lukasa oops, thanks, brain fart :D

PR #109, opened 2019-04-29
Ignore WINDOW_UPDATE on closed streams.

glbrntt:
Motivation:

RFC 7540 Section 6.9: "WINDOW_UPDATE can be sent by a peer that has sent
a frame bearing the END_STREAM flag.  This means that a receiver could
receive a WINDOW_UPDATE frame on a "half-closed (remote)" or "closed"
stream. A receiver MUST NOT treat this as an error (see Section 5.1)."

Previously WINDOW_UPDATE frames received in this state would be treated
as an error.

Modifications:

- Allow frames to be ignored if they are for a stream which has been
  closed.
- Add a test to verify WINDOW_UPDATE on a closed stream.
- Update tests which relied on a stream error in this case.

Result:

More compliant code.

### DIFF
--- a/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
+++ b/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
@@ -168,14 +168,16 @@ struct ConnectionStreamState {
     /// - parameters:
     ///     - streamID: The ID of the stream to modify.
     ///     - ignoreRecentlyReset: Whether a recently reset stream should be ignored. Should be set to `true` when receiving frames.
+    ///     - ignoreClosed: Whether a closed stream should be ignored. Should be set to `true` when receiving window update frames.
     ///     - modifier: A block that will be invoked to modify the stream state, if present.
     /// - returns: The result of the state modification, as well as any state change that occurred to the stream.
     @inline(__always)
     mutating func modifyStreamState(streamID: HTTP2StreamID,
                                     ignoreRecentlyReset: Bool,
+                                    ignoreClosed: Bool = false,
                                     _ modifier: (inout HTTP2StreamStateMachine) -> StateMachineResultWithStreamEffect) -> StateMachineResultWithStreamEffect {
         guard let result = self.activeStreams[streamID].autoClosingTransform(modifier) else {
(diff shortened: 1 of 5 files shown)


Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
```diff
@@ -168,14 +168,16 @@ struct ConnectionStreamState {
     /// - parameters:
     ///     - streamID: The ID of the stream to modify.
     ///     - ignoreRecentlyReset: Whether a recently reset stream should be ignored. Should be set to `true` when receiving frames.
+    ///     - ignoreClosed: Whether a closed stream should be ignored. Should be set to `true` when receiving window update frames.
     ///     - modifier: A block that will be invoked to modify the stream state, if present.
     /// - returns: The result of the state modification, as well as any state change that occurred to the stream.
     @inline(__always)
     mutating func modifyStreamState(streamID: HTTP2StreamID,
                                     ignoreRecentlyReset: Bool,
+                                    ignoreClosed: Bool,
```
Lukasa: Can we give `ignoreClosed` a default argument and remove it everywhere other than `receiveWindowUpdate`?

glbrntt: Done!

Sources/NIOHTTP2/ConnectionStateMachine/FrameReceivingStates/ReceivingPushPromiseState.swift
```diff
@@ -55,7 +55,7 @@ extension ReceivingPushPromiseState {
             try self.streamState.createRemotelyPushedStream(streamID: childStreamID,
                                                             remoteInitialWindowSize: self.remoteInitialWindowSize)
 
-           let result = self.streamState.modifyStreamState(streamID: originalStreamID, ignoreRecentlyReset: true) {
+            let result = self.streamState.modifyStreamState(streamID: originalStreamID, ignoreRecentlyReset: true) {
```
Lukasa: Mind fixing this indentation?

glbrntt: It was under-indented before, this fixes it!

PR #111, opened 2019-05-03
Avoid CoW when appending inbound frame bytes.

Lukasa:
Motivation:

This is basically a similar issue to #104: we have a CoW data structure (ByteBuffer) in some of
these states, and when calling append on them they force an unnecessary CoW. We can reproduce the
fix from that patch, which is to define an empty state and swap into it before we run the append.
We also define a helper that ensures that we don't screw this up.

Modifications:

- Added new appending state to HTTP2FrameDecoder
- Added helpers for avoiding a CoW
- Avoided CoW in HTTP2FrameDecoder.

Result:

Fewer unnecessary copies


Lukasa:
This is another roughly 1% perf gain on the h2load benchmark.

PR #112, opened 2019-05-03
Reduce intermediate header copies.

Lukasa:
Motivation:

We were copying HTTP headers around quite a lot when transforming from HTTP/1 to HTTP/2 and
back. Due to the different underlying representations we cannot get away from needing to do
one copy, as the backing array storage is of a different type, but we were frequently incurring
CoWs as well that were just totally unnecessary.

Modifications:

- Refactored the header code to work more confidently on HPACKHeaders.
- Removed some mutating functions that triggered unnecessary CoWs.
- Improved HPACKHeaders(httpHeaders:) initializer to avoid unnecessary intermediate array
- Improved HTTPHeaders(requestHead:protocolString:) to avoid CoWs and reallocations
- Improved HTTPHeaders(responseHead:) to avoid CoWs and reallocations

Result:

Fewer heap allocations


Lukasa:
For those keeping track, the perf gain here is better than some of the previous patches: about 4%.

PR #115, opened 2019-05-15
Avoid 8 heap allocations on child channel creation

Lukasa:
Motivation:

When users create a child channel and want to get hold of it,
they will pass a promise to createStreamChannel. This promise
will force a call to map{} and a call to cascade(), both of
which heap allocate. We don't need to have these heap allocations,
and they're easily removed, so we should.

Modifications:

Manually cascade the result.

Result:

Slightly fewer heap allocations in some workloads.


weissi:
@Lukasa wrong title, it actually saves 8 allocations, not 2 👊. Good job!

Issue #87, opened 2019-04-08
Quadratic runtime when sending many requests


MrMage:
Using NIOHTTP2 version 1.0.1.

Caveat: We might be doing something wrong in SwiftGRPC and/or this might be fixed by @Lukasa's recent changes, but I figured you guys might want to take a look.

When I run https://github.com/grpc/grpc-swift/blob/da1a59c919555367abfd6e2ffb285a175ffa11c0/Tests/SwiftGRPCNIOTests/NIOFunctionalTests.swift#L79 _in Release mode_ with `numberOfRequests = 2_000`:

```
Test Suite 'Selected tests' started at 2019-04-08 14:11:22.901
Test Suite 'SwiftGRPCNIOTests.xctest' started at 2019-04-08 14:11:22.901
Test Suite 'NIOFunctionalTestsInsecureTransport' started at 2019-04-08 14:11:22.901
Test Case '-[SwiftGRPCNIOTests.NIOFunctionalTestsInsecureTransport testUnaryLotsOfRequests]' started.
1000 requests sent so far, elapsed time: 0.32004
total time to send 2000 requests: 0.852957
total time to receive 2000 responses: 2.479385
Test Case '-[SwiftGRPCNIOTests.NIOFunctionalTestsInsecureTransport testUnaryLotsOfRequests]' passed (1.644 seconds).
Test Suite 'NIOFunctionalTestsInsecureTransport' passed at 2019-04-08 14:11:24.545.
	 Executed 1 test, with 0 failures (0 unexpected) in 1.644 (1.644) seconds
Test Suite 'SwiftGRPCNIOTests.xctest' passed at 2019-04-08 14:11:24.545.
	 Executed 1 test, with 0 failures (0 unexpected) in 1.644 (1.644) seconds
Test Suite 'Selected tests' passed at 2019-04-08 14:11:24.546.
	 Executed 1 test, with 0 failures (0 unexpected) in 1.644 (1.645) seconds
Program ended with exit code: 0
```

When run with `numberOfRequests = 4_000`:

```
Test Suite 'Selected tests' started at 2019-04-08 14:11:36.312
Test Suite 'SwiftGRPCNIOTests.xctest' started at 2019-04-08 14:11:36.312
Test Suite 'NIOFunctionalTestsInsecureTransport' started at 2019-04-08 14:11:36.312
[...]

PR #117, opened 2019-05-15
HTTP/2 peformance tests

weissi:
Motivation:

NIO wants to be fast, we should have perf tests everywhere.

Modifications:

Start with one HTTP/2 perf test.

Result:

more perf tests.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -18,6 +18,7 @@ import PackageDescription
 let package = Package(
     name: "swift-nio-http2",
     products: [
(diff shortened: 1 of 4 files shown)


weissi:
CC @glbrntt if you don't mind checking if that's roughly what you have been testing with gRPC on top.

glbrntt:
@weissi more or less; the only difference to note is that in the gRPC tests the unary requests are sent in batches; i.e. fire 100 requests, wait, repeat. This is due to: https://github.com/apple/swift-nio-http2/issues/87#issuecomment-483542401

PR #116, opened 2019-05-15
Replace Future with a function call.

Lukasa:
Motivation:

Sometimes when you have a hammer everything looks like a nail.
In this case, my hammer was Futures. But we don't need to use a
Future here: these two classes are tightly coupled, and so we can
safely just call from one to the other.

A side note: while this appears to create a reference cycle, the
act of calling the function in question will actually break that
cycle. So it's all good!

Modifications:

- Remove a future callback.
- Have the stream channel hold a reference to the multiplexer.

Result:

Fewer heap allocations.

PR #126, opened 2019-05-17
Remove indirection between channel and multiplexer

Lukasa:
Motivation:

Previously the HTTP2StreamChannel forwarded all frames to the parent
Channel by way of the abstract Channel existential API. While this was
straightforward, it added unnecessary overhead due to the extra pipeline
steps and the extra indirection of the types providing an inlining
barrier.

Given that in #116 we gave the HTTP2StreamChannel a reference to its
parent multiplexer, we may as well take advantage of this and allow
direct calls from the stream channel into the multiplexer.

Modifications:

Avoided calls to parent.write and parent.flush.
Replaced them with a new interface.

Result:

Slightly better performance.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
@@ -30,6 +30,7 @@ public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboun
     private var streams: [HTTP2StreamID: HTTP2StreamChannel] = [:]
     private let inboundStreamStateInitializer: ((Channel, HTTP2StreamID) -> EventLoopFuture<Void>)?
     private let channel: Channel
+    private var context: ChannelHandlerContext!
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
```diff
@@ -30,13 +30,15 @@ public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboun
     private var streams: [HTTP2StreamID: HTTP2StreamChannel] = [:]
     private let inboundStreamStateInitializer: ((Channel, HTTP2StreamID) -> EventLoopFuture<Void>)?
     private let channel: Channel
+    private var context: ChannelHandlerContext!
```
weissi: needs to be nilled out in `handlerRemoved`

Lukasa: Addressed.

PR #132, opened 2019-05-24
Coalesce flushes during the read cycle.

Lukasa:
Motivation:

The HTTP2StreamMultiplexer currently passes frames from child streams
through to the rest of the pipeline, and does the same with flushes.
Because HTTP2 is multiplexed, it is highly likely that in any
channelRead cycle multiple child channels will make sufficient progress
to want to flush their results. This means we can end up with multiple,
redundant, flush calls.

It would be better in general to try to reduce the amount of flushing
necessary to no more than once per channelRead cycle.

Modifications:

- Keep track of the channelRead cycle and where we are in it.
- Avoid flushing when we can expect a channelReadComplete.
- Flush on channelReadComplete if necessary.

Result:

Fewer flushes, better performance under high read load.


Lukasa:
Transient test failure due to HPACK perf test.

Lukasa:

Lukasa:

PR #142, opened 2019-07-17
Support child channel writability management.

Lukasa:
Motivation:

It's important to ensure that child channel users can observe the
writability state of a child channel in order to avoid ballooning data
in memory. Up until now we haven't supported writability notifications,
but we really should.

Modifications:

- Added a flow controller to manage writability.
- Managed data flow through promises.
- Lots of tests.

Result:

Child channels will provide a much better signal as to whether they are
writable.


Lukasa:
One of my concerns with this patch is that it comes with quite a hefty performance cost: a throughput based test showed a reduction in req/s from 10905.00 to 9853.33. This represents an approximately 10% performance loss.

Admittedly, this test fairly aggressively tests frame throughput: it uses only HEADERS and zero-length DATA frames, which means that the cost of managing the promises is more expensive than it would be in a production server. The cost would also be lower in cases where the writes already have promises attached to them.

This functionality is definitely more important than the performance we see here, but we do need to investigate whether we can do something that performs better.

Lukasa:

docker/docker-compose.1804.50.yaml
```diff
@@ -16,9 +16,9 @@ services:
   integration-tests:
     image: swift-nio-http2:18.04-5.0
     environment:
-      - MAX_ALLOCS_ALLOWED_create_client_stream_channel=70010
+      - MAX_ALLOCS_ALLOWED_create_client_stream_channel=72010
```
Lukasa: This increase (2 allocs per test run) is for the atomic the child channel has: one for the `Atomic` `class`, and one for the heap allocation for the atomic itself.

docker/docker-compose.1804.50.yaml
```diff
@@ -16,9 +16,9 @@ services:
...
       - MAX_ALLOCS_ALLOWED_hpack_decoding=5050
-      - MAX_ALLOCS_ALLOWED_client_server_request_response=360000
+      - MAX_ALLOCS_ALLOWED_client_server_request_response=374000
```
Lukasa: This increase (14 per test run, approximately) is for channel atomics and the new futures and their callbacks.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -338,7 +361,28 @@ final class HTTP2StreamChannel: Channel, ChannelCore {
...
+        promise.futureResult.hop(to: self.eventLoop).whenComplete { (_: Result<Void, Error>) in
+            if case .changed(newValue: let value) = self.writabilityManager.wroteBytes(writeSize) {
+                self._isWritable.store(value)
+                self.pipeline.fireChannelWritabilityChanged()
+            }
+        }
         self.pendingWrites.append((frame, promise))
+
+        // Ok, we can make an outcall now, which means we can safely deal with the flow control.
+        if case .changed(newValue: let value) = self.writabilityManager.bufferedBytes(writeSize) {
```
normanmaurer: nit: it seems like this code block and the one in the callback of the future is the same, so just make it a private function (can even be just declared in this function itself.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -599,4 +643,59 @@ internal extension HTTP2StreamChannel {
...
+            // it's not worth having a heuristic.
+            let paddingBytes = p.paddingBytes.map { $0 + 1 } ?? 0
+            return paddingBytes + 9
+        case .rstStream:
+            return 9 + 4
+        case .windowUpdate:
+            return 9 + 4
+        default:
+            // Unknown or unexpected control frame: say 9 bytes.
+            return 9
```
normanmaurer: as we use 9 here multiple times I think it would be good to make this a static let or something like this and give it a meaningful name (and so also give a hind where this is coming from, like the "frame header" size).

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
```diff
@@ -172,6 +176,16 @@ public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboun
         context.fireUserInboundEventTriggered(event)
     }
 
+    public func channelWritabilityChanged(context: ChannelHandlerContext) {
+        let newWritability = context.channel.isWritable
+
+        for channel in self.streams.values {
+            channel.parentChannelWritabilityChanged(newValue: newWritability)
```
normanmaurer: I think it is not save to use the `newWritability` here as this value may change in between. General speaking this method may even re-entrance and so it would lead to out-of-order notifies when using `newWritability` here.

Lukasa:
Thanks @normanmaurer, feedback addressed.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -599,4 +646,61 @@ internal extension HTTP2StreamChannel {
...
+    /// hard code fixed lengths for fixed length frames, use a calculated length for
+    /// variable length frames, and just ignore encoded headers because it's not worth doing a better
+    /// job.
+    fileprivate var bufferBytes: Int {
+        let frameHeaderSize = 9
+
+        switch self.payload {
+        case .data(let d):
+            let paddingBytes = d.paddingBytes.map { $0 + 1 } ?? 0
+            return d.data.readableBytes + paddingBytes + frameHeaderSize
```
normanmaurer: I am a bit concerned that data readableBytes could change over the life-time. Wouldn't it be better to capture it when the frame is created and always return the same ?

Lukasa: > I am a bit concerned that data readableBytes could change over the life-time.

Over the lifetime of what?

normanmaurer: lifetime of the frame... that said I just remembered we are not in java-land here and so it should be fine (as it is a struct).

Lukasa: Indeed, and as this is Java land the only risk is that this frame could be lengthen by the HTTP/2 implementation itself. As it cannot, there isn't a risk here, and we should be good.

PR #148, opened 2019-08-13
fix allocation counter limits

weissi:
Motivation:

We regressed 4 allocations in the security update
https://github.com/apple/swift-nio-http2/security/advisories/GHSA-jchv-x857-q8fq
. Four allocations are a great tradeoff for a number of security
vulnerabilities fixed but we need to raise the limits accordingly.

Modifications:

Raise the limits to the new limit so tests can pass again.

Result:

PRs can pass again and master should build again.


Lukasa:
Thanks!

PR #153, opened 2019-09-02
accept more valid-in-HTTP/1 but invalid-in-H2 messages

### LINKED ISSUE #152: HTTP2to1ServerCodec: should we be more tolerant with the headers the user provides
- we should probably lower-case the header names automatically
- we should probably just ignore `connection: keep-alive/close`
- we should probably drop `transfer-encoding: chunked`

weissi:
Motivation:

HTTP/2 is mostly just HTTP(/1) semantics with streams on a new wire
protocol. However, there are certain things that are legal in HTTP/1 but
which are illegal in HTTP/2.

Modification:

Accept more of those messages and transform them into their HTTP/2
equivalent.

Result:

- Easier to write HTTP/2 clients & servers in NIO.
- fixes #152

### DIFF
--- a/Sources/NIOHPACK/HPACKHeader.swift
+++ b/Sources/NIOHPACK/HPACKHeader.swift
@@ -19,10 +19,31 @@ import NIOHTTP1
 /// information.
 public struct HPACKHeaders {
     internal var headers: [HPACKHeader]
-    
+
+    // see 8.1.2.2. Connection-Specific Header Fields in RFC 7540
+    static let illegalHeaders: [String] = ["connection", "keep-alive", "proxy-connection",
+                                           "transfer-encoding", "upgrade"]
+
+    /// Constructor that can be used to map between `HTTPHeaders` and `HPACKHeaders` types.
+    public init(httpHeaders: HTTPHeaders, normalizeHTTPHeaders: Bool) {
+        if normalizeHTTPHeaders {
+            self.headers = httpHeaders.map { HPACKHeader(name: $0.name.lowercased(), value: $0.value) }
+
+            let connectionHeaderValue = httpHeaders[canonicalForm: "connection"]
+            if !connectionHeaderValue.isEmpty {
+                self.headers.removeAll { header in
+                    return HPACKHeaders.illegalHeaders.contains(header.name) ||
+                        connectionHeaderValue.contains(header.name[...])
+                }
+            }
+        } else {
+            self.headers = httpHeaders.map { HPACKHeader(name: $0.name, value: $0.value) }
+        }
+    }
+
     /// Constructor that can be used to map between `HTTPHeaders` and `HPACKHeaders` types.
     public init(httpHeaders: HTTPHeaders) {
-        self.headers = httpHeaders.map { HPACKHeader(name: $0.name, value: $0.value) }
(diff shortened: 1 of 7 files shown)


MrMage:
I think https://github.com/apple/swift-nio-http2/pull/95#issuecomment-482685538 should also be considered in this context, there _could_ be value in rejecting these headers by default to encourage and police spec-conformant implementations by default.

weissi:
> think [#95 (comment)](https://github.com/apple/swift-nio-http2/pull/95#issuecomment-482685538) should also be considered in this context, there _could_ be value in rejecting these headers by default to encourage and police spec-conformant implementations by default.

If you set `normalizeHeaders: false`, you'll get the rejection. Also the RFC [states](https://tools.ietf.org/html/rfc7540#section-8.1.2.2) that

>    This means that an intermediary transforming an HTTP/1.x message to
   HTTP/2 will need to remove any header fields nominated by the
   Connection header field, along with the Connection header field
   itself.  Such intermediaries SHOULD also remove other connection-
   specific header fields, such as Keep-Alive, Proxy-Connection,
   Transfer-Encoding, and Upgrade, even if they are not nominated by the
   Connection header field.

And you could consider the 2to1/1to2 handlers such an intermediary, right?

MrMage:
Sorry, I missed that this was only about the transcoder. I agree that this is the correct behavior for the transcoder.

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -19,10 +19,28 @@ import NIOHTTP1
 /// information.
 public struct HPACKHeaders {
     internal var headers: [HPACKHeader]
-    
+
+    // see 8.1.2.2. Connection-Specific Header Fields in RFC 7540
+    let illegalHeaders: [String] = ["connection", "keep-alive", "proxy-connection", "transfer-encoding", "upgrade"]
```
Lukasa: I think strictly this is a `static`, though I don't know if we gain much by writing that down.

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -19,10 +19,28 @@ import NIOHTTP1
...
+
+    /// Constructor that can be used to map between `HTTPHeaders` and `HPACKHeaders` types.
+    public init(httpHeaders: HTTPHeaders, normalizeHTTPHeaders: Bool) {
+        self.headers = httpHeaders.map { HPACKHeader(name: normalizeHTTPHeaders ? $0.name.lowercased() : $0.name,
+                                                     value: $0.value) }
```
Lukasa: Rather than branch on `normalizeHTTPHeaders` every time it may be better to have a single `if` that sets up the header block differently.

Lukasa: Oh hey, we already have that `if` block below, so let's do that. That also allows us to rewrite this code to make only a single pass over the headers map, instead of doing two.

Sources/NIOHPACK/HPACKHeader.swift
weissi: done, thanks

Sources/NIOHPACK/HPACKHeader.swift
weissi: done, thanks

PR #158, opened 2019-09-24
swift 5.1 docker setup

tomerd:
motivation: support swift 5.1

changes:
* add docker compose file for swift 5.1, will add CI job once merged
* update swiftformat version to 0.40.12
* use more standard dockerfile setup

### DIFF
--- a/docker/docker-compose.1804.51.yaml
+++ b/docker/docker-compose.1804.51.yaml
@@ -1,0 +1,37 @@
+version: "3"
+
+services:
+
+  runtime-setup:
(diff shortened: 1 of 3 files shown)


docker/docker-compose.1804.51.yaml
```diff
@@ -0,0 +1,37 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: swift-nio-http2:18.04-5.1
```
weissi: this should be `16.04` and `xenial` so that we maximise what we test for.

PR #170, opened 2019-11-12
Force slow path header comparison to not inline.

Lukasa:
Motivation:

We have a slow path in compareCaseInsensitiveASCIIBytes. On the server
this path will never be hit as to hit it requires having a
non-contiguous String, so we really really want the compiler to use all
its inlining smarts on the fast path instead. Currently it does, but as
Sequence.elementsEqual is inlinable we need to be sure that doesn't
change.

Modifications:

Force the slow path not to be inlined.

Result:

Guarantees about good inlining.


Lukasa:

Issue #151, opened 2019-08-27
Proposed change to BadStreamStateTransition for debuggability


nealsid:
I'm just beginning to dig into the HTTP/2 and Swift-NIO code and wanted to get some feedback on a proposal to modify BadStreamStateTransition to have from/to states to improve debuggability when it gets logged (and maybe in the future, stack traces when it gets created)

A few questions:

- The HTTP2StateMachine.State enum is private, and comments indicate that it's so people don't erroneously set the state directly.  Isn't the right place to do this in setting the member state variable to private (which it already is)?   If that works I could make it internal and move the state enum outside of HTTP2StateMachine

- I'm seeing some comments that (to me) don't agree with the RFC. I am no HTTP expert by any means and I'm fairly new to Swift, as well, so I wouldn't be surprised if I"m wrong here, but on line 379, it indicates sending HEADERS on an idle stream isn't permitted, but the state transition diagram and the RFC both seem to indicate that it should transition a stream to open.  

Apologies for the basic questions, like I mentioned I'm new. Thanks in advance for your help.

Neal

Lukasa:
This seems like a good idea, I’m in favour of it! In answer to your questions:

> The HTTP2StateMachine.State enum is private, and comments indicate that it's so people don't erroneously set the state directly. Isn't the right place to do this in setting the member state variable to private (which it already is)?

At the time there was no good reason to expose the type to other parts of the code. You have found one, so the status quo should change. By all means change the visibility. Please leave it nested inside the state machine type though; we don’t gain anything by hoisting it out, and it helps avoid the type seeming to be generally useful. 

> on line 379, it indicates sending HEADERS on an idle stream isn't permitted, but the state transition diagram and the RFC both seem to indicate that it should transition a stream to open.

A server is not permitted to send a HEADERS frame on an idle stream. RFC 7540 does say this, just not in the connection state machine diagram. 

If you read the comment in the code next to our reproduction of the state machine diagram you’ll notice that it says that the RFC 7540 diagram is so incomplete as to be actively misleading. In this case, our (much more detailed) state machine is right: RFC 7540’s diagram is wrong.

Lukasa:
> One scenario that I was thinking about was exposing ]information about the streams in a server built off this code. If someone wants to drill down into the various open streams by connection state, is it possible to select all, say, .idle streams independent of the associated data?

[...]

PR #175, opened 2019-11-30
Issue #151 Update BadStreamStateTransition to have stream state

blueeor:
swift-nio-http2 commit hash: 85d5cea3f9134a242ad4baaa3a4af1ff080a3f81

Context:
Address the issue #151. Update BadStreamStateTransition to have a property with the stream state when the error happened. This public steam state is a enum without any associated values.
Also update some unit test functions to return an optional error. 

% swift --version
Apple Swift version 5.1.2 (swiftlang-1100.0.278 clang-1100.0.33.9)
Target: x86_64-apple-darwin19.0.0

### DIFF
--- a/Sources/NIOHTTP2/HTTP2Error.swift
+++ b/Sources/NIOHTTP2/HTTP2Error.swift
@@ -61,7 +61,11 @@ public enum NIOHTTP2Errors {
 
     /// A stream state transition was attempted that was not valid.
     public struct BadStreamStateTransition: NIOHTTP2Error {
-        public init() { }
+        let fromState: NIOHTTP2StreamState
+
+        init(from state: NIOHTTP2StreamState) {
+            fromState = state
+        }
     }
 
     /// An attempt was made to change the flow control window size, either via
--- a/Sources/NIOHTTP2/HTTP2Error.swift
+++ b/Sources/NIOHTTP2/HTTP2Error.swift
@@ -60,8 +60,12 @@ public enum NIOHTTP2Errors {
     }
 
     /// A stream state transition was attempted that was not valid.
-    public struct BadStreamStateTransition: NIOHTTP2Error {
+    public struct BadStreamStateTransition: NIOHTTP2Error, CustomStringConvertible {
         let fromState: NIOHTTP2StreamState
+        
+        public var description: String {
+            return "BadStreamStateTransition in state \(fromState)"
+        }
 
         init(from state: NIOHTTP2StreamState) {
             fromState = state
--- a/Sources/NIOHTTP2/HTTP2Error.swift
+++ b/Sources/NIOHTTP2/HTTP2Error.swift
@@ -61,14 +61,19 @@ public enum NIOHTTP2Errors {
 
     /// A stream state transition was attempted that was not valid.
     public struct BadStreamStateTransition: NIOHTTP2Error, CustomStringConvertible {
-        let fromState: NIOHTTP2StreamState
-        
+        public let fromState: NIOHTTP2StreamState?
+
         public var description: String {
-            return "BadStreamStateTransition in state \(fromState)"
+            let stateName = fromState != nil ? "\(fromState!)" : "unknown state"
+            return "BadStreamStateTransition in state \(stateName)"
         }
 
         init(from state: NIOHTTP2StreamState) {
             fromState = state
+        }
+
+        public init() {
+            self.fromState = nil
         }
     }
 
--- a/Sources/NIOHTTP2/StreamStateMachine.swift
+++ b/Sources/NIOHTTP2/StreamStateMachine.swift
@@ -1023,8 +1023,7 @@ private extension HPACKHeaders {
     }
 }
 
-
-/// State of a HTTP2 Stream.
+/// A public state of a `HTTP2StreamStateMachine`. This copy mirrors `HTTP2StreamStateMachine.state` but without associated values.
 public enum NIOHTTP2StreamState {
     case idle
     case reservedRemote
--- a/Sources/NIOHTTP2/StreamStateMachine.swift
+++ b/Sources/NIOHTTP2/StreamStateMachine.swift
@@ -1024,7 +1024,7 @@ private extension HPACKHeaders {
 }
 
 /// A state of a `HTTP2StreamStateMachine`. This copy mirrors `HTTP2StreamStateMachine.state` but without associated values.
-enum NIOHTTP2StreamState {
+public enum NIOHTTP2StreamState {
     case idle
     case reservedRemote
     case reservedLocal
--- a/Sources/NIOHTTP2/StreamStateMachine.swift
+++ b/Sources/NIOHTTP2/StreamStateMachine.swift
@@ -96,7 +96,7 @@ import NIOHPACK
 /// Operations on the state machine are performed by calling specific functions corresponding to
 /// the operation that is about to occur.
 struct HTTP2StreamStateMachine {
-    private enum State {
+    fileprivate enum State {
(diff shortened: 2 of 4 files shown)


Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -61,7 +61,11 @@ public enum NIOHTTP2Errors {
 
     /// A stream state transition was attempted that was not valid.
     public struct BadStreamStateTransition: NIOHTTP2Error {
-        public init() { }
+        let fromState: NIOHTTP2StreamState
```
Lukasa: As this is currently only internal, users outside this package can't see it. We have two options here: the first is that we could make this public, the second is that we can conform this type to `CustomStringConvertible` and have the state be visible when printing this error. To begin with, I think I'd start with conforming to `CustomStringConvertible`.

Sources/NIOHTTP2/StreamStateMachine.swift
```diff
@@ -1023,3 +1022,35 @@ private extension HPACKHeaders {
         return self.first { $0.name == ":status" }?.value.first == "1"
     }
 }
+
+
+/// State of a HTTP2 Stream.
+public enum NIOHTTP2StreamState {
```
Lukasa: Can we elaborate on this doc string and explain what this type is for?

Sources/NIOHTTP2/HTTP2Error.swift
blueeor: I updated the code to conform to CustomStringConvertible.

Sources/NIOHTTP2/StreamStateMachine.swift
blueeor: I updated the doc string. How does it look now?

Lukasa: Looks good. Given the type is now used only internally, can we make it `internal`?

Tests/NIOHTTP2Tests/ConnectionStateMachineTests.swift
```diff
@@ -195,6 +217,50 @@ class ConnectionStateMachineTests: XCTestCase {
...
+
+        assertBadStreamStateTransition(type: .halfClosedLocalPeerActive, self.server.sendData(streamID: streamOne, contentLength: 0, flowControlledBytes: 0, isEndStreamSet: true))
+        self.server = savedServerState
+        assertBadStreamStateTransition(type: .halfClosedLocalPeerActive, self.server.sendPushPromise(originalStreamID: streamOne, childStreamID: streamTwo, headers: testPushPromiseHeaders))
+        self.server = savedServerState
+        assertBadStreamStateTransition(type: .halfClosedLocalPeerActive, self.server.receivePushPromise(originalStreamID: streamOne, childStreamID: streamTwo, headers: testPushPromiseHeaders))
+        self.server = savedServerState
+        assertBadStreamStateTransition(type: .halfClosedLocalPeerActive, self.server.sendHeaders(streamID: streamOne, headers: testPushPromiseHeaders, isEndStreamSet: false))
+        self.server = savedServerState
+    }
```
Lukasa: Mind adding a test for the `CustomStringConvertible` conformance? Just try to print it.

Sources/NIOHTTP2/StreamStateMachine.swift
blueeor: Will do, but from the original issue it's not clear to me if they would want access to the error from outside the library. I understand the wish to keep types internal or private unless there is an explicit need to be public.

Tests/NIOHTTP2Tests/ConnectionStateMachineTests.swift
blueeor: Done.

Sources/NIOHTTP2/StreamStateMachine.swift
Lukasa: I grant that, but right now they can't actually _see_ the error from outside the library because the error that is thrown holds the stream state in an `internal` property. For now I think leaving things internal is fine: we can always make them public later.

Lukasa:

Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -60,8 +60,16 @@ public enum NIOHTTP2Errors {
     }
 
     /// A stream state transition was attempted that was not valid.
-    public struct BadStreamStateTransition: NIOHTTP2Error {
-        public init() { }
```
Lukasa: This empty public `init` cannot be removed.

With that in mind, let's go back to making `NIOHTTP2StreamState` public, and add `public` visibility to `BadStreamStateTransition.fromState`. Then, `fromState` can be optional and we can default it to `nil` for backwards compatibility reasons.

blueeor: ok, I made those changes.

Sources/NIOHTTP2/StreamStateMachine.swift
```diff
@@ -1023,3 +1022,34 @@ private extension HPACKHeaders {
         return self.first { $0.name == ":status" }?.value.first == "1"
     }
 }
+
+/// A state of a `HTTP2StreamStateMachine`. This copy mirrors `HTTP2StreamStateMachine.state` but without associated values.
+public enum NIOHTTP2StreamState {
```
Lukasa: Ah, sorry, this should have occurred to me earlier: we need to make this an "extensible enum". To do that we actually need to define a wrapping struct, like this:

```swift
public struct NIOHTTP2StreamState: Hashable {
    private enum State {
        case idle
        case reservedRemote
        case reservedLocal
        case halfOpenLocalPeerIdle
        case halfOpenRemoteLocalIdle
        case fullyOpen
        case halfClosedLocalPeerIdle
        case halfClosedLocalPeerActive
        case halfClosedRemoteLocalIdle
        case halfClosedRemoteLocalActive
        case closed
    }

    private var state: State

    private init(state: State) {
        self.state = state
    }

    public static let idle = NIOHTTP2StreamState(state: .idle)
    public static let reservedRemote = NIOHTTP2StreamState(state: .reservedRemote)
    public static let reservedLocal = NIOHTTP2StreamState(state: .reservedLocal)
    public static let halfOpenLocalPeerIdle = NIOHTTP2StreamState(state: .halfOpenLocalPeerIdle)
    public static let halfOpenRemoteLocalIdle = NIOHTTP2StreamState(state: .halfOpenRemoteLocalIdle)
    public static let fullyOpen = NIOHTTP2StreamState(state: .fullyOpen)
    public static let halfClosedLocalPeerIdle = NIOHTTP2StreamState(state: .halfClosedLocalPeerIdle)
    public static let halfClosedLocalPeerActive = NIOHTTP2StreamState(state: .halfClosedLocalPeerActive)
    public static let halfClosedRemoteLocalIdle = NIOHTTP2StreamState(state: .halfClosedRemoteLocalIdle)
    public static let halfClosedRemoteLocalActive = NIOHTTP2StreamState(state: .halfClosedRemoteLocalActive)
    public static let closed = NIOHTTP2StreamState(state: .closed)
}
```

We can then keep the `fileprivate init` as-is.

The reasoning for doing this change is that in semantic versioning it is not possible to add new cases to an `enum` without shipping a new major version, because it's possible to exhaustively switch over enumerations. It is not possible to do so over `struct`s, so having a `struct` backed by an `enum` is an easy way to get a similar programming interface without the exhaustive switch support.

blueeor: I will work on this change.

blueeor: Done.

Lukasa:

PR #178, opened 2019-12-02
Add a subscript to HPACKHeaders to return the first value with a matc…

glbrntt:
…hing name

Motivation:

In some cases you only expect (or care about) a single value for a
header. However, getting this value without allocating an array or
traversing the whole array requres using first(where:) provided by
Sequence. This is not very ergonomic as it requires users to do the
case-insensitive comparison and then exctact the value from the
HPACKHeader. Additionally, users cannot use NIO HTTP/2s case insensitive
ASCII comparison.

Modifications:

- Provide a subscript on HPACKHeaders which returns the value from the
  first header whose name is a case insentive match of the subscript
  parameter.

Result:

- It's easier to get only the first header matching a given name

### DIFF
--- a/Sources/NIOHPACK/HPACKHeader.swift
+++ b/Sources/NIOHPACK/HPACKHeader.swift
@@ -161,8 +161,27 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral {
         
         return array
     }
-    
-    /// Checks if a header is present
+
+    /// Retrieves the first value for a given header field name from the block.
+    ///
+    /// This method uses case-insensitive comparisons for the header field name. It
+    /// does not return the first value from a maximally-decomposed list of the header fields,
+    /// but instead returns the first value from the original representation: that means
+    /// that a comma-separated header field list may contain more than one entry, some of
+    /// which contain commas and some do not. If you want a representation of the header fields
+    /// suitable for performing computation on, consider `getCanonicalForm`.
+    ///
+    /// - Parameter name: The header field name whose first value should be retrieved.
+    /// - Returns: The first value for the header field name.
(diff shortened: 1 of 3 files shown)


Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -161,8 +161,27 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral {
...
+    /// This method uses case-insensitive comparisons for the header field name. It
+    /// does not return the first value from a maximally-decomposed list of the header fields,
+    /// but instead returns the first value from the original representation: that means
+    /// that a comma-separated header field list may contain more than one entry, some of
+    /// which contain commas and some do not. If you want a representation of the header fields
+    /// suitable for performing computation on, consider `getCanonicalForm`.
+    ///
+    /// - Parameter name: The header field name whose first value should be retrieved.
+    /// - Returns: The first value for the header field name.
+    public subscript(first name: String) -> String? {
```
Lukasa: @weissi What are your thoughts about the spelling of this API? It sits a bit weirdly with me, but I don't know if it really matters.

weissi: I think we should do something more similar to what the stdlib does for [`first(where:)`](https://developer.apple.com/documentation/swift/array/1848165-first). What about `first(name: String)`?

Lukasa: That's fine by me: the only risk is that this type already has a `first(where:)` so that API may get lost in the autocorrect.

glbrntt: `first(name:)` sounds good to me too.

PR #185, opened 2020-02-18
Update build_podspec script & added SwiftNIOHPACK Pod

Jake-Prickett:
- `SwiftNIOHTTP2` has a dependency on internal target `SwiftNIOHPACK` so validating the podspec was failing
- Updated the script to allow for the generation of both `SwiftNIOHTTP2.podspec` & `SwiftNIOHPACK .podspec`
- In `NIOHTTP2StreamChannel.swift`, it imports `NIOConcurrencyHelpers` which was not added as a dependency explicitly. Added the dependency which allowed the successful validation of the `SwiftNIOHTTP2` Pod. 

### Cocoapods
[SwiftNIOHTTP2](http://cocoapods.org/pods/SwiftNIOHTTP2)
[SwiftNIOHPACK](http://cocoapods.org/pods/SwiftNIOHPACK)

Newly created Pods from Release 1.9.1 - still need to add the other maintainers as Pod owners!

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -31,7 +31,7 @@ let package = Package(
         .target(name: "NIOHTTP2PerformanceTester",
                 dependencies: ["NIOHTTP2"]),
         .target(name: "NIOHTTP2",
-            dependencies: ["NIO", "NIOHTTP1", "NIOTLS", "NIOHPACK"]),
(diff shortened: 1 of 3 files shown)


Lukasa:

PR #191, opened 2020-03-11
Exert backpressure via stream flow control windows

glbrntt:
Motivation:

When a window update event is received in the stream channel we notify
the inbound window manager. Doing so may emit a WINDOW_UPDATE frame
to indicate the remote may send us more DATA. However, this happens even
if there are inbound frames waiting to be delivered to the channel. This
may lead to the remote sending us more DATA which we have to buffer
until it is read.

Modifications:

- Update the inbound window manager to track the number of bytes
  buffered in the stream channel
- Account for buffered bytes when receiving a new window size event
- Possibly emit WINDOW_UPDATE after unbuffering inbound frames into the
  stream channel's pipeline

Result:

The timing of WINDOW_UPDATE frames is better aligned with when frames
are actually read in a stream.


Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -554,11 +554,19 @@ private extension HTTP2StreamChannel {
         self.pendingReads = CircularBuffer(initialCapacity: 0)
     }
 
-    /// DinitialCapacityreads to the channel.
+    /// Deliver all pending reads to the channel.
     private func deliverPendingReads() {
         assert(self.isActive)
         while self.pendingReads.count > 0 {
-            self.pipeline.fireChannelRead(NIOAny(self.pendingReads.removeFirst()))
+            let frame = self.pendingReads.removeFirst()
```
Lukasa: For minor CoW purposes it would be best if we calculated the flow controlled length _first_. This would allow us to pass the frame into `fireChannelRead` and no longer own the data.

glbrntt: Ah, good call!

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -554,11 +554,27 @@ private extension HTTP2StreamChannel {
...
+
+            let flowControlledFameSize: Int?
+            if case .data(let data) = frame.payload {
+                flowControlledFameSize = data.data.readableBytes
+            } else {
+                flowControlledFameSize = nil
```
Lukasa: We should not call this the flow-controlled frame size because it isn't: at this stage we don't know what the padding was. This doesn't actually _matter_ here, it's a good thing, but we should just say "dataLength" or something similar.

PR #193, opened 2020-03-16
Added helper to configure HTTP1 or H2 accordingly to the negotiated protocol

mariosangiorgio:
Motivation:

Applications might want to support H2 (when available) and HTTP1 (as a fallback).
The application logic should not change between the two versions of the protocol and the code to configure support for both the protocols is boilerplate and can be provided as an helper function.

Modifications:

This PR adds an helper function to configure HTTP1 or H2 pipelines according to the protocol negotiated by the TLS handler.

Result:

Applications that need to support clients using both HTTP1 and H2 can configure their channels by calling `configureCommonHTTPServerPipeline`, which should reduce the boilerplate they have to write.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
+++ b/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
@@ -151,13 +151,13 @@ extension Channel {
         return self.pipeline.addHandler(alpnHandler)
     }
 
-    /// Configures a `ChannelPipeline` to speak either HTTP or HTTP/2 accordingly to what can be negotiated with the client.
+    /// Configures a `ChannelPipeline` to speak either HTTP or HTTP/2 according to what can be negotiated with the client.
     ///
     /// This helper takes care of configuring the server pipeline such that it negotiates whether to
-    /// use HTTP/1.1 or HTTP/2. Once the protocol to use for the channel has been negotitated, the
+    /// use HTTP/1.1 or HTTP/2. Once the protocol to use for the channel has been negotiated, the
     /// provided callback will configure the application-specific handlers in a protocol-agnostic way.
     ///
-    /// This function does't configure the TLS handler. Callers of this function need to add a TLS
+    /// This function doesn't configure the TLS handler. Callers of this function need to add a TLS
     /// handler appropriately configured to perform protocol negotiation.
     ///
     /// - parameters:
--- a/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
+++ b/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
@@ -56,6 +56,7 @@ public extension ChannelPipeline {
     ///         negotiated, or if no protocol was negotiated. Must return a future that completes when the
     ///         pipeline has been fully mutated.
     /// - returns: An `EventLoopFuture<Void>` that completes when the pipeline is ready to negotiate.
+    @available(*, deprecated, message: "Please use Channel.configureHTTP2SecureUpgrade(h2ChannelConfigurator, http1ChannelConfigurator)")
     func configureHTTP2SecureUpgrade(h2PipelineConfigurator: @escaping (ChannelPipeline) -> EventLoopFuture<Void>,
                                      http1PipelineConfigurator: @escaping (ChannelPipeline) -> EventLoopFuture<Void>) -> EventLoopFuture<Void> {
         let alpnHandler = ApplicationProtocolNegotiationHandler { result in
@@ -74,6 +75,14 @@ public extension ChannelPipeline {
         }
 
         return self.addHandler(alpnHandler)
+    }
+}
+
+private final class ErrorHandler: ChannelInboundHandler {
+    typealias InboundIn = Never
+
+    func errorCaught(context: ChannelHandlerContext, error: Error) {
+        context.close(promise: nil)
     }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -56,6 +56,7 @@ public extension ChannelPipeline {
     ///         negotiated, or if no protocol was negotiated. Must return a future that completes when the
     ///         pipeline has been fully mutated.
     /// - returns: An `EventLoopFuture<Void>` that completes when the pipeline is ready to negotiate.
+    @available(*, deprecated, message: "Please use Channel.configureHTTP2SecureUpgrade(h2ChannelConfigurator, http1ChannelConfigurator)")
```
Lukasa:
```suggestion
    @available(*, deprecated, renamed: "Channel.configureHTTP2SecureUpgrade(h2ChannelConfigurator:http1ChannelConfigurator:)")
```

mariosangiorgio: As discussed with @Lukasa, I'm deprecating `ChannelPipeline.configureHTTP2SecureUpgrade` and replacing it `Channel.configureHTTP2SecureUpgrade`.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -77,6 +78,14 @@ public extension ChannelPipeline {
     }
 }
 
+private final class ErrorHandler: ChannelInboundHandler {
```
Lukasa: I think we shouldn't provide one of these ourselves: we should just make sure the user is able to put one where they need to.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -104,4 +113,84 @@ extension Channel {
 
         return self.pipeline.addHandlers(handlers, position: position).map { multiplexer }
     }
+
+    /// Configures a channel to perform a HTTP/2 secure upgrade.
```
mariosangiorgio: This is essentially a copy and paste with `configureHTTP2SecureUpgrade`. Things I changed:
- `h2ChannelConfigurator` and `http1ChannelConfigurator` names to reflect that they are configuring channels
- changed type `(ChannelPipeline) -> EventLoopFuture<Void>` to `(Channel) -> EventLoopFuture<Void>`

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -104,4 +113,84 @@ extension Channel {
...
+        let h2ChannelConfigurator = { (channel: Channel) -> EventLoopFuture<Void> in
+            channel.configureHTTP2Pipeline(mode: .server) { (streamChannel, streamID) -> EventLoopFuture<Void> in
+                streamChannel.pipeline.addHandler(HTTP2ToHTTP1ServerCodec(streamID: streamID)).flatMap { () -> EventLoopFuture<Void> in
+                    configurator(streamChannel)
+                }.flatMap { () -> EventLoopFuture<Void> in
+                    streamChannel.pipeline.addHandler(ErrorHandler())
+                }
+            }.flatMap { (_: HTTP2StreamMultiplexer) in
+                return channel.pipeline.addHandler(ErrorHandler())
+            }
```
Lukasa: This specific use of `ErrorHandler` is the tricky one. Let's add another closure that will be invoked, something like `h2ConnectionChannelConfigurator` that is invoked explicitly on the HTTP/2 connection channel.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -104,4 +113,84 @@ extension Channel {
...
+    ///
+    /// This function does't configure the TLS handler. Callers of this function need to add a TLS
+    /// handler appropriately configured to perform protocol negotiation.
+    ///
+    /// - parameters:
+    ///     - configurator: A callback that will be invoked after a protocol has been negotiated.
+    ///         The callback only needs to add application-specific handlers and must return a future
+    ///         that completes when the channel has been fully mutated.
+    /// - returns: `EventLoopFuture<Void>` that completes when the channel is ready.
+    public func configureCommonHTTPServerPipeline(_ configurator: @escaping (Channel) -> EventLoopFuture<Void>) -> EventLoopFuture<Void> {
```
mariosangiorgio: This is the new code I wrote. Just a wrapper around `configureHTTP2SecureUpgrade` with adapters to make the provided callback work with the two protocols

Tests/NIOHTTP2Tests/ConfiguringPipelineTests.swift
```diff
@@ -181,4 +183,114 @@ class ConfiguringPipelineTests: XCTestCase {
...
+        self.interactInMemory(self.clientChannel, self.serverChannel)
+        (self.clientChannel.eventLoop as! EmbeddedEventLoop).run()
+        do {
+            try requestPromise.futureResult.wait()
+            XCTFail("Did not throw")
+        } catch is NIOHTTP2Errors.StreamClosed {
+            // ok
+        } catch {
+            XCTFail("Unexpected error: \(error)")
+        }
```
Lukasa: Let's replace this with:

```swift
XCTAssertThrowsError(try requestPromise.futureResult.wait()) { error in
    XCTAssertTrue(error is NIOHTTP2Errors.StreamClosed)
}
```

mariosangiorgio: I'm going to replace it also in `testBasicPipelineCommunicates` since I copy and pasted the code for this test from there and applied some small modifications.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -104,4 +113,84 @@ extension Channel {
...
+        let h2ChannelConfigurator = { (channel: Channel) -> EventLoopFuture<Void> in
+            channel.configureHTTP2Pipeline(mode: .server) { (streamChannel, streamID) -> EventLoopFuture<Void> in
+                streamChannel.pipeline.addHandler(HTTP2ToHTTP1ServerCodec(streamID: streamID)).flatMap { () -> EventLoopFuture<Void> in
+                    configurator(streamChannel)
+                }.flatMap { () -> EventLoopFuture<Void> in
+                    streamChannel.pipeline.addHandler(ErrorHandler())
```
mariosangiorgio: What about this handler? Would a default behaviour (closing the channel?) be fine or should we provide two callbacks for the error cases?

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -104,4 +105,86 @@ extension Channel {
...
+    ///         The callback only needs to add application-specific handlers and must return a future
+    ///         that completes when the channel has been fully mutated.
+    /// - returns: `EventLoopFuture<Void>` that completes when the channel is ready.
+    public func configureCommonHTTPServerPipeline(
+        h2ConnectionChannelConfigurator: @escaping (Channel) -> EventLoopFuture<Void>,
+        _ configurator: @escaping (Channel) -> EventLoopFuture<Void>) -> EventLoopFuture<Void> {
+        let h2ChannelConfigurator = { (channel: Channel) -> EventLoopFuture<Void> in
+            channel.configureHTTP2Pipeline(mode: .server) { (streamChannel, streamID) -> EventLoopFuture<Void> in
+                streamChannel.pipeline.addHandler(HTTP2ToHTTP1ServerCodec(streamID: streamID)).flatMap { () -> EventLoopFuture<Void> in
+                    configurator(streamChannel)
```
mariosangiorgio: @Lukasa is it fine to expect that `configurator` do the channel close I previously added separately with `ErrorHandler`?

The reason I added was a cargo-culting of what I've seen in https://github.com/apple/swift-nio-examples/blob/master/http2-server/Sources/http2-server/main.swift#L137:L139 but after having another look I think we might just want to configure both at the same time in the callback.

Lukasa: Yes, that's totally fine.

Lukasa:

Lukasa:
Looks like there may be an optional promotion issue in 5.0? Try using `case .some(.whatever))` instead.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -104,4 +105,86 @@ extension Channel {
...
+        return self.pipeline.addHandler(alpnHandler)
+    }
+
+    /// Configures a `ChannelPipeline` to speak either HTTP or HTTP/2 accordingly to what can be negotiated with the client.
+    ///
+    /// This helper takes care of configuring the server pipeline such that it negotiates whether to
+    /// use HTTP/1.1 or HTTP/2. Once the protocol to use for the channel has been negotitated, the
+    /// provided callback will configure the application-specific handlers in a protocol-agnostic way.
+    ///
+    /// This function does't configure the TLS handler. Callers of this function need to add a TLS
```
Lukasa:
```suggestion
    /// This function doesn't configure the TLS handler. Callers of this function need to add a TLS
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -104,4 +105,86 @@ extension Channel {
...
+            }
+        }
+
+        return self.pipeline.addHandler(alpnHandler)
+    }
+
+    /// Configures a `ChannelPipeline` to speak either HTTP or HTTP/2 accordingly to what can be negotiated with the client.
+    ///
+    /// This helper takes care of configuring the server pipeline such that it negotiates whether to
+    /// use HTTP/1.1 or HTTP/2. Once the protocol to use for the channel has been negotitated, the
```
Lukasa:
```suggestion
    /// use HTTP/1.1 or HTTP/2. Once the protocol to use for the channel has been negotiated, the
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -104,4 +105,86 @@ extension Channel {
...
+                // We negotiated something that isn't HTTP/1.1. This is a bad scene, and is a good indication
+                // of a user configuration error. We're going to close the connection directly.
+                return self.close().flatMap { self.eventLoop.makeFailedFuture(NIOHTTP2Errors.InvalidALPNToken()) }
+            }
+        }
+
+        return self.pipeline.addHandler(alpnHandler)
+    }
+
+    /// Configures a `ChannelPipeline` to speak either HTTP or HTTP/2 accordingly to what can be negotiated with the client.
```
Lukasa:
```suggestion
    /// Configures a `ChannelPipeline` to speak either HTTP or HTTP/2 according to what can be negotiated with the client.
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -104,4 +105,86 @@ extension Channel {
...
+    ///
+    /// - parameters:
+    ///     - h2ConnectionChannelConfigurator: A callback that will be invoked only
+    ///         when the negotiated protocol is H2 to configure the connection channel.
+    ///     - configurator: A callback that will be invoked after a protocol has been negotiated.
+    ///         The callback only needs to add application-specific handlers and must return a future
+    ///         that completes when the channel has been fully mutated.
+    /// - returns: `EventLoopFuture<Void>` that completes when the channel is ready.
+    public func configureCommonHTTPServerPipeline(
+        h2ConnectionChannelConfigurator: @escaping (Channel) -> EventLoopFuture<Void>,
```
Lukasa: This closure should be allowed to be optional.

mariosangiorgio: Are you thinking of something like:
```swift
            }.flatMap { (_: HTTP2StreamMultiplexer) in
                if let h2ConnectionChannelConfigurator = h2ConnectionChannelConfigurator {
                    return h2ConnectionChannelConfigurator(channel)
                } else {
                    return channel.eventLoop.makeSucceededFuture(())
                }
            }
```
I'm not sure if there is any better default behaviour than doing nothing by using `return channel.eventLoop.makeSucceededFuture(())`

Lukasa: Yup, that's the right default behaviour.

PR #195, opened 2020-04-17
link swift-nio-extras in docs

weissi:

PR #196, opened 2020-04-20
Fixup the docker setup.

Lukasa:
Unblocks #195 and future work.

### DIFF
--- a/docker/docker-compose.1804.53.yaml
+++ b/docker/docker-compose.1804.53.yaml
@@ -1,0 +1,36 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: swift-nio-http2:18.04-5.3
+    build:
+      args:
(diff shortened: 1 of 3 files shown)


docker/docker-compose.1804.53.yaml
```diff
@@ -0,0 +1,37 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: swift-nio-http2:18.04-5.3
+    build:
+      args:
+        ubuntu_version: "bionic"
+        swift_version: "5.3"
```
weissi, line 10: this one can't be right. Did you try the `docker-compose` command? Here I think you need to use `base_image=swiftlang/swift:...` to get the nightlies

Lukasa:
Good catch, fixed.

weissi:
@tomerd could you create us a 5.3 CI job for this?

PR #198, opened 2020-04-28
Add debugging info to HTTP2StreamChannel too.

Lukasa:
Motivation:

Sometimes folks will want to print things about the HTTP2StreamChannel,
and we should give them the things we can safely get. We also may want a
shorter log for NIOHTTP2Handler as the standard description, with the
wall of text available in the debug description.

Modifications:

- Conform HTTP2StreamChannel to CustomStringConvertible
- Conform NIOHTTP2Handler to CustomDebugStringConvertible.

Result:

Even better debug info.


Lukasa:
5.3 failure is compiler crash, merging over it.

PR #202, opened 2020-05-05
Make multiplexer's targetWindowSize configurable

johnkassebaum:
Make multiplexer's `targetWindowSize` configurable and use configurability when creating stream channels.
    
Motivation:
`targetWindowSize` constrains throughput. The default hardcoded `65535` is an unecessarily low limit that should be otherwise configrable. A `TODO` in the code stated as such.
    
Modifications:    
- In `HTTP2StreamMultiplexer`, responding to a `TODO` where stream multiplexer creates a stream channel, created a multiplexr stored property to be used which is set in the default initilaizer.
- In `HTTP2PipelineHelpers`, added versions of `configureHTTP2Pipeline(..)` and `configureCommonHTTP2ServerPipeline(..)` that require an `activeTargetWindow` arg.
    
Result:    
Clients, such as `GRPCClient`, can configure pipelines with targetWindowSizes to their liking.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
+++ b/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
@@ -97,10 +97,33 @@ extension Channel {
                                        initialLocalSettings: [HTTP2Setting] = nioDefaultSettings,
                                        position: ChannelPipeline.Position = .last,
                                        inboundStreamStateInitializer: ((Channel, HTTP2StreamID) -> EventLoopFuture<Void>)? = nil) -> EventLoopFuture<HTTP2StreamMultiplexer> {
+        return self.configureHTTP2Pipeline(mode: mode, initialLocalSettings: initialLocalSettings, position: position, targetWindowSize: 65535, inboundStreamStateInitializer: inboundStreamStateInitializer)
+    }
+    
+    /// Configures a `ChannelPipeline` to speak HTTP/2.
+    ///
+    /// In general this is not entirely useful by itself, as HTTP/2 is a negotiated protocol. This helper does not handle negotiation.
+    /// Instead, this simply adds the handlers required to speak HTTP/2 after negotiation has completed, or when agreed by prior knowledge.
+    /// Whenever possible use this function to setup a HTTP/2 server pipeline, as it allows that pipeline to evolve without breaking your code.
+    ///
+    /// - parameters:
+    ///     - mode: The mode this pipeline will operate in, server or client.
+    ///     - initialLocalSettings: The settings that will be used when establishing the connection. These will be sent to the peer as part of the
+    ///         handshake.
+    ///     - position: The position in the pipeline into which to insert these handlers.
+    ///     - targetWindowSize: The target size of the HTTP/2 flow control window.
+    ///     - inboundStreamStateInitializer: A closure that will be called whenever the remote peer initiates a new stream. This should almost always
+    ///         be provided, especially on servers.
+    /// - returns: An `EventLoopFuture` containing the `HTTP2StreamMultiplexer` inserted into this pipeline, which can be used to initiate new streams.
+    public func configureHTTP2Pipeline(mode: NIOHTTP2Handler.ParserMode,
+                                       initialLocalSettings: [HTTP2Setting] = nioDefaultSettings,
+                                       position: ChannelPipeline.Position = .last,
+                                       targetWindowSize: Int,
+                                       inboundStreamStateInitializer: ((Channel, HTTP2StreamID) -> EventLoopFuture<Void>)? = nil) -> EventLoopFuture<HTTP2StreamMultiplexer> {
         var handlers = [ChannelHandler]()
         handlers.reserveCapacity(2)  // Update this if we need to add more handlers, to avoid unnecessary reallocation.
         handlers.append(NIOHTTP2Handler(mode: mode, initialSettings: initialLocalSettings))
-        let multiplexer = HTTP2StreamMultiplexer(mode: mode, channel: self, inboundStreamStateInitializer: inboundStreamStateInitializer)
+        let multiplexer = HTTP2StreamMultiplexer(mode: mode, channel: self, targetWindowSize: targetWindowSize, inboundStreamStateInitializer: inboundStreamStateInitializer)
         handlers.append(multiplexer)
 
         return self.pipeline.addHandlers(handlers, position: position).map { multiplexer }
@@ -170,8 +193,32 @@ extension Channel {
     public func configureCommonHTTPServerPipeline(
         h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)? = nil,
         _ configurator: @escaping (Channel) -> EventLoopFuture<Void>) -> EventLoopFuture<Void> {
+        return self.configureCommonHTTPServerPipeline(h2ConnectionChannelConfigurator: h2ConnectionChannelConfigurator, targetWindowSize: 65535, configurator)
+    }
+    
+    /// Configures a `ChannelPipeline` to speak either HTTP or HTTP/2 according to what can be negotiated with the client.
+    ///
+    /// This helper takes care of configuring the server pipeline such that it negotiates whether to
+    /// use HTTP/1.1 or HTTP/2. Once the protocol to use for the channel has been negotiated, the
+    /// provided callback will configure the application-specific handlers in a protocol-agnostic way.
+    ///
+    /// This function doesn't configure the TLS handler. Callers of this function need to add a TLS
+    /// handler appropriately configured to perform protocol negotiation.
+    ///
+    /// - parameters:
+    ///     - h2ConnectionChannelConfigurator: An optional callback that will be invoked only
+    ///         when the negotiated protocol is H2 to configure the connection channel.
+    ///     - targetWindowSize: The target size of the HTTP/2 flow control window.
+    ///     - configurator: A callback that will be invoked after a protocol has been negotiated.
+    ///         The callback only needs to add application-specific handlers and must return a future
+    ///         that completes when the channel has been fully mutated.
+    /// - returns: `EventLoopFuture<Void>` that completes when the channel is ready.
+    public func configureCommonHTTPServerPipeline(
+        h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)? = nil,
+        targetWindowSize: Int,
+        _ configurator: @escaping (Channel) -> EventLoopFuture<Void>) -> EventLoopFuture<Void> {
         let h2ChannelConfigurator = { (channel: Channel) -> EventLoopFuture<Void> in
-            channel.configureHTTP2Pipeline(mode: .server) { (streamChannel, streamID) -> EventLoopFuture<Void> in
+            channel.configureHTTP2Pipeline(mode: .server, targetWindowSize: targetWindowSize) { (streamChannel, streamID) -> EventLoopFuture<Void> in
                 streamChannel.pipeline.addHandler(HTTP2ToHTTP1ServerCodec(streamID: streamID)).flatMap { () -> EventLoopFuture<Void> in
                     configurator(streamChannel)
                 }
@@ -191,4 +238,5 @@ extension Channel {
         return self.configureHTTP2SecureUpgrade(h2ChannelConfigurator: h2ChannelConfigurator,
                                                 http1ChannelConfigurator: http1ChannelConfigurator)
     }
+
 }
--- a/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
@@ -37,6 +37,7 @@ public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboun
     private var didReadChannels: StreamChannelList = StreamChannelList()
     private let streamChannelOutboundBytesHighWatermark: Int
     private let streamChannelOutboundBytesLowWatermark: Int
+    private let targetWindowSize: Int
 
     public func handlerAdded(context: ChannelHandlerContext) {
         // We now need to check that we're on the same event loop as the one we were originally given.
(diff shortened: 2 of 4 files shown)


Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -96,11 +96,12 @@ extension Channel {
     public func configureHTTP2Pipeline(mode: NIOHTTP2Handler.ParserMode,
                                        initialLocalSettings: [HTTP2Setting] = nioDefaultSettings,
                                        position: ChannelPipeline.Position = .last,
+                                       targetWindowSize: Int32 = 65535,
```
Lukasa, line 99: Unfortunately, for `public func`s we can't add new arguments to the methods, even when they're defaulted, as it is still API breaking. Instead, we need to add a new function that has the same signature as the previous one _plus_ the extra argument, not defaulted.

Do you mind doing that?

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -169,9 +170,9 @@ extension Channel {
     /// - returns: `EventLoopFuture<Void>` that completes when the channel is ready.
     public func configureCommonHTTPServerPipeline(
         h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)? = nil,
-        _ configurator: @escaping (Channel) -> EventLoopFuture<Void>) -> EventLoopFuture<Void> {
+        _ configurator: @escaping (Channel) -> EventLoopFuture<Void>, targetWindowSize: Int32 = 65535) -> EventLoopFuture<Void> {
```
Lukasa, line 173: This change falls foul of the new argument rule as well.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
```diff
@@ -205,7 +206,7 @@ public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboun
     ///         receiving a `HEADERS` frame from a client. For clients, these are channels created by
     ///         receiving a `PUSH_PROMISE` frame from a server. To initiate a new outbound channel, use
     ///         `createStreamChannel`.
-    public convenience init(mode: NIOHTTP2Handler.ParserMode, channel: Channel, targetWindowSize: Int = 65535, inboundStreamStateInitializer: ((Channel, HTTP2StreamID) -> EventLoopFuture<Void>)? = nil) {
+    public convenience init(mode: NIOHTTP2Handler.ParserMode, channel: Channel, targetWindowSize: Int32, inboundStreamStateInitializer: ((Channel, HTTP2StreamID) -> EventLoopFuture<Void>)? = nil) {
```
Lukasa, line 209: This addition also needs a new function.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
```diff
@@ -233,15 +234,16 @@ public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboun
     ///         `createStreamChannel`.
     public init(mode: NIOHTTP2Handler.ParserMode,
                 channel: Channel,
-                targetWindowSize: Int = 65535,
+                targetWindowSize: Int32,
```
Lukasa, line 237: This addition also needs a new function.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
glbrntt: This will definitely need documenting too. `targetWindowSize` should be the first argument as well so we can use trailing closure syntax.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
glbrntt: Or just drop this change altogether: we need the default to not break API. We can just override it from the helpers.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
glbrntt: Same here re: dropping this change.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
johnkassebaum: Will drop.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
johnkassebaum: Will drop.

johnkassebaum: I removed the default value to guard against coder error in the convenience, forcing use of the convenience init's arg, but doing so required fixing all the calls in tests.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -105,6 +105,35 @@ extension Channel {
...
+    /// - parameters:
+    ///     - mode: The mode this pipeline will operate in, server or client.
+    ///     - initialLocalSettings: The settings that will be used when establishing the connection. These will be sent to the peer as part of the
+    ///         handshake.
+    ///     - position: The position in the pipeline into which to insert these handlers.
+    ///     - targetWindowSize: The target size of the inbound flow control window.
+    ///     - inboundStreamStateInitializer: A closure that will be called whenever the remote peer initiates a new stream. This should almost always
+    ///         be provided, especially on servers.
+    /// - returns: An `EventLoopFuture` containing the `HTTP2StreamMultiplexer` inserted into this pipeline, which can be used to initiate new streams.
+    public func configureHTTP2Pipeline(mode: NIOHTTP2Handler.ParserMode,
```
glbrntt, line 124: Could you have the other implementation call this implementation with the default value of `targetWindowSize`? I'd rather duplicate the default value than the function body.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -191,4 +220,48 @@ extension Channel {
...
+    ///
+    /// - parameters:
+    ///     - h2ConnectionChannelConfigurator: An optional callback that will be invoked only
+    ///         when the negotiated protocol is H2 to configure the connection channel.
+    ///     - targetWindowSize: The target size of the inbound flow control window.
+    ///     - configurator: A callback that will be invoked after a protocol has been negotiated.
+    ///         The callback only needs to add application-specific handlers and must return a future
+    ///         that completes when the channel has been fully mutated.
+    /// - returns: `EventLoopFuture<Void>` that completes when the channel is ready.
+    public func configureCommonHTTPServerPipeline(
```
glbrntt, line 241: Same here: I'd like to avoid the duplication of the function body.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
johnkassebaum: Sure, thought about that, but didn't know who determined the default value -- is the presumption that default value is `SETTINGS_INITIAL_WINDOW_SIZE`'s default of 65535?

I.e., if calling the new from the old with 65535, the default values in `HTTP2StreamMultiplexer` are essentially meaningless since they are never invoked without a `targetWindowSize` arg value.

glbrntt: > Sure, thought about that, but didn't know who determined the default value -- is the presumption that default value is `SETTINGS_INITIAL_WINDOW_SIZE`'s default of 65535?

The default should be whatever the multiplexer defaults to.

> I.e., if calling the new from the old with 65535, the default values in `HTTP2StreamMultiplexer` are essentially meaningless since they are never invoked without a `targetWindowSize` arg value.

That's only true if you're using these functions to construct an HTTP2 pipeline: it's still valid to add the HTTP2 handler and multiplexer manually.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -170,8 +193,32 @@ extension Channel {
...
+    /// use HTTP/1.1 or HTTP/2. Once the protocol to use for the channel has been negotiated, the
+    /// provided callback will configure the application-specific handlers in a protocol-agnostic way.
+    ///
+    /// This function doesn't configure the TLS handler. Callers of this function need to add a TLS
+    /// handler appropriately configured to perform protocol negotiation.
+    ///
+    /// - parameters:
+    ///     - h2ConnectionChannelConfigurator: An optional callback that will be invoked only
+    ///         when the negotiated protocol is H2 to configure the connection channel.
+    ///     - targetWindowSize: The target size of the inbound flow control window.
```
Lukasa, line 211: Can we call out that this is the HTTP/2 flow control window specifically?

johnkassebaum:
Duplicated all tests in the `NIOHTTP2Tests` that create a pipeline.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -97,10 +97,33 @@ extension Channel {
                                        initialLocalSettings: [HTTP2Setting] = nioDefaultSettings,
                                        position: ChannelPipeline.Position = .last,
                                        inboundStreamStateInitializer: ((Channel, HTTP2StreamID) -> EventLoopFuture<Void>)? = nil) -> EventLoopFuture<HTTP2StreamMultiplexer> {
+        return configureHTTP2Pipeline(mode: mode, initialLocalSettings: initialLocalSettings, position: position, targetWindowSize: 65535, inboundStreamStateInitializer: inboundStreamStateInitializer)
```
Lukasa, line 100: Nit: NIO prefers to use the explicit `self` keyword.

```suggestion
        return self.configureHTTP2Pipeline(mode: mode, initialLocalSettings: initialLocalSettings, position: position, targetWindowSize: 65535, inboundStreamStateInitializer: inboundStreamStateInitializer)
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -170,8 +193,32 @@ extension Channel {
     public func configureCommonHTTPServerPipeline(
         h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)? = nil,
         _ configurator: @escaping (Channel) -> EventLoopFuture<Void>) -> EventLoopFuture<Void> {
+        return configureCommonHTTPServerPipeline(h2ConnectionChannelConfigurator: h2ConnectionChannelConfigurator, targetWindowSize: 65535, configurator)
```
Lukasa, line 196: Same nit here:

```suggestion
        return self.configureCommonHTTPServerPipeline(h2ConnectionChannelConfigurator: h2ConnectionChannelConfigurator, targetWindowSize: 65535, configurator)
```

glbrntt:

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
johnkassebaum: No problem

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
johnkassebaum: ditto

glbrntt:

Lukasa:

Lukasa:
Can you run the `generate_linux_tests.rb` script in the `script` directory and check in the resulting changes?

johnkassebaum:
@Lukasa If the request to run the `generate_linux_tests.rb` script was directed to me, then I'm missing something because it generated no output or changes

Lukasa:
@johnkassebaum CI claims it will generate this diff:

```
16:20:27 --- a/Tests/NIOHTTP2Tests/ConfiguringPipelineTests+XCTest.swift
16:20:27 +++ b/Tests/NIOHTTP2Tests/ConfiguringPipelineTests+XCTest.swift
16:20:27 @@ -27,10 +27,15 @@ extension ConfiguringPipelineTests {
16:20:27     static var allTests : [(String, (ConfiguringPipelineTests) -> () throws -> Void)] {
16:20:27        return [
16:20:27                  ("testBasicPipelineCommunicates", testBasicPipelineCommunicates),
16:20:27 +                ("testBasicPipelineCommunicatesWithTargetWindowSize", testBasicPipelineCommunicatesWithTargetWindowSize),
16:20:27                  ("testPipelineRespectsPositionRequest", testPipelineRespectsPositionRequest),
16:20:27 +                ("testPipelineRespectsPositionRequestWithTargetWindowSize", testPipelineRespectsPositionRequestWithTargetWindowSize),
16:20:27                  ("testPreambleGetsWrittenOnce", testPreambleGetsWrittenOnce),
16:20:27 +                ("testPreambleGetsWrittenOnceWithTargetWindowSize", testPreambleGetsWrittenOnceWithTargetWindowSize),
16:20:27                  ("testClosingParentChannelClosesStreamChannel", testClosingParentChannelClosesStreamChannel),
16:20:27 +                ("testClosingParentChannelClosesStreamChannelWithTargetWindowSize", testClosingParentChannelClosesStreamChannelWithTargetWindowSize),
16:20:27                  ("testNegotiatedHTTP2BasicPipelineCommunicates", testNegotiatedHTTP2BasicPipelineCommunicates),
16:20:27 +                ("testNegotiatedHTTP2BasicPipelineCommunicatesWithTargetWindowSize", testNegotiatedHTTP2BasicPipelineCommunicatesWithTargetWindowSize),
16:20:27                  ("testNegotiatedHTTP1BasicPipelineCommunicates", testNegotiatedHTTP1BasicPipelineCommunicates),
16:20:27             ]
16:20:27     }
```

docker/docker-compose.1804.53.yaml
```diff
@@ -6,7 +6,7 @@ services:
     image: swift-nio-http2:18.04-5.3
     build:
       args:
-        base_image: "swiftlang/swift:nightly-master-bionic"
```
Lukasa, line 9: I presume this change was not supposed to come along?

johnkassebaum: Heh. Some confusion on my part it seems around being behind master. I will remove it

Lukasa:

johnkassebaum:
Sorry if this appears as a noob question: When is the appropriate time to hit 'Update branch' button?

Lukasa:
Whenever suits you: but bear in mind that if you force-push to that branch again you can undo the update.

johnkassebaum:
That's what I observe, and which I just did.

Lukasa:

Lukasa:

Lukasa:

Lukasa:
5.3 compiler crash is expected.

PR #204, opened 2020-05-18
Explicitly test we tolerate GREASE.

Lukasa:
Motivation:

As Chrome has started GREASEing HTTP/2 settings and frames, we should
ensure that we have tests that explicitly validate that we tolerate
GREASEing these parameters.

A future version of swift-nio-http2 should start emitting GREASE as
well, but for now we'll satisfy ourselves with this.

Two relevant citations:

https://tools.ietf.org/html/draft-bishop-httpbis-grease-00

https://bugs.chromium.org/p/chromium/issues/detail?id=1083706

Modifications:

- Added a frame parser test
- Added a end-to-end settings test.

Result:

Validated that GREASE is tolerated.


Lukasa:
Merging over the 5.3 compiler crash.

PR #205, opened 2020-06-01
Respect target window size on inbound channels.

Lukasa:
Motivation:

As part of #202 we added support for configuring the target window size
of stream channels. That was plumbed through into outbound channels, but
we missed the inbound ones, meaning that this only really worked for
clients, not servers. That hardly seems fair!

Modifications:

Pass target window size to inbound channels too.

Result:

We can control target window size of inbound channels.


Lukasa:
5.3 compiler failure is well-known.

PR #209, opened 2020-06-15
fix warnings with newer Swift compilers

weissi:
Motiviation:

A few `#file` parameters have been forwarded to (now) `#filePath` ones.

Modification:

- silence the `#file`/`#filePath` warning by adding `(#file)` because we
  can't forward them properly right now
  (https://bugs.swift.org/browse/SR-12934). That's okay because even on
  the Swift master branch, `#file` == `#filePath`.

Result:

No warnings in newer Swift compilers.

### DIFF
--- a/Tests/NIOHPACKTests/HeaderTableTests.swift
+++ b/Tests/NIOHPACKTests/HeaderTableTests.swift
@@ -19,7 +19,7 @@ import NIO
 func XCTAssertEqualTuple<T1: Equatable, T2: Equatable>(_ expression1: @autoclosure () throws -> (T1, T2)?,
                                                        _ expression2: @autoclosure () throws -> (T1, T2)?,
                                                        _ message: @autoclosure () -> String = "",
-                                                       file: StaticString = #file, line: UInt = #line) {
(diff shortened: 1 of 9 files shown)


lukeredpath:
@Lukasa came across this PR when trying to fix a similar problem in some of my own custom assertions but it doesn't seem to work in the first Xcode 12 beta, was wondering if you'd noticed the same - changing `#file` to `(#file)` to silence warnings results in test failures not being highlighted on the correct line. Neither does just using `#file`, which both warns and fails to highlight. If you're using Xcode 12/Swift 5.3 you need to use `#filePath`.

Lukasa:
Yeah, we're aware. However, `#filePath` is unusable for us because of the fact that we support all the way back to Swift 5.0, which does not contain `#filePath`. There's some [discussion on the Swift forums](https://forums.swift.org/t/revisiting-the-source-compatibility-impact-of-se-0274-concise-magic-file-names/37720).

lukeredpath:
Thanks for that link, I was searching on their and couldn't seem to find something recent - it's a real PITA.

Issue #214, opened 2020-07-28
Lift requirement to create streams in order


Lukasa:
In swift-nio-http2 today we have an unspoken hard requirement, which is that if you call `createStreamChannel` multiple times you _must_ send the first write on each of those channels in the order you created them. This is because `createStreamChannel` allocates a stream ID _immediately_ for each of those channels, but that allocation isn't meaningful until we try to write to the network. If you write out of order, the later streams will use the higher stream ID allocated to them, and will implicitly retire the lower IDs used by the earlier channels. When those earlier channels try to send, the core state machine will reject them for violating stream ID ordering.

After discussing with @glbrntt we think this is made up of several changes:

- [x] Define a new internal protocol: `HTTP2FrameConvertible`. This is a protocol that encapsulates creating `HTTP2Frame`s from a given object and vice-versa. We will have two conforming types: `HTTP2Frame` (trivial) and `HTTP2Frame.Payload` (straightforward, but not trivial). (#216)
- [x] Define a new protocol: `HTTP2FramePayloadConvertible`. This is the inverse of the previous protocol: creates a payload from the given object, or vice versa. (#216)
- [x] Make `HTTP2StreamChannel` generic over an outbound payload type conforming to `HTTP2FrameConvertible`. This will become the inbound and outbound message type. This requires updates to the various inbound and outbound reading and writing functions. This should not require any interface changes on the interface between the multiplexer and the stream. The multiplexer will need to be updated to hold the new type (`HTTP2StreamChannel<HTTP2Frame>`). (#218)
- [x] Define a `struct` that provides the interface `HTTP2StreamMultiplexer` expects from `HTTP2StreamChannel`, that holds the channel privately. This should be backed by an `enum`, which we will use later to add a new case for the new channel type. Refactor `HTTP2StreamMultiplexer` to use this type. (#215)
- [x] Update `HTTP2StreamChannel` to tolerate its streamID being optional, and nil on initialisation. This requires a new function on `HTTP2StreamMultiplexer` to get the next stream ID, which is currently done in `createStreamChannel`. (#217)
- [x] Add a new `createStreamChannel` function with a new initializer that no longer gets a stream ID. Plumb through this initializer to create new stream channels that use `HTTP2Frame.Payload` instead of `HTTP2Frame`. Update the `enum` created above for the new case. (#221)
[...]

PR #215, opened 2020-07-28
Define abstraction for multiplexer holding channel.

Lukasa:
Motivation:

As part of the work in #214 we need a way for the HTTP2StreamMultiplexer
to hold multiple kinds of channels. It would be best if doing so didn't
require substantial changes to HTTP2StreamMultiplexer, which is
difficult because right now it is tightly coupled to HTTP2StreamChannel.

There are a few ways we could do this. The most obvious option to most
Swift programmers would be to define a protocol that covers the surface
area we want to use, and then hold that. This works fairly well, but the
downside of it is that it will force all calls from
HTTP2StreamMultiplexer to become dynamically dispatched. These indirect
calls are optimisation boundaries and so provide some performance costs.

Another option would be to create a subclassing relationship. This is
ok, but it also forces us into the realm of dynamic dispatch on some of
these calls, and also provides general difficulty if we ever want to
provide a backing channel that isn't easily made into a subclass.

The remaining option is to provide an abstract object that uses an
`enum` to store the kinds of channel we know about, and jump through it.
This works really well for small numbers of channel types (1 and 2
channels being the best outcomes), as it introduces very few extra
branches which are highly predictable, as well as allowing direct
function calls. This is great: one predictable branch that cannot
possibly miss cache, followed by a direct function call. Joy of joys!

However, it would be best if we didn't tie ourselves into this
implementation strategy. After all, down the road we could well end up
with more kinds of channels, and at a certain point the cost of the
indirect function call becomes _lower_ than the cost of jumping through
the enumeration. For that reason, we should define an abstraction.

Enter: `MultiplexerAbstractChannel`. This concrete type defines the
interface that `HTTP2StreamMultiplexer` will use to interact with stream
channels. As it's internal, the abstraction is totally porous: Swift is
capable of optimising it away to nothing. This type is explicitly a
reference type, maintaining the semantic expected by
`HTTP2StreamMultiplexer`.

This provides a nice encapsulation over our abstraction.
`HTTP2StreamMultiplexer` never needs to know about how we abstract over
multiple channel types: the entire abstraction lives in a single, simple
file, and can easily be rewritten without difficulty.

Modifications:

- Define `MultiplexerAbstractChannel` and the interface expected by
  `HTTP2StreamMultiplexer`.
- Define a `BaseChannel` `enum` that will be used to dispatch to
  multiple `Channel` types.
- Rewrite `HTTP2StreamMultiplexer` to use this new abstraction.

Result:

A customisation point for new channels will exist.


Sources/NIOHTTP2/MultiplexerAbstractChannel.swift
```diff
@@ -0,0 +1,156 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
glbrntt, line 5:
```suggestion
// Copyright (c) 2020 Apple Inc. and the SwiftNIO project authors
```

PR #218, opened 2020-07-28
Make HTTP2StreamChannel generic over a message type

glbrntt:
Motivation:

As part of #214 we need to make `HTTP2StreamChannel` generic over the type
of message it expects to read and write. This allows us to define an
`HTTP2StreamChannel` which uses `HTTP2Frame.FramePayload` as its
currency type, rather than `HTTP2Frame`.

Modifications:

- Make `HTTP2StreamChannel` generic over `Message` which is constrained
  by conformance to `HTTP2FrameConvertible` and
  `HTTP2FramePayloadConvertible`
- `HTTP2StreamChannel` now expects to read in `Message`s (as opposed to
  `HTTP2Frame`s) and have `Message`s written in to it.
- Added typealiases for frame and payload based stream channels.
- Added support for the payload based channel in
  `MultiplexerAbstractChannel` (although one can't be created yet).

Result:

We can create payload based stream channels.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2FrameConvertible.swift
+++ b/Sources/NIOHTTP2/HTTP2FrameConvertible.swift
@@ -13,6 +13,9 @@
 //===----------------------------------------------------------------------===//
 
 protocol HTTP2FrameConvertible {
+    /// Initialize `Self` from an `HTTP2Frame`.
+    init(http2Frame: HTTP2Frame)
+
     /// Makes an `HTTPFrame` with the given `streamID`.
     ///
     /// - Parameter streamID: The `streamID` to use when constructing the frame.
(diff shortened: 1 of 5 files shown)


Sources/NIOHTTP2/HTTP2FrameConvertible.swift
```diff
@@ -23,3 +23,39 @@ protocol HTTP2FramePayloadConvertible {
     /// Makes a `HTTP2Frame.FramePayload`.
     var payload: HTTP2Frame.FramePayload { get }
 }
+
+extension HTTP2FrameConvertible where Self: HTTP2FramePayloadConvertible {
```
Lukasa, line 27: This extension seems to properly simply belong on `HTTP2FramePayloadConvertible`, no?

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -615,7 +621,7 @@ internal extension HTTP2StreamChannel {
     ///
     /// - parameters:
     ///     - frame: The `HTTP2Frame` received from the network.
-    func receiveInboundFrame(_ frame: HTTP2Frame) {
+    func receiveInbound(_ message: Message) {
```
Lukasa, line 624: I think I'd rather that this didn't change, and instead that we added a line to convert the frame to `Message` in this method.

Sources/NIOHTTP2/MultiplexerAbstractChannel.swift
```diff
@@ -77,62 +84,80 @@ extension MultiplexerAbstractChannel {
         switch self.baseChannel {
         case .frameBased(let base):
             base.configure(initializer: initializer, userPromise: promise)
+        case .payloadBased(let base):
+            base.configure(initializer: initializer, userPromise: promise)
```
Lukasa, line 88: We should `fatalError` here: we don't ever want this branch to be hit.

Sources/NIOHTTP2/HTTP2FrameConvertible.swift
glbrntt: From a purely implementation perspective yes. But we use only this property in the context of writing (i.e. we require `HTTP2FrameConvertible` rather than `HTTP2FramePayloadConvertible`). WDYT?

Lukasa: Hmmm. I guess, yeah. 👍

PR #221, opened 2020-07-29
Allow payload-based stream channels to be created.

glbrntt:
Motivation:

To lift the requirement that stream creation order must match the order
of the first write on those channels we added a payload-based stream
channel. We need a way for users to create these channels and initialze
these channels when they are inbound.

Modifications:

- Adds a `createStreamChannel` using a stream initialzer without a
  stream ID
- Adds a new multiplexer `init` where the inbound stream initialzer does
  not use stream IDs
- Adds an additional 'configure' to MultiplexerAbstractChannel
- Replaces the `init` in `MultiplexerAbstractChannel` with two `static func`s:
  we can't switch on an optional stream ID since inbound streams always
  have a stream ID; we need to be be able to explicitly support choosing
  the type of channel we want.
- Minor refactoring in `HTTP2StreamChannel` to avoid duplication between the
  `configure` functions

Result:

- The multiplexer can create payload based streams manually and when a
  new stream is inbound.


glbrntt:
Hmm allocations have increased for `create_client_stream_channel` on 5.0... 👀

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -163,34 +163,66 @@ final class HTTP2StreamChannel<Message: HTTP2FramePayloadConvertible & HTTP2Fram
...
+    private func getAutoReadFromParent() -> EventLoopFuture<Bool> {
+        // This force unwrap is safe as parent is assigned in the initializer, and never unassigned.
+        // Note we also don't set the value here: the additional `map` causes an extra allocation
+        // when using a Swift 5.0 compiler.
+        return self.parent!.getOption(ChannelOptions.autoRead)
+    }
 
-            promise?.fail(error)
+    /// Activates the channel if the parent channel is active and succeeds the given `promise`.
+    private func activate(promise: EventLoopPromise<Channel>?) {
```
Lukasa, line 203: Can we give this a more evocative name? There's already a `performActivation` function here. I'd like this to indicate that it's called after the initializer fires, as part of initialisation.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
```diff
@@ -75,17 +89,39 @@ public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboun
...
-                parent: self.channel,
-                multiplexer: self,
-                streamID: streamID,
-                targetWindowSize: Int32(self.targetWindowSize),
-                outboundBytesHighWatermark: self.streamChannelOutboundBytesHighWatermark,
-                outboundBytesLowWatermark: self.streamChannelOutboundBytesLowWatermark
-            )
-            self.streams[streamID] = channel
-            channel.configure(initializer: self.inboundStreamStateInitializer, userPromise: nil)
+            let channel: MultiplexerAbstractChannel
```
Lukasa, line 92: Can we arrange for `MultiplexerAbstractChannel` to paper over this distinction? I think we can do it if we make the `enum` for `inboundStreamStateInitializer` something we pass to the `MultiplexerAbstractChannel` and have it end up choosing which base channel to construct.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
glbrntt: Not overly pleased with the new name. Suggestions gladly welcomed.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
glbrntt: Yeah, that's sensible. Done.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
Lukasa: Honestly I think the new name is fine.

Sources/NIOHTTP2/MultiplexerAbstractChannel.swift
```diff
@@ -25,21 +25,42 @@ import NIO
 /// The implementation of `Equatable` & `Hashable` on this type reinforces that requirement.
 struct MultiplexerAbstractChannel {
     private var baseChannel: BaseChannel
+    private var inboundStreamStateInitializer: InboundStreamStateInitializer
+
+    private init(baseChannel: BaseChannel, inboundStreamStateInitializer: InboundStreamStateInitializer) {
```
Lukasa, line 30: Hmm, this isn't quite what I wanted either. I wanted to pass the initializer to the relevant functions. This forces us to store the inbound stream state initializer on all streams, which is mostly extra work for us. Any reason not to just pass it to `configureInboundStream`?

We don't need to _store_ this function: we just need to _have_ it in the `init` to know what channel to create. That's totally sufficient.

glbrntt: Gotcha.

PR #222, opened 2020-07-30
Add HTTP2Frame.FramePayload codecs

glbrntt:
Motivation:

As part of the work for #214 we need new codecs which transform
`HTTP2Frame.FramePayload` to and from the appropriate request and
response parts.

Modifications:

- Add `HTTP2FramePayloadToHTTP1ClientCodec`
- Add `HTTP2FramePayloadToHTTP1ServerCodec`
- Duplicate the HTTP2ToHTTP1CodecTests and update the relevant parts to
  use payloads instead of frames.
- Add relevant test helpers.
- Note: the HTTP2 to HTTP1 (frame based) codecs haven't been deprecated:
  doing so without warnings depends on #221.

Result:

- We can transform `HTTP2Frame.FramePayload` types to the relevant HTTP
  client and server request and response types.


Lukasa:
@glbrntt has some conflicts in `LinuxMain.swift`.

PR #226, opened 2020-08-04
Add new pipeline helpers and deprecate old paths

glbrntt:
Motivation:

As part of #214 we need to introduce new pipeline helpers to configure
the channel pipeline with the new handlers. This PR adds those helpers
and deprecates the old paths.

Modifications:

- Add pipeline helpers with the new streamID-less initializer
- Switch old pipeline helpers which provided codecs to use the new path
  and codecs
- Duplicate SimpleClientServerTests and ConfiguringPipelineTests to add
  frame payload based counterparts
- Deprecate older helpers and multiplexer init, delete tests which don't
  hit deprecated paths (note: these still exist in the payload based
  counterparts)
- Update Bench1Conn10kRequests to be payload based

Result:

- More 'FramePayload' based Tests
- Paths for creating an 'HTTP2Frame' based stream via
  'HTTP2StreamMultiplexer' are now deprecated


glbrntt:
Oops. Looks like I missed some tests.

glbrntt:
The API breakage is expected: we changed the argument name and removed the default value of an `init` which is yet to be released.

```
11:35:34 Constructor HTTP2StreamMultiplexer.init(mode:channel:targetWindowSize:outboundBufferSizeHighWatermark:outboundBufferSizeLowWatermark:inboundStreamStateInitializer:) has parameter 5 type change from ((NIO.Channel) -> NIO.EventLoopFuture<Swift.Void>)? to ((NIO.Channel, NIOHTTP2.HTTP2StreamID) -> NIO.EventLoopFuture<Swift.Void>)?
11:35:34 Constructor HTTP2StreamMultiplexer.init(mode:channel:targetWindowSize:outboundBufferSizeHighWatermark:outboundBufferSizeLowWatermark:inboundStreamStateInitializer:) has removed default argument from parameter 3
11:35:34 Constructor HTTP2StreamMultiplexer.init(mode:channel:targetWindowSize:outboundBufferSizeHighWatermark:outboundBufferSizeLowWatermark:inboundStreamStateInitializer:) has removed default argument from parameter 4
```

PR #228, opened 2020-08-04
Verify payload-based stream channel fixes #214

### LINKED ISSUE #214: Lift requirement to create streams in order

glbrntt:
Motivation:

We made a bunch of changes to resolve #214. We should add a test to
ensure we actually fixed the issue.

Modifications:

- Add a test verifying that we can write on streams in a different order
  to the order in which we create them
- Add a corresponding test for frame-based streams verifying that the
  first write on each stream must match the order in which the streams
  were created
- Remove an unnecessary `throws` in the base HTTP2 to HTTP1 server codec

Result:

- No functionality change, just more tests.

### DIFF
--- a/Tests/NIOHTTP2Tests/HTTP2FramePayloadStreamMultiplexerTests.swift
+++ b/Tests/NIOHTTP2Tests/HTTP2FramePayloadStreamMultiplexerTests.swift
@@ -1716,4 +1716,50 @@ final class HTTP2FramePayloadStreamMultiplexerTests: XCTestCase {
         XCTAssertNoThrow(try channel.finish(acceptAlreadyClosed: false))
     }
 
+    @available(*, deprecated, message: "Deprecated so deprecated functionality can be tested without warnings")
+    func testWeCanCreateFrameAndPayloadBasedStreamsOnAMultiplexer() throws {
+        let frameRecorder = FrameWriteRecorder()
+        XCTAssertNoThrow(try self.channel.pipeline.addHandler(frameRecorder).wait())
+
+        let multiplexer = HTTP2StreamMultiplexer(mode: .client, channel: self.channel, inboundStreamInitializer: nil)
+        XCTAssertNoThrow(try self.channel.pipeline.addHandler(multiplexer).wait())
+        XCTAssertNoThrow(try self.channel.connect(to: SocketAddress(unixDomainSocketPath: "/whatever"), promise: nil))
+
+        // Create a payload based stream.
+        let streamAPromise = self.channel.eventLoop.makePromise(of: Channel.self)
+        multiplexer.createStreamChannel(promise: streamAPromise) { channel in
+            return channel.eventLoop.makeSucceededFuture(())
+        }
+        self.channel.embeddedEventLoop.run()
+        let streamA = try assertNoThrowWithValue(try streamAPromise.futureResult.wait())
+        // We haven't written on the stream yet: it shouldn't have a stream ID.
+        XCTAssertThrowsError(try streamA.getOption(HTTP2StreamChannelOptions.streamID).wait()) { error in
+            XCTAssert(error is NIOHTTP2Errors.NoStreamIDAvailable)
+        }
+
+        // Create a frame based stream.
+        let streamBPromise = self.channel.eventLoop.makePromise(of: Channel.self)
+        multiplexer.createStreamChannel(promise: streamBPromise) { channel, streamID in
+            // stream A doesn't have an ID yet.
+            XCTAssertEqual(streamID, HTTP2StreamID(1))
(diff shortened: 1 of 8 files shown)


Tests/NIOHTTP2Tests/HTTP2FramePayloadStreamMultiplexerTests.swift
```diff
@@ -1716,4 +1716,46 @@ final class HTTP2FramePayloadStreamMultiplexerTests: XCTestCase {
...
+        XCTAssertNoThrow(try self.channel.pipeline.addHandler(multiplexer).wait())
+        XCTAssertNoThrow(try self.channel.connect(to: SocketAddress(unixDomainSocketPath: "/whatever"), promise: nil))
+
+        // Create a payload based stream.
+        let streamAPromise = self.channel.eventLoop.makePromise(of: Channel.self)
+        multiplexer.createStreamChannel(promise: streamAPromise) { channel in
+            return channel.eventLoop.makeSucceededFuture(())
+        }
+        self.channel.embeddedEventLoop.run()
+        let streamA = try assertNoThrowWithValue(try streamAPromise.futureResult.wait())
```
Lukasa, line 1734: Should we validate that stream A has no stream ID here?

Tests/NIOHTTP2Tests/SimpleClientServerFramePayloadStreamTests.swift
```diff
@@ -1881,4 +1881,33 @@ class SimpleClientServerFramePayloadStreamTests: XCTestCase {
...
+        let streamB = try assertNoThrowWithValue(try streamBPromise.futureResult.wait())
+
+        let headers = HPACKHeaders([(":path", "/"), (":method", "GET"), (":authority", "localhost"), (":scheme", "https")])
+        let headersFramePayload = HTTP2Frame.FramePayload.headers(.init(headers: headers, endStream: false))
+
+        // Write on 'B' first.
+        XCTAssertNoThrow(try streamB.writeAndFlush(headersFramePayload).wait())
+
+        // Now write on stream 'A'. This would fail on frame-based stream channel.
+        XCTAssertNoThrow(try streamA.writeAndFlush(headersFramePayload).wait())
```
Lukasa, line 1911: Let's add some assertions about stream IDs here?

PR #229, opened 2020-08-06
Fix the -f option in build_podspec.sh.

MrMage:


Lukasa:

PR #231, opened 2020-08-13
Propagate read() through the pipeline.

Lukasa:
Motivation:

NIO's backpressure management systems rely on catching calls to read()
in the ChannelPipeline. This only works if your Channel implementation
actually makes those calls to read()! Sadly, HTTP2StreamChannel did not:
we instead just recursed directly into the Channel's read0().

While I'm here I also noticed that we produced a lot of
channelReadComplete messages in the pipeline unnecessarily. We did this
in the service of having a read "fast-path" where new frames would be
automatically delivered into the ChannelPipeline without being buffered
until channelReadComplete. This fast-path is probably inadvisable: I
don't think it provided a meaningful performance benefit anyway,
especially as the removal of that fast-path allowed us to suppress
unnecessary calls to channelReadComplete, and therefore unnecessary
calls to read() as well.

Modifications:

- Forced autoRead through the pipeline.
- Removed read() fast-path.
- Suppressed channelReadComplete if no frames were delivered.
- Updated tests.

Result:

Clearer I/O path, better auto-read handling.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamChannel.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamChannel.swift
@@ -237,9 +237,7 @@ final class HTTP2StreamChannel<Message: HTTP2FramePayloadConvertible & HTTP2Fram
 
         self.modifyingState { $0.activate() }
         self.pipeline.fireChannelActive()
-        if self.autoRead {
-            self.read0()
-        }
+        self.tryToAutoRead()
         self.deliverPendingWrites()
     }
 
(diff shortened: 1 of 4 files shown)


Lukasa:

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -688,19 +694,19 @@ internal extension HTTP2StreamChannel {
...
-        if self.unsatisfiedRead {
-            // We don't need to account for this frame in the window manager: it's being delivered
-            // straight into the pipeline.
-            self.pipeline.fireChannelRead(NIOAny(message))
-        } else {
+//        if self.unsatisfiedRead {
+//            // We don't need to account for this frame in the window manager: it's being delivered
+//            // straight into the pipeline.
+//            self.pipeline.fireChannelRead(NIOAny(message))
+//        } else {
```
glbrntt, line 701: Looks like these were missed

PR #233, opened 2020-08-24
Avoid emitting WINDOW_UPDATE frames when receiving stream closed

glbrntt:
Motivation:

In some cases it's possible for a stream to emit a WINDOW_UPDATE frame
after receiving close from the network. This is unfortunate because in
the eyes of the `NIOHTTP2Handler` the stream no longer exists which
results in a connection-level error.

Modifications:

- Add a flag to `deliverPendingReads` telling it whether it is allowed
  to emit a WINDOW_UPDATE frame
- Use the above functionality when the stream is closing

Result:

WINDOW_UPDATE frames are no longer emitted from the stream as a result
of delivering pending reads while closing.

### DIFF
--- a/Sources/NIOHTTP2/InboundWindowManager.swift
+++ b/Sources/NIOHTTP2/InboundWindowManager.swift
@@ -25,6 +25,9 @@ struct InboundWindowManager {
     /// The number of bytes of buffered frames.
     private var bufferedBytes: Int = 0
 
+    /// Whether the inbound window is closed. If this is true then no window size changes will be emitted.
+    var closed = false
+
     init(targetSize: Int32) {
         assert(targetSize <= HTTP2FlowControlWindow.maxSize)
         assert(targetSize >= 0)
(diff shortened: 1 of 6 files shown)


Lukasa (commented):
I think passing this as a flag doesn't quite work: if for any reason we get a re-entrant call to `read0` we'll lose track of the state.

Perhaps this should be state on the `InboundWindowManager`: we can flip a bit there to say that there is no point sending window update frames anymore. This can then be done in a bunch of places, including when we receive `END_STREAM`.

Sources/NIOHTTP2/InboundWindowManager.swift
```diff
@@ -67,7 +71,16 @@ struct InboundWindowManager {
         return self.calculateWindowIncrement(windowSize: lastWindowSize + self.bufferedBytes)
     }
 
+    mutating func close() {
```
Lukasa, line 74: I don't think the indirection is needed; lets just have `closed` be an internal property on the window manager and set it directly.

Lukasa (commented):
Broadly looks good, a quick note on the implementation style.

Lukasa (approved):
Fantastic, this looks really good.

PR #234, opened 2020-08-28
Remove needless branch, clean up variable names.

Lukasa:
Motivation:

No point having conditional branches when the branch is always taken!
This doesn't hurt perf at all, the optimiser will trivially see through
this, but it does hurt readability. Relatedly, the single-letter
variable names here aren't great.

Modifications:

- Removed needless branch.
- Gave the variables better names.

Result:

Easier to understand code.


glbrntt (approved):
Nice one.

PR #239, opened 2020-09-18
Shrink NIOHTTP2WindowUpdatedEvent.

### LINKED ISSUE #235: Shrink the size of the NIOHTTP2WindowUpdatedEvent
NIOHTTP2 relies on a number of HTTP/2 user events that are passed down the channel pipeline. These are necessarily stuffed into existential containers for the `Any` type, and so need to fit into 3 words or fewer to avoid a heap allocation.

As these user events exist to pass data through the channel pipeline, they will _always_ be stuffed into an existential after their construction. That means if any of these are wider than 3 words or use an excessively large number of refcounted objects, they should probably become class-backed-structs to reduce the cost of passing them around. Turns out, `NIOHTTP2WindowUpdatedEvent` is 33 bytes wide, which makes it too big. Let's shrink this!

While it stores a pair of `Int?`, it forcibly constrains them in the initializer to fit into the Int32 range. This already basically doesn't work as they can subsequently be edited, so we should probably update this to store some private Int32s and just use computed properties to access the values as `Int`.

Lukasa:
Motivation:

NIOHTTP2WindowUpdatedEvent was larger than 24 bytes, forcing heap
allocations whenever we wanted to put it into an existential. Given that
the range of the two optional integers stored on the type are both
explicitly documented to be the Int32 range, this type can happily just
contain Int32s.

Modifications:

- Changed the backing storage from Int? to Int32?.

Result:

NIOHTTP2WindowUpdatedEvent will be smaller.
Resolves #235.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2UserEvents.swift
+++ b/Sources/NIOHTTP2/HTTP2UserEvents.swift
@@ -45,10 +45,28 @@ public struct NIOHTTP2WindowUpdatedEvent {
     public let streamID: HTTP2StreamID
 
     /// The new inbound window size for this stream, if any. May be nil if this stream is half-closed.
-    public let inboundWindowSize: Int?
+    public var inboundWindowSize: Int? {
+        get {
+            return self._inboundWindowSize.map { Int($0) }
+        }
+        set {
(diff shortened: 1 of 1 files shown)


glbrntt (approved):
Looks good.

PR #242, opened 2020-09-22
Make HTTP2StreamChannel non-generic

glbrntt:
Motivation:

As part of the work done for #214 we made `HTTP2StreamChannel` generic
over the type of data it reads and writes. Unfortunately this cost us a
non-trivial amount of performance.

Modifications:

- Remove the genericism from 'HTTP2StreamChannel' by adding a mode for the
  stream data type the channel is operating on (frames or frame
  payloads), wrapping and unwrapping the data where necessary.
- Simplify the implementation of 'MultiplexerAbstractChannel'
- Remove 'HTTP2FrameConvertible' since it is no longer required

Result:

A perf win, hopefully.


glbrntt:

Lukasa (approved):
Looks reasonable to me. No particular notes here, thanks for cleaning this up!

PR #247, opened 2020-09-24
Add a location to each of 'NIOHTTP2Errors'

glbrntt:
Motivation:

Without knowing where an error was thrown it can be unnecessarily
difficult to work out the reasons an application ended up in an error
state.

Modifications:

- Add a 'file' string to each of the 'NIOHTTP2Errors' denoting the file
  and line where an error was created
- Add static methods to create errors to 'NIOHTTP2Errors' defaulting
  file and line, this is unfortunately necessary as defaulting file and
  line in the `init` makes the defaulted and non-defaulted errors
  indistinguishable, and the compiler will pick the one with fewest
  defaulted values.
- Add equatable conformance manually so as to ignore 'file'
- Add backing storage to errors which no longer be stored in an existential
  container
- Add 'CustomStringConvertible' conformance to the errors using backing
  storage so that the 'file' is actually visible when converting the
  error
- Deprecate the old `init`s
- Fix all deprecation warnings

Result:

- NIOHTTP2Errors carry a 'file' with them indicating where they were
  created


Davidde94 (approved):
LGTM

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -539,9 +545,12 @@ extension NIOHTTP2Handler {
...
             let failedWrites = self.outboundBuffer.streamClosed(cAndCData.streamID)
-            let error = NIOHTTP2Errors.StreamClosed(streamID: cAndCData.streamID, errorCode: .cancel)
-            for promise in failedWrites {
-                promise?.fail(error)
+            // 'NIOHTTP2Errors.StreamClosed' allocates backing storage; avoid that if there are no writes to fail.
+            if !failedWrites.isEmpty {
+                let error = NIOHTTP2Errors.streamClosed(streamID: cAndCData.streamID, errorCode: .cancel)
+                for promise in failedWrites {
+                    promise?.fail(error)
+                }
```
Lukasa, line 553: Is it worth us factoring out this repeated pattern to a helper function?

Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -17,27 +17,220 @@ public protocol NIOHTTP2Error: Equatable, Error { }
...
     /// The outbound frame buffers have become filled, and it is not possible to buffer
     /// further outbound frames. This occurs when the remote peer is generating work
     /// faster than they are consuming the result. Additional buffering runs the risk of
     /// memory exhaustion.
     public struct ExcessiveOutboundFrameBuffering: NIOHTTP2Error {
-        public init() { }
+        /// The file and line where the error was created.
+        public var file: String
+
+        @available(*, deprecated, renamed: "excessiveOutboundFrameBuffering")
```
Lukasa, line 180: I think these need the fully-qualified function name, as this function is not on the same type as its replacement and so the fixit won't work.

Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -49,199 +242,763 @@ public enum NIOHTTP2Errors {
...
         }
 
-        init(from state: NIOHTTP2StreamState) {
-            fromState = state
+        internal init(from state: NIOHTTP2StreamState?, file: String = #file, line: UInt = #line) {
+            self.fromState = state
+            self.file = location(file: file, line: line)
         }
 
         public init() {
```
Lukasa, line 299: Is this supposed to be deprecated?

Lukasa (commented):
This looks really good, thanks for doing this large chunk of work! I've left some notes in the diff, but nothing substantial.

Sources/NIOHTTP2/HTTP2Error.swift
glbrntt: I thought that'd be the case as well, but Xcode turns instances of `NIOHTTP2Errors.ExcessiveOutboundFrameBuffering` into `NIOHTTP2Errors.NIOHTTP2Errors.excessiveOutboundFrameBuffering`!

FWIW all the replacements in this PR were done by applying fixits.

Lukasa: Weird, ok then!

Sources/NIOHTTP2/HTTP2Error.swift
glbrntt: Sure is!

Lukasa (approved):
Nice one @glbrntt, good work!

PR #256, opened 2020-11-18
Add benchmarks and improve benchmarking.

Lukasa:
Motivation:

Performance benchmarking wants accurate numbers, and a wide range of
possible benchmarks. One way to get somewhat more accurate numbers is to
use a tool like Cachegrind to measure the instruction counts of a
program. Cachegrind allows us to isolate a number of noise-introducing
effects, as "number of executed instructions" is a much more stable
number than how long a program takes to run.

When running under cachegrind we want to change the way our performance
benchmarks work. Instead of running lots of times to try to average out
noise, we only want to run once to save ourselves time, as the benchmark
itself should be drastically more stable.

While we're here, let's add some much more isolated benchmarks than our
current "make lots of streams" benchmark. This is basically the same
logic, but it does less of NIO and so makes the impact of NIOHTTP2
substantially more prominent.

Modifications:

- Added a cachegrind build mode to the benchmarks.
- Added new benchmarks and alloc counters.

Result:

More info

### DIFF
--- a/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_1k_requests.swift
+++ b/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_1k_requests.swift
@@ -1,0 +1,215 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2020 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOHPACK
+import NIOHTTP2
+
+struct ServerOnly1KRequestsBenchmark {
+    private let concurrentStreams: Int
+
+    private let channel: EmbeddedChannel
+
+    // This offset is preserved across runs.
+    var streamID = HTTP2StreamID(1)
+
+    // A manually constructed data frame.
+    private var dataFrame: ByteBuffer = {
+        var buffer = ByteBuffer(repeating: 0xff, count: 1024 + 9)
+
+        // UInt24 length, is 1024 bytes.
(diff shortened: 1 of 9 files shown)


Sources/NIOHTTP2PerformanceTester/main.swift
```diff
@@ -48,13 +48,21 @@ public func measure(_ fn: () throws -> Int) rethrows -> [TimeInterval] {
 let limitSet = CommandLine.arguments.dropFirst()
 
 public func measureAndPrint(desc: String, fn: () throws -> Int) rethrows -> Void {
+    #if CACHEGRIND
+    // When CACHEGRIND is set we only run a single requested test, and we only run it once. No printing or other mess.
+    if limitSet.contains(desc) {
+        _ = try fn()
+
+    }
```
glbrntt, line 56:
```suggestion
    }
```

IntegrationTests/tests_01_allocation_counters/test_01_resources/test_1k_requests.swift
```diff
@@ -0,0 +1,215 @@
...
+struct ServerOnly1KRequestsBenchmark {
+    private let concurrentStreams: Int
+
+    private let channel: EmbeddedChannel
+
+    // This offset is preserved across runs.
+    var streamID = HTTP2StreamID(1)
+
+    // A manually constructed data frame.
+    private var dataFrame: ByteBuffer? = {
```
glbrntt, line 28: Why are `dataFrame` and `headersFrame` optional?

glbrntt (commented):
Couple of nits/questions but looks good otherwise. It's a shame we have so much duplication though, it'd be nice if we could remove some of that in the future.

IntegrationTests/tests_01_allocation_counters/test_01_resources/test_1k_requests.swift
Lukasa: Oh, that's a remnant from an older design where I thought I was going to swap them with nil. Will fix.

glbrntt (approved):
Ace.

Lukasa:

Lukasa:
Huh, that's weird. Apparently we reduced the allocation count in NIO such that it is now reducing allocations in one of our benchmarks.

PR #260, opened 2020-11-24
Add benchmark and alloc test for GOAWAY teardown.

Lukasa:
Motivation:

We don't have any benchmarks or alloc counter tests for doing stream
teardown with GOAWAY frames. While this isn't a terribly hot code path
and isn't terribly performance heavy with allocations, it's good to have
more data here.

Modifications:

- Write benchmarks and alloc counter tests that tear down many streams
  with GOAWAY.

Result:

Better testing and alloc counter.


glbrntt (approved):
Looks good to me.

PR #261, opened 2020-11-25
Improve bulk stream teardown.

Lukasa:
Motivation:

The logic for doing bulk stream teardown in the ConnectionStreamsState
was always a bit janky: we'd grab all the stream IDs out of the map,
filter out anything that was too small, insert them one-by-one into a
circular buffer, and then report back. I basically just translated that
directly into StreamMap, but on StreamMap this potentially performs very
poorly, as attempting to find the pivot point becomes a linear scan
through the stream data even when there are _loads_ of streams.

This patch swaps over to add a specialized function for bulk-removal of
streams. The specialized function passes the slice of data to be removed
to the caller so that they can compute on it. It does this by a closure
to try to avoid a CoW where possible.

While we're here I tried to address the loop over the CircularBuffer of
recently reset streams. This flushed out an awkward bug that has existed
in our recently reset streams logic for a very long time, which is that
while we _tried_ to maintain the size of the circular buffer, in
practice we misunderstood the CircularBuffer invariants and so this
never worked. I've added a bunch of unit tests for that logic and fixed
it up.

These two changes together will improve performance in stream bulk
teardown. This is an uncommon case (it only happens on GOAWAY), but it's
still good to make that faster.

Modifications:

- New operation on StreamMap
- New prependWithoutExpanding(contentsOf:) operation
- Unit tests for the circular buffer and stream map logic.

Result:

Better performance on bulk stream teardown. Improves instructions executed in the stream teardown benchmark by 0.2%.


Sources/NIOHTTP2/StreamMap.swift
```diff
@@ -220,6 +243,51 @@ extension CircularBuffer where Element: PerStreamData {
...
+                right = pivot
+            } else if currentStreamID == streamID {
+                // The stream ID is the one after.
+                return self.index(after: pivot)
+            } else {
+                left = self.index(after: pivot)
+            }
+        }
+
+        // ?
```
glbrntt, line 288: Not a confidence inspiring comment!

glbrntt (approved):
LGTM

Sources/NIOHTTP2/StreamMap.swift
Lukasa: Hah! I left that for myself when I was unsure if the logic was quite right. I convinced myself it was.

Sources/NIOHTTP2/StreamMap.swift
```diff
@@ -130,6 +130,29 @@ struct StreamMap<Element: PerStreamData> {
...
+                                                  _ block: (CircularBuffer<Element>.SubSequence) -> Void) {
+        switch role {
+        case .client:
+            let index = self.clientInitiated.findIndexForFirstStreamIDLargerThan(streamID)
+            block(self.clientInitiated[index...])
+            self.clientInitiated.removeSubrange(index...)
+        case .server:
+            let index = self.serverInitiated.findIndexForFirstStreamIDLargerThan(streamID)
+            block(self.serverInitiated[index...])
+            self.serverInitiated.removeSubrange(index...)
```
PeterAdams-A, line 153: The duplication in places like this does look pretty sad.

PeterAdams-A (approved):
Looks good.

PR #262, opened 2020-11-25
Move the huffman benchmark base64 out of init

Lukasa:
Motivation:

The init() of all benchmarks will execute, even when we filter a
benchmark run down to a single benchmark. We should avoid doing heavy
work there, because it can confuse us when we look at traces elsewhere.

Modifications:

- Move some heavy work in the huffman benchmark base64 decode out of
  init().
- Prevent setUp() and tearDown() from running if a test is filtered out.
- Avoid even creating benchmarks unless they're filtered in.

Result:

Less confusing benchmark traces.

### DIFF
--- a/Sources/NIOHTTP2PerformanceTester/Benchmark.swift
+++ b/Sources/NIOHTTP2PerformanceTester/Benchmark.swift
@@ -17,13 +17,3 @@ protocol Benchmark: class {
     func tearDown()
     func run() throws -> Int
 }
-
-func measureAndPrint<B: Benchmark>(desc: String, benchmark bench: B) throws {
-    try bench.setUp()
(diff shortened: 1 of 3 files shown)


Lukasa:
As a note: in our case, without the base64 jazz needed to set up the HPACK cases, we see a _17%_ speedup in the 100 concurrent streams benchmark. We were wasting _so much time_ doing base64 nonsense in these benchmarks. All prior speedup results I've quoted in the past few days should be proportionally increased to account for the fact that they were swamped by this nonsense work we shouldn't have been doing.

glbrntt (approved):

PeterAdams-A (approved):

PR #264, opened 2020-11-30
Cheaper calculation of flushes.

Lukasa:
Motivation:

The outbound flow control buffer currently does a linear scan through
the pending data objects to work out how many bytes were buffered. Let's
burn an integer and just calculate these on the way in, making flushes
way cheaper.

Modifications:

- Update the number of bytes when we buffer the write, and make flushes
  cheaper.

Result:

Moderate performance improvement in high throughput server benchmarks
(around 4%).


glbrntt (approved):
Looks good.

PR #265, opened 2020-12-03
Reduce the cost of flushing

Lukasa:
Motivation:

The OutboundFlowControlBuffer is responsible for managing DATA and
HEADERS frames for all streams in order to ensure that we obey HTTP/2
flow control rules, even when users don't. This object is therefore on
the data path for the vast majority of frames, and is extremely
performance sensitive.

Unfortunately, the current algorithm has some nasty performance cliffs.
Right now it has the awkward property of making flushes take an amount
of time linear in the number of streams on the connection, not the
number of streams that had had frames sent on them. The result is that
connections with highly parallel numbers of streams see extremely poor
performance when flushing, a problem that is exacerbated if other parts
of the HTTP/2 stack fail to coalesce those flushes.

We can clean this up in a few ways, but the easiest way is to just keep
track of which streams have done I/O since the last flush. While we're
there, we can make a few other optimisations to ensure we only flush
data on streams we are actually processing, to further reduce the risk
of wasting our time processing unnecessary work.

Modifications:

- Add a Set to keep track of where we've done writes since the last
   flush.
- Streams whose flow control window goes to zero stop being writable.
- Cheaper and more accurate calculation of zero-length writes, which
   avoids the possibility of awkward stalls with zero-length DATA frames
   and END_STREAM.
- Stop doing math about the number of flushed bytes, we don't use that
   information any more.

Results:

An approximately 10% performance gain on the 100 concurrent streams
server-only benchmark. The relative performance gain shifts with the
number of active streams, but it should be positive in all cases: no regression
is observed on the 1 concurrent stream equivalent test.

### DIFF
--- a/Sources/NIOHTTP2/Frame Buffers/OutboundFlowControlBuffer.swift
+++ b/Sources/NIOHTTP2/Frame Buffers/OutboundFlowControlBuffer.swift
@@ -14,12 +14,44 @@
 import NIO
 
 /// A structure that manages buffering outbound frames for active streams to ensure that those streams do not violate flow control rules.
+///
+/// This buffer is an extremely performance sensitive part of the HTTP/2 stack. This is because all outbound data passes through it, and
+/// all application data needs to be actively buffered by it. The result of this is that state management is extremely expensive, and we
+/// need efficient algorithms to process them.
+///
+/// The core of the data structure are a collection of `StreamFlowControlState` objects. These objects keep track of the flow control window
+/// available for a given stream, as well as any pending writes that may be present on the stream. The pending writes include both DATA and
+/// HEADERS frames, as if we've buffered any DATA frames we need to queue HEADERS frames up behind them.
+///
+/// Data is appended to these objects on write(). Each write() will trigger both a lookup in the stream data buffers _and_ an append to a
+/// data buffer. It is therefore deeply important that both of these operations are as cheap as possible.
+///
+/// However, these operations are fundamentally constant in size. Once we have found the stream data buffer, we do not need to worry about
+/// all the others. The cost of the write() operation is therefore no more expensive than the cost of the lookup. The same is unfortunately
+/// not true for flush().
+///
+/// When we get a flush(), we need to update a bunch of state. In particular, we need to record any previously-written frames that are
+/// now flushed, as well as compute which streams may have become writable as a result of the flush call. In early versions of this code
+/// we would do a linear scan across _all buffers_, check whether they were writable before, mark their flush point, check if they'd become
+/// writable, and then store them in the set of flushable streams. This was monstrously expensive, and worse still the cost was not proportional
+/// to the amount of writing done but to the amount of flushing done and the number of active streams.
+///
+/// The current design avoids this by having the `StreamFlowControlState` be much more explicit about when it transitions from non-writable to
+/// writable and back again. This allows us to keep an Array of writable streams that we use to avoid needing to touch all the streams whenever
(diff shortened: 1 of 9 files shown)


Tests/NIOHTTP2Tests/SimpleClientServerFramePayloadStreamTests.swift
```diff
@@ -198,9 +198,14 @@ class SimpleClientServerFramePayloadStreamTests: XCTestCase {
     /// Establish a basic HTTP/2 connection.
     func basicHTTP2Connection(clientSettings: HTTP2Settings = nioDefaultSettings,
                               serverSettings: HTTP2Settings = nioDefaultSettings,
+                              maximumBufferedControlFrames: Int = 10000,
```
Lukasa, line 201: The changes in this file makes these tests run faster: I was sick of waiting for them.

scripts/cachegrindify.sh
```diff
@@ -0,0 +1,57 @@
...
+##
+## Copyright (c) 2020 Apple Inc. and the SwiftNIO project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
```
Lukasa, line 14: This is a helper script I've added to do broader-scale cachegrind comparison benchmarking.

Sources/NIOHTTP2/Frame Buffers/OutboundFlowControlBuffer.swift
```diff
@@ -210,17 +267,72 @@ private struct StreamFlowControlState: PerStreamData {
...
+    /// Removes all pending writes, invalidating this structure as it does so.
+    mutating func evacuatePendingWrites() -> MarkedCircularBuffer<DataBuffer.BufferElement> {
+        return self.dataBuffer.evacuatePendingWrites()
+    }
+
+    func failAllWrites(error: ChannelError) {
+        self.dataBuffer.failAllWrites(error: error)
+    }
+
+    mutating func nextWrite(maxSize: Int) -> (DataBuffer.BufferElement, WritabilityState) {
```
glbrntt, line 313: IIUC this should only be called if we're marked, worth a comment/assertion to that end?

glbrntt (approved):
I think this looks good.

Sources/NIOHTTP2/Frame Buffers/OutboundFlowControlBuffer.swift
```diff
@@ -210,17 +267,72 @@ private struct StreamFlowControlState: PerStreamData {
...
+        self.currentWindowSize = newSize
+
+        // If we have no marked writes, nothing changed.
+        guard self.dataBuffer.hasMark else {
+            return .unchanged
+        }
+
+        if oldWindowSize <= 0 && self.currentWindowSize > 0 {
+            // Window opened. We can now write.
+            return .changed(newValue: true)
```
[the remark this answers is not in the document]
Lukasa: Yes, this is correct. The writability notion here is entirely regarding the HTTP/2 flow control window, and has nothing to do with the stream channel itself. Nothing in this object handles stream channel logic in any way.

Sources/NIOHTTP2/Frame Buffers/OutboundFlowControlBuffer.swift
Lukasa: Yeah, seems like a good idea to me.

PR #269, opened 2020-12-04
Make HPACKHeaders inlinable.

Lukasa:
Motivation:

HPACKHeaders is a trivial data structure (a wrapper around Array) that
is present in the public API of all NIO HTTP/2 programs. This makes it a
tempting target for adding inlinable calls, primarily as a way to drive
down ARC. The fact that strings frequently travel across the API
boundary means that otherwise innocuous use-cases can trigger unexpected
ARC traffic.

Making as much of the implementation inlinable as possible reduces a lot
of this overhead and opens up the possibility for better performance in
a number of parts of the code, including in the core HTTP/2 protocol
engine, which operates extensively on this data type.

Modifications:

- Make basically everything inlinable.

Result:

Our benchmarks are about 0.7% faster.

### DIFF
--- a/Sources/NIOHPACK/HPACKHeader.swift
+++ b/Sources/NIOHPACK/HPACKHeader.swift
@@ -22,10 +22,12 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral {
     internal var headers: [HPACKHeader]
 
     // see 8.1.2.2. Connection-Specific Header Fields in RFC 7540
+    @usableFromInline
     static let illegalHeaders: [String] = ["connection", "keep-alive", "proxy-connection",
                                            "transfer-encoding", "upgrade"]
 
     /// Constructor that can be used to map between `HTTPHeaders` and `HPACKHeaders` types.
+    @inlinable
     public init(httpHeaders: HTTPHeaders, normalizeHTTPHeaders: Bool) {
         if normalizeHTTPHeaders {
             self.headers = httpHeaders.map { HPACKHeader(name: $0.name.lowercased(), value: $0.value) }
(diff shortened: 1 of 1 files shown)


Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -486,14 +528,16 @@ extension String.UTF8View {
             }
     }
 
+    @inlinable
     @inline(never)
-    private func _compareCaseInsensitiveASCIIBytesSlowPath(to other: String.UTF8View) -> Bool {
+    func _compareCaseInsensitiveASCIIBytesSlowPath(to other: String.UTF8View) -> Bool {
```
PeterAdams-A, line 533: This looks odd - inalienable but never?

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -461,7 +502,8 @@ extension String.UTF8View {
     ///
     /// - Parameter bytes: The string constant in the form of a collection of `UInt8`
     /// - Returns: Whether the collection contains **EXACTLY** this array or no, but by ignoring case.
-    fileprivate func compareCaseInsensitiveASCIIBytes(to other: String.UTF8View) -> Bool {
+    @inlinable
```
PeterAdams-A, line 505: I've not read the rules but I don't see any logical reason why something shouldn't be both inlineable and fileprivate.

PeterAdams-A (approved):
Generally looks good but I've made a couple of comments.

Sources/NIOHPACK/HPACKHeader.swift
Lukasa: Yup. This is a logically sensible stanza. `@inlinable` means the definition is available to the compiler when compiling the module using the code. This allows, for example, specialization. But the method may not be _inlined_. In this case the primary use is to let the compiler see through the call and observe that we never write to any memory here.

Sources/NIOHPACK/HPACKHeader.swift
Lukasa: Totally reasonable except that the rule is it has to be `internal` or higher.

glbrntt (approved):

PR #270, opened 2020-12-04
Remove some generics that were triggering ARC

Lukasa:
Motivation:

To save myself some lines of code I wrote a function that was generic
over a protocol to copy out a pair of fields from a wide range of types.
It turns out this generic was triggering a retain/release on the state
machine states, which triggers 12 ARC operations total (6 retain/release
pairs).

Removing this generic and just, you know, calling a function with some
arguments turns out to save us a bunch of this ARC traffic.

Modifications:

- Remove the generic initializer of StateMachineResultWithEffect.

Result:

Approximately 1% performance improvement in hot benchmarks.


PeterAdams-A (approved):
Looks good.

PR #271, opened 2021-01-07
Avoid unnecessary futures in stream creation

glbrntt:
Motivation:

The stream channel configure function uses a chain of futures where the
same can be achieved in fewer futures.

Modifications:

- Merge the future chains where possible in HTTP2StreamChannel.configure

Result:

A 1.3% reduction in instructions in server_only_10k_requests_1_concurrent

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamChannel.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamChannel.swift
@@ -178,39 +178,63 @@ final class HTTP2StreamChannel: Channel, ChannelCore {
         self._pipeline = ChannelPipeline(channel: self)
     }
 
-    internal func configure(initializer: ((Channel, HTTP2StreamID) -> EventLoopFuture<Void>)?, userPromise promise: EventLoopPromise<Channel>?){
+    internal func configure(initializer: ((Channel, HTTP2StreamID) -> EventLoopFuture<Void>)?, userPromise promise: EventLoopPromise<Channel>?) {
         assert(self.streamDataType == .frame)
         // We need to configure this channel. This involves doing four things:
         // 1. Setting our autoRead state from the parent
         // 2. Calling the initializer, if provided.
         // 3. Activating when complete.
         // 4. Catching errors if they occur.
-        self.getAutoReadFromParent().flatMap { autoRead in
-            self.autoRead = autoRead
-            // This initializer callback can only be invoked if we already have a stream ID.
-            // So we force-unwrap here.
-            return initializer?(self, self.streamID!) ?? self.eventLoop.makeSucceededFuture(())
-        }.map {
-            self.postInitializerActivate(promise: promise)
-        }.whenFailure { error in
-            self.configurationFailed(withError: error, promise: promise)
+        self.getAutoReadFromParent().whenComplete { autoReadResult in
+            switch autoReadResult {
+            case .success(let autoRead):
+                self.autoRead = autoRead
+                if let initializer = initializer {
+                    // This initializer callback can only be invoked if we already have a stream ID.
+                    // So we force-unwrap here.
+                    initializer(self, self.streamID!).whenComplete { result in
+                        switch result {
+                        case .success:
+                            self.postInitializerActivate(promise: promise)
+                        case .failure(let error):
+                            self.configurationFailed(withError: error, promise: promise)
+                        }
+                    }
+                } else {
+                    self.postInitializerActivate(promise: promise)
+                }
+            case .failure(let error):
+                self.configurationFailed(withError: error, promise: promise)
+            }
         }
     }
 
-    internal func configure(initializer: ((Channel) -> EventLoopFuture<Void>)?, userPromise promise: EventLoopPromise<Channel>?){
+    internal func configure(initializer: ((Channel) -> EventLoopFuture<Void>)?, userPromise promise: EventLoopPromise<Channel>?) {
         assert(self.streamDataType == .framePayload)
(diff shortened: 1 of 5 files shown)


Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -178,39 +178,63 @@ final class HTTP2StreamChannel: Channel, ChannelCore {
...
-        self.getAutoReadFromParent().flatMap { autoRead in
-            self.autoRead = autoRead
-            // This initializer callback can only be invoked if we already have a stream ID.
-            // So we force-unwrap here.
-            return initializer?(self, self.streamID!) ?? self.eventLoop.makeSucceededFuture(())
-        }.map {
-            self.postInitializerActivate(promise: promise)
-        }.whenFailure { error in
-            self.configurationFailed(withError: error, promise: promise)
+        self.getAutoReadFromParent().whenComplete { autoReadResult in
```
weissi, line 188: This seems like an unfortunate patch to me. It's harder to read, more error prone (two calls to `postInitializerActivate...` and `configurationFailed...` instead of one) for a 1% reduction in instructions.

- Is this a measurable improvement in actual wall clock performance?
- do we have test coverage for all branches (because there are many more now)?
- shouldn't the old code become just as fast once we start caching `ELF<Void>`?

weissi: Also this seems to predominantly reduce allocations so if we believe this is important, we should have an alloc counter test covering this?

Lukasa: Re 3: It won't be just as fast, if only because more closures have to be constructed. And we do have alloc counter tests that cover this.

weissi:

Sources/NIOHTTP2/HTTP2StreamChannel.swift
Lukasa: The alloc counter tests are why the unit tests failed :wink:

Lukasa:
Sadly the perf tests have the same baseline issue as with swift-nio, so we're measuring against perf from either 3 months ago or 1 month ago.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
weissi: Re 3: fair enough, there'll be a small penalty. But is it worth regressing on code readability? Stream creation is obviously a hot path so I can see this being justified but I still struggle with that. The perf bot should hopefully tell us how much we gain in actual performance here.

weissi:
> Sadly the perf tests have the same baseline issue as with swift-nio, so we're measuring against perf from either 3 months ago or 1 month ago.

Yeah, we should quickly run an empty PR. Problem is worse though: Doesn't seem to produce any results :(. Are we taking changes like this without knowing if/how much perf we gain?

Lukasa:
Oh shiny, they're actually _worse_ than they were a month ago. :grin:

weissi:
> Oh shiny, they're actually _worse_ than they were a month ago. 😁

Where do you see the numbers? Or is _worse_ == they broke?

Lukasa:
Yeah, worse is "it's now totally broken".

Lukasa:
We reduced allocs, so my inclination is to take the patch _anyway_. It'll be a perf win.

Lukasa:
With that said, there's no rush to merge this.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
glbrntt: > This seems like an unfortunate patch to me. It's harder to read, more error prone (two calls to `postInitializerActivate...` and `configurationFailed...` instead of one) for a 1% reduction in instructions.

I agree it's unfortunate; I saw it and thought it _might_ be worth doing so threw up a PR.

weissi:
> We reduced allocs, so my inclination is to take the patch _anyway_. It'll be a perf win.

I guess. There used to be 3 closures that capture, after the patch it's just one (even after caching ELF<Void>). So this should indeed reduce allocs by 2 per stream. For gRPC this shouldn't matter at all (because protobuf allocates _a lot_) but there could be others for who 2 allocs per stream do matter... Your call.

Lukasa:
One of grpc's major costs is stream creation, which is _extremely_ allocation heavy. This is not a substantial dent in that, but in practice this can often cost more than the proto work in unary RPCs.

Lukasa:

glbrntt:
@Lukasa / @weissi what's happening with this PR?

weissi:
@glbrntt it makes me a little sad but even after caching `ELF<Void>` it'll still save a few allocs. Given that it's NIO code I'm happy for this to go in. But Cory's call.

What's surprising is that the biggest alleged win is in the stream teardown benchmark which can't really be due to this one, can it?

Lukasa:
> What's surprising is that the biggest alleged win is in the stream teardown benchmark which can't really be due to this one, can it?

Gotta create the streams to tear them down! The stream teardown does lots of stream creation and stream termination without running actual load on those streams. The multiplexer is still in the pipeline, so those streams still come up. We even add a handler to them just to prove they're there.

Lukasa (approved):

glbrntt:
> @glbrntt it makes me a little sad but even after caching `ELF<Void>` it'll still save a few allocs. 

It makes me sad too; but I think it's fair tradeoff given it's a hot path.

PR #274, opened 2021-01-22
Use welcoming language

fabianfett:
To comply with the new [Swift code of conduct](https://swift.org/code-of-conduct/), let's make the language a little more welcoming.

### Modifications:

- Replaced usages of "sanity" with "soundness"
- Added "sanity" to the list of unacceptable terms.

### Result:

- Swift NIO HTTP2 is nicer.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -22,9 +22,9 @@ services:
       - CAP_NET_RAW
       - CAP_NET_BIND_SERVICE
 
-  sanity:
+  soundness:
     <<: *common
-    command: /bin/bash -xcl "./scripts/sanity.sh"
+    command: /bin/bash -xcl "./scripts/soundness.sh"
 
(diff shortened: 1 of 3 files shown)


Lukasa:

docker/docker-compose.yaml
```diff
@@ -22,9 +22,9 @@ services:
       - CAP_NET_RAW
       - CAP_NET_BIND_SERVICE
 
-  sanity:
```
Lukasa, line 25: As with NIO, I believe this change requires coordination with @tomerd in order to ensure the CI continues to function effectively.

Lukasa (approved):
Generally fine modulo concerns with CI.

Lukasa:
Oh, it seems that @tomerd has already made that change comprehensively across our CI systems. I suspect this may actually have broken some repos.

PR #281, opened 2021-03-09
Update allocation count requirements

glbrntt:
Motivation:

NIO 2.26.0 saved us some allocations, some of our limits are now too
high and CI fails as a result.

Modifications:

Adjust the allocation counter limits.

Result:

CI passes.


Lukasa (approved):

glbrntt:
nightly CI failure is expected, merging over it

PR #282, opened 2021-03-15
Add sync option support to HTTP2StreamChannel

glbrntt:
Motivation:

SwiftNIO 2.27.0 added support for synchronous channel options allowing
callers to get options synchronously -- saving a future allocation -- if
they know they're on the correct event loop. 'HTTP2StreamChannel' should
support this.

Modifications:

- Add 'HTTP2StreamChannel.SynchronousOptions' with conformance to
  'NIOSynchronousChannelOptions'

Result:

Callers can get and set options synchronously on 'HTTP2StreamChannel'

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamChannel.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamChannel.swift
@@ -344,31 +344,31 @@ final class HTTP2StreamChannel: Channel, ChannelCore {
     }
 
     func setOption<Option: ChannelOption>(_ option: Option, value: Option.Value) -> EventLoopFuture<Void> {
-        if eventLoop.inEventLoop {
+        if self.eventLoop.inEventLoop {
             do {
-                return eventLoop.makeSucceededFuture(try setOption0(option, value: value))
+                return self.eventLoop.makeSucceededFuture(try self.setOption0(option, value: value))
             } catch {
-                return eventLoop.makeFailedFuture(error)
+                return self.eventLoop.makeFailedFuture(error)
             }
         } else {
-            return eventLoop.submit { try self.setOption0(option, value: value) }
+            return self.eventLoop.submit { try self.setOption0(option, value: value) }
         }
     }
 
     public func getOption<Option: ChannelOption>(_ option: Option) -> EventLoopFuture<Option.Value> {
-        if eventLoop.inEventLoop {
+        if self.eventLoop.inEventLoop {
             do {
-                return eventLoop.makeSucceededFuture(try getOption0(option))
+                return self.eventLoop.makeSucceededFuture(try self.getOption0(option))
             } catch {
-                return eventLoop.makeFailedFuture(error)
+                return self.eventLoop.makeFailedFuture(error)
             }
         } else {
-            return eventLoop.submit { try self.getOption0(option) }
+            return self.eventLoop.submit { try self.getOption0(option) }
         }
     }
 
     private func setOption0<Option: ChannelOption>(_ option: Option, value: Option.Value) throws {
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -344,31 +344,31 @@ final class HTTP2StreamChannel: Channel, ChannelCore {
...
             }
         } else {
-            return eventLoop.submit { try self.getOption0(option) }
+            return self.eventLoop.submit { try self.getOption0(option) }
         }
     }
 
     private func setOption0<Option: ChannelOption>(_ option: Option, value: Option.Value) throws {
-        assert(eventLoop.inEventLoop)
+        self.eventLoop.assertInEventLoop()
```
Lukasa, line 371: As these are now called publicly through APIs where the user is required to get them correct, I think we should lift these up to preconditions.

Lukasa (approved):
LGTM! Want to follow-up with a separate patch to _use_ the sync options API on the underlying channel to get at the autoRead channel option when it's available?

Lukasa (changes requested):
Sorry, per earlier note I do have feedback here.

glbrntt:
All the CI -- except for the nightly job, which is expected to fail -- run without failures until they hit an issue reporting the status.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -878,3 +878,34 @@ extension HTTP2StreamChannel {
         return "HTTP2StreamChannel(streamID: \(String(describing: self.streamID)), isActive: \(self.isActive), isWritable: \(self.isWritable))"
     }
 }
+
+extension HTTP2StreamChannel {
+    public struct SynchronousOptions: NIOSynchronousChannelOptions {
```
Lukasa, line 883: Does this struct need to be `public`?

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -878,3 +878,34 @@ extension HTTP2StreamChannel {
...
+        /// - Important: Must be called on the `EventLoop` the `Channel` is running on.
+        @inlinable
+        public func setOption<Option: ChannelOption>(_ option: Option, value: Option.Value) throws {
+            try self.channel.setOption0(option, value: value)
+        }
+
+        /// Get the value of `option` for this `Channel`.
+        ///
+        /// - Important: Must be called on the `EventLoop` the `Channel` is running on.
+        @inlinable
```
Lukasa, line 901: We can remove the `inlinable` on these two, we can't inline through this.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -878,3 +878,34 @@ extension HTTP2StreamChannel {
...
+        public func getOption<Option: ChannelOption>(_ option: Option) throws -> Option.Value {
+            return try self.channel.getOption0(option)
+        }
+    }
+
+    /// Returns a view of the `Channel` exposing synchronous versions of `setOption` and `getOption`.
+    public final var syncOptions: NIOSynchronousChannelOptions? {
```
Lukasa, line 908: `HTTP2StreamChannel` is a `final` class so we don't need `final` here.

Lukasa (changes requested):

Lukasa (approved):

Lukasa:

Lukasa:
Merging over the nightly failure.

PR #283, opened 2021-03-15
Use sync options when creating a stream channel

glbrntt:
Motivation:

SwiftNIO 2.27.0 added synchronous channel options. We can use these if
available when creating a stream channel to avoid a future allocation.

Modifications:

- 'getAutoReadFromParent' is synchronous when possible and now takes a
  closure to handle the result.
- Update allocation counts (and sort them to make updating them easier,
  this also includes switching the order we run test_1k_requests
  interleaved and noninterleaved)

Result:

Fewer allocations.


Lukasa (approved):
LGTM, nice!

PR #284, opened 2021-03-16
Stop using deprecated API in allocation counting tests

glbrntt:
Motivation:

Some of the allocation counting tests used frame-based stream channels
rather than the frame payload based ones. These were deprecated some
time ago.

Modifications:

- Update test handlers to be `HTTP2Frame.FramePayload` based
- Update the 'create_client_stream_channel' test to count streams rather
  than summing their stream ID (as stream IDs aren't assigned until the
  stream channel delivers pending writes for the first time).
- Update allocation limits

Result:

Allocation counting tests don't use deprecated APIs.

### DIFF
--- a/docker/docker-compose.1604.51.yaml
+++ b/docker/docker-compose.1604.51.yaml
@@ -19,8 +19,8 @@ services:
       - MAX_ALLOCS_ALLOWED_1k_requests_interleaved=53000
       - MAX_ALLOCS_ALLOWED_1k_requests_noninterleaved=52000
       - MAX_ALLOCS_ALLOWED_client_server_h1_request_response=336000
-      - MAX_ALLOCS_ALLOWED_client_server_request_response=300000
-      - MAX_ALLOCS_ALLOWED_create_client_stream_channel=55000
+      - MAX_ALLOCS_ALLOWED_client_server_request_response=301000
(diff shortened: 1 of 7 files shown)


docker/docker-compose.1604.51.yaml
```diff
@@ -19,8 +19,8 @@ services:
       - MAX_ALLOCS_ALLOWED_1k_requests_interleaved=53000
       - MAX_ALLOCS_ALLOWED_1k_requests_noninterleaved=52000
       - MAX_ALLOCS_ALLOWED_client_server_h1_request_response=336000
-      - MAX_ALLOCS_ALLOWED_client_server_request_response=300000
-      - MAX_ALLOCS_ALLOWED_create_client_stream_channel=55000
+      - MAX_ALLOCS_ALLOWED_client_server_request_response=310000
```
glbrntt, line 22: The regression here is the 'pending streams' dictionary in the multiplexer resizing (it isn't used for the frame-based stream channels).

Lukasa (approved):

devhot1984 (commented):
# [Developer](https://www.icloud.com/shortcuts/b395f0af718d4498b903807515ab7c77) #
**[Official Site](https://dandeveloper.carrd.co)**

PR #288, opened 2021-05-27
Add fuzzing harness to swift-nio-http2.

Lukasa:
Motivation:

Fuzzing is great, we should have a good sensible place to centralise a
harness.

Modifications:

- Add a fuzzing harness.

Result:

There's a common way to do fuzzing.

### DIFF
--- a/FuzzTesting/Sources/FuzzHTTP2/main.swift
+++ b/FuzzTesting/Sources/FuzzHTTP2/main.swift
@@ -1,0 +1,58 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+// Sources/protoc-gen-swift/main.swift - Protoc plugin main
+//
+// Copyright (c) 2020 - 2021 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+
+
+import Foundation
+
+import NIO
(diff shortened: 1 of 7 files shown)


FuzzTesting/Sources/FuzzHTTP2/main.swift
```diff
@@ -0,0 +1,58 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+// Sources/protoc-gen-swift/main.swift - Protoc plugin main
+//
+// Copyright (c) 2020 - 2021 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
```
glbrntt, line 20: Intentional? / Should we add an entry to NOTICES.txt?

scripts/soundness.sh
```diff
@@ -18,7 +18,7 @@ here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 function replace_acceptable_years() {
     # this needs to replace all acceptable forms with 'YEARS'
-    sed -e 's/2017-20[12][890]/YEARS/g' -e 's/2019/YEARS/g' -e 's/2020/YEARS/g'
+    sed -e 's/2017-20[12][890]/YEARS/g' -e 's/2019/YEARS/g' -e 's/2020/YEARS/g' -e 's/2021/YEARS/g'
```
glbrntt, line 21: I'm surprised we got this far into the year without this change!

glbrntt (approved):
Was `FuzzTesting/FailCases/.empty` meant to be checked in? Looks good otherwise.

Lukasa:
Yeah, I checked `.empty` in on purpose to make the directory exist. Git doesn't create empty directories on clone.

FuzzTesting/Sources/FuzzHTTP2/main.swift
Lukasa: Done.

NOTICES.txt
```diff
@@ -0,0 +1,17 @@
+
```
glbrntt, line 1: There's already a NOTICE.txt, we should merge them

Lukasa: Good catch!

PR #290, opened 2021-06-08
Update our allocation counter numbers.

Lukasa:
Motivation:

Our 5.5 allocation counts were off. Rather than just update that by hand
we should bring over some tooling to do it automatically.

Modifications:

- Bring over and adapt the update-alloc-limits scripts from NIO.
- Run them and commit the updates.

Result:

CI should pass again.

### DIFF
--- a/dev/update-alloc-limits-to-last-completed-ci-build
+++ b/dev/update-alloc-limits-to-last-completed-ci-build
@@ -1,0 +1,47 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
+## Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+set -eu
+set -o pipefail
+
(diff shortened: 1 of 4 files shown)


dev/update-alloc-limits-to-last-completed-ci-build
```diff
@@ -0,0 +1,47 @@
...
+set -eu
+set -o pipefail
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+url_prefix=${1-"https://ci.swiftserver.group/job/swift-nio-http2-2-"}
+target_repo=${2-"$here/.."}
+tmpdir=$(mktemp -d /tmp/.last-build_XXXXXX)
+
+for f in swift51 swift52 swift53 swift54 swift55 nightly; do
```
glbrntt, line 25: With https://github.com/apple/swift-nio/pull/1860 in mind, should we drop `swift51` from this list?

Lukasa: For now I'd like to keep it: when we make the equivalent PR here, we'll need to remove 51 from this script, just as the linked PR does.

glbrntt: Should we add `swift50` then since we still track allocations for it in http/2?

Lukasa: Nope, we'll just stop tracking allocs for it.

glbrntt: Even better!

glbrntt (approved):

PR #296, opened 2021-08-16
Enable tests for Android

buttaface:
Motivation:

Run the tests on Android.

Modifications:

- Enable the tests on Android.
- Replace `/tmp` with `FileManager.default.temporaryDirectory`.

Result:

All the tests pass on Android AArch64 and x86_64.

I used [an earlier version of this patch to add this package to my daily Android CI](https://github.com/buttaface/swift-android-sdk/commit/ff03a7f390e1a282924125d56ca18214b4d36b43#diff-57060aace578293008118ccd2c43be3b8fb7440b36e018c3ab9fd610b74296b1), though I had to disable 32-bit armv7 compilation because [some tests wouldn't compile](https://github.com/buttaface/swift-android-sdk/runs/3339767095?check_suite_focus=true#step:27:153). The issue appears to be some tests and [the functions they call that take `UInt`/`Int` but are passed in larger numbers](https://github.com/apple/swift-nio-http2/blob/37919ba1b7c904c96c7d8d7df736ee1e060df61e/Tests/NIOHTTP2Tests/SimpleClientServerFramePayloadStreamTests.swift#L1890), which works for `Int64` but not `Int32`.


Lukasa:
@swift-nio-bot add to allowlist

Lukasa (approved):
Seems straightforwardly effective to me

PR #297, opened 2021-08-18
Use non-deprecated API for removing handlers

fabianfett:
### Motivation:

removeHandlers(channel:) was deprecated in NIO 2.32.0.

### Modifications:

- Raise minimum required NIO version to 2.32.0
- Use removeHandlers(pipeline:)

### Result:

We don't use deprecated API.


glbrntt (approved):
Ta!

fabianfett:
Compiler crashed in `main-nightly`. This is a non required test. Merging.

PR #298, opened 2021-09-03
Fix doc generation and jazzy version

glbrntt:
From: https://github.com/apple/swift-nio-extras/pull/141

Motivation:

The combination of Ruby and Jazzy are throwing up issues in CI on older
Ubuntu versions.

Modifications:

- Only install Jazzy on focal
- Fix doc generation script to use more recent source-kitten syntax

Results:

- CI doesn't complain
- Docs get built


fabianfett (approved):
LGTM

PR #307, opened 2021-09-29
Ignore empty values when returning canonical form of header values

glbrntt:
Motivation:

The return value from `HPACKHeaders.subscript(canonicalForm:)` is
inconsistent with respect to returning empty strings.

If a value for a header field is an empty string then the canonical form
will not return that empty string. However, if a value is a string
containing only whitespace then an empty string will be returned.

Modifications:

- Filter out empty substrings after trimming whitespace
- Add a test

Result:

- `HPACKHeaders.subscript(canonicalForm:)` does not produce empty strings

### DIFF
--- a/Sources/NIOHPACK/HPACKHeader.swift
+++ b/Sources/NIOHPACK/HPACKHeader.swift
@@ -49,7 +49,7 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral {
     public init(httpHeaders: HTTPHeaders) {
         self.init(httpHeaders: httpHeaders, normalizeHTTPHeaders: true)
     }
-    
+
     /// Construct a `HPACKHeaders` structure.
     ///
     /// The indexability of all headers is assumed to be the default, i.e. indexable and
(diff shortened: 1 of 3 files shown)


Lukasa (approved):

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -260,7 +260,15 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral {
...
-        return result.flatMap { $0.split(separator: ",") }.map { String($0._trimWhitespace()) }
+        return result.flatMap {
+            $0.split(separator: ",")
+        }.map {
+            $0._trimWhitespace()
+        }.filter { // `split(separator:)` drops empty strings, we should too.
+            !$0.isEmpty
+        }.map {
+            String($0)
+        }
```
Lukasa, line 271: We may need to rewrite this to use lazy iteration now.

glbrntt: Yup, looks like a bunch of allocation regressions.

Lukasa (approved):

PR #304, opened 2021-09-27
Improve performance of `HPACKHeaders.subscript(canonicalForm:)`

glbrntt:
Motivation:

When getting the canonical form of header values any ascii whitespace
is stripped from the values. The current implementation does this by
doing equality checks on Character and dropping them from a Substring.

Modifications:

- Update the Substring._trimWhitespace() function to trim on a UTF8 view
- Add test

Result:

Retrieving the canonical form of header values is cheaper, significantly
so when values contain a lot of leading or trailing whitespace.

PR #308, opened 2021-09-29
Use case-insensitive comparision rather than lowercasing

glbrntt:
Motivation:

When getting the canonical form of headers we check whether the header
name is 'set-cookie' to avoid splitting values which are not safe to
split.

This check is done against a `lowercased()` version of the header name.
However, we don't need to produce a new `String` here as we already have
a case insensitive string comparision function.

Modifications:

- Add a test to verify that getting the canonical form of "set-cookie"
  does not split any header values
- Switch to `isEqualCaseInsensitiveASCIIBytes`

Result:

More tests, cheaper comparision.

### DIFF
--- a/Sources/NIOHPACK/HPACKHeader.swift
+++ b/Sources/NIOHPACK/HPACKHeader.swift
@@ -256,7 +256,7 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral {
         }
 
         // It's not safe to split Set-Cookie on comma.
-        guard name.lowercased() != "set-cookie" else {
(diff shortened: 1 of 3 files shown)


Lukasa (approved):

glbrntt:

glbrntt:
We regressed about 5-10% since #304 -- I suspect that was actually in #307

Lukasa:
Yup, but we're still well ahead of where we were.

glbrntt:
> Yup, but we're still well ahead of where we were.

Agreed, this is not significantly different to #307 (only 1-2%).

PR #311, opened 2021-11-24
Remove forbidden HTTP1 headers even if no connection header is present

fabianfett:
### Motivation

Currently we only remove forbidden HTTP/1.1 headers when translating to HTTP/2.0, if a connection header is present. Since a connection header is not required on HTTP/1.1, forward forbidden headers are currently sometimes forwarded (like "transfer-encoding").

### Changes

- Remove forbidden HTTP/1.1 headers, even if no connection header is present.

### Result

We don't leak forbidden headers.


Lukasa (approved):

Lukasa:
There's an allocation regression on nightly, but that doesn't have anything to do with this patch. Merging over the failure.

PR #312, opened 2021-11-24
Optimise reading of contiguous integers

lmcd:
I noticed this optimisation merged into `swift-nio` a few days ago: https://github.com/apple/swift-nio/pull/1987 by @weissi 
This PR applies this new read method to a few instances in `HTTP2FrameParser` when contiguous integers were being read from a `ByteBuffer`

### DIFF
--- a/Sources/NIOHTTP2/HTTP2FrameParser.swift
+++ b/Sources/NIOHTTP2/HTTP2FrameParser.swift
@@ -686,11 +686,10 @@ struct HTTP2FrameDecoder {
 
         let priorityData: HTTP2Frame.StreamPriorityData?
         if flags.contains(.priority) {
-            let values = bytes.readMultipleIntegers(as: (UInt32, UInt8).self)!
-            let raw = values.0
+            let (raw, weight) = bytes.readMultipleIntegers(as: (UInt32, UInt8).self)!
             priorityData = HTTP2Frame.StreamPriorityData(exclusive: (raw & 0x8000_0000 != 0),
                                                          dependency: HTTP2StreamID(networkID: raw),
-                                                         weight: values.1)
+                                                         weight: weight)
             bytesToRead -= 5
         } else {
             priorityData = nil
@@ -723,11 +722,10 @@ struct HTTP2FrameDecoder {
             throw InternalError.codecError(code: .frameSizeError)
         }
 
-        let values = bytes.readMultipleIntegers(as: (UInt32, UInt8).self)!
-        let raw: UInt32 = values.0
+        let (raw, weight) = bytes.readMultipleIntegers(as: (UInt32, UInt8).self)!
         let priorityData = HTTP2Frame.StreamPriorityData(exclusive: raw & 0x8000_0000 != 0,
                                                          dependency: HTTP2StreamID(networkID: raw),
-                                                         weight: values.1)
+                                                         weight: weight)
         return .priority(priorityData)
     }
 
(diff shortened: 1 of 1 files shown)


Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -686,10 +686,11 @@ struct HTTP2FrameDecoder {
 
         let priorityData: HTTP2Frame.StreamPriorityData?
         if flags.contains(.priority) {
-            let raw: UInt32 = bytes.readInteger()!
+            let values = bytes.readMultipleIntegers(as: (UInt32, UInt8).self)!
```
Lukasa, line 689: This can be cleaned up as:

```suggestion
            let (raw, weight) = bytes.readMultipleIntegers(as: (UInt32, UInt8).self)!
```

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -722,10 +723,11 @@ struct HTTP2FrameDecoder {
             throw InternalError.codecError(code: .frameSizeError)
         }
 
-        let raw: UInt32 = bytes.readInteger()!
+        let values = bytes.readMultipleIntegers(as: (UInt32, UInt8).self)!
```
Lukasa, line 726: Same here.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -779,8 +781,9 @@ struct HTTP2FrameDecoder {
         var consumed = 0
         while consumed < length {
             // TODO: name here should be HTTP2SettingsParameter(fromNetwork:), but that's currently defined for NGHTTP2's Int32 value
-            let identifier = HTTP2SettingsParameter(fromPayload: bytes.readInteger()!)
-            let value: UInt32 = bytes.readInteger()!
+            let values = bytes.readMultipleIntegers(as: (UInt16, UInt32).self)!
```
Lukasa, line 784: This can be cleaned up as:

```suggestion
            let (rawIdentifier, value) = bytes.readMultipleIntegers(as: (UInt16, UInt32).self)!
```

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -855,8 +858,9 @@ struct HTTP2FrameDecoder {
             throw InternalError.codecError(code: .frameSizeError)
         }
 
-        let raw: UInt32 = bytes.readInteger()!
-        let errcode: UInt32 = bytes.readInteger()!
+        let values = bytes.readMultipleIntegers(as: (UInt32, UInt32).self)!
```
Lukasa, line 861: Same cleanup can be done here.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -1197,15 +1201,18 @@ fileprivate struct FrameHeader {
...
         let saveSelf = self
-        guard let lenHigh = self.readInteger(as: UInt16.self),
-            let lenLow = self.readInteger(as: UInt8.self),
-            let type = self.readInteger(as: UInt8.self),
-            let flags = self.readInteger(as: UInt8.self),
-            let rawStreamID = self.readInteger(as: UInt32.self) else {
-                self = saveSelf
-                return nil
+
+        guard let values = self.readMultipleIntegers(as: (UInt16, UInt8, UInt8, UInt8, UInt32).self) else {
```
Lukasa, line 1205: And the same cleanup can be done with this conditional binding too.

Lukasa (changes requested):
Thanks for this, it generally looks great! I've left a few notes in the diff. Sadly, we can't merge this until we tag a new NIO release, and then we'll need to update this patch to require that NIO version.

lmcd:
@Lukasa - I've made those changes. I've left the first one (inside the `guard`) as it's debatable which approach is better for clarity, but I can change that too if you like.

Also added another commit that does the same for writes.

Lukasa (changes requested):
Great, if you can update Package.swift to lift the lower NIO version bound to the newly-released 2.35.0 we should be able to merge this.

weissi:
friendly ping @lmcd

PR #313, opened 2021-11-24
Reduce allocations..?

lmcd:
I've been profiling large HTTP upload requests in swift-nio, and have noticed a lot of copying/allocations going on in `HTTPFrameDecoder`.

Is it possible in this instance to prevent the copy by just switching the accumulating byte buffer to the incoming one if the accumulate bytes are still zero? I don't know NIO well enough yet to understand what side effects this may have.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2FrameParser.swift
+++ b/Sources/NIOHTTP2/HTTP2FrameParser.swift
@@ -64,7 +64,12 @@ struct HTTP2FrameDecoder {
         }
 
         mutating func accumulate(bytes: inout ByteBuffer) {
-            _ = self.unusedBytes.writeBuffer(&bytes)
+            if self.unusedBytes.readableBytes == 0 {
+                self.unusedBytes = bytes
+            }
+            else {
+                _ = self.unusedBytes.writeBuffer(&bytes)
+            }
         }
     }
 


Lukasa:
In practical cases this _may_ reduce allocations. We'll get no more than one actual read per read cycle if the NIOSSLHandler is in front of us in the pipeline, in which case this can reduce the allocations. Did you profile your allocations? If so, can you attach an Instruments trace? We can likely track down exactly where those allocations are coming from and work out if they're necessary.

Lukasa:
Note also that we accumulate bytes in more than one place, so you'd want to tweak almost all of these.

lmcd:
I don't have the machine in front of me right now, but from memory this was consuming about 2-3% of CPU, and after this update it wasn't showing up in the instruments trace at all.

I addressed that specific struct as that was the only one that was registering some CPU pressure. This might not be an issue under normal use, I'm just specifically profiling long-running file uploads locally (e.g. uploading a 1GB file)

Lukasa:
I'd really like to see an instruments trace if we can. Out of curiosity, were you running in release mode or debug mode?

Lukasa:
Incidentally, one reason I'm asking for this is that this change _could_ just move the work from this function to a different one. NIO attempts to re-use read buffers, and depending on the I/O pattern this can prevent that from happening.

PR #314, opened 2021-11-25
Swift nightly has improved alloc counts.

Lukasa:
This patch updates the alloc counter to enforce the new numbers.


Lukasa:
This is not as good as it looks: this is merely fixing a regression from 5.5.

glbrntt (approved):

dnadoba (approved):

PR #315, opened 2021-11-26
Deliver outbound stream errors to channels.

Lukasa:
Motivation:

When we hit a stream error on an outbound frame, we don't currently
distinguish between them and channel errors. This makes it impossible
for anyone to tell them apart. It also makes it impossible for the
multiplexer to deliver these errors into the relevant stream channels
that triggered them, making it essentially certain they'll be lost.

Better diagnostics would be good, so we are providing them.

Modifications:

- Add new StreamError type that carries a stream ID and an underlying
  error.
- Emit this StreamError type for outbound stream errors.
- Deliver this StreamError type into the relevant child channels when
  possible.

Result:

Better diagnostics for stream errors.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
@@ -205,6 +205,15 @@ public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboun
         context.fireChannelWritabilityChanged()
     }
 
+    public func errorCaught(context: ChannelHandlerContext, error: Error) {
+        if let streamError = error as? NIOHTTP2Errors.StreamError,
+           let channel = self.streams[streamError.streamID] {
+            channel.receiveStreamError(streamError)
+        }
+
+        context.fireErrorCaught(error)
+    }
+
     private func newConnectionWindowSize(newSize: Int, context: ChannelHandlerContext) {
         guard let increment = self.connectionFlowControlManager.newWindowSize(newSize) else {
             return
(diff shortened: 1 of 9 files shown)


Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
```diff
@@ -205,6 +205,15 @@ public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboun
         context.fireChannelWritabilityChanged()
     }
 
+    public func errorCaught(context: ChannelHandlerContext, error: Error) {
+        if let streamError = error as? NIOHTTP2Errors.StreamError,
+           let channel = self.streams[streamError.streamID] {
+            channel.receiveStreamError(streamError)
+        }
```
glbrntt, line 212: Is it possible to receive a stream error for a stream we don't know about? Wondering whether we should drop stream errors for streams we aren't aware of (to avoid downstream handlers, for example, closing the channel on receiving an error).

Lukasa: Yes, it is, but I don't think we should drop them. Just because the multiplexer doesn't know about them doesn't mean the error is not of interest.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
```diff
@@ -205,6 +205,15 @@ public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboun
...
+
+        context.fireErrorCaught(error)
```
dnadoba, line 214: Do we really want to unconditionaly forward it to the next channel handler even if we deliver it to one of our child channels?

dnadoba (approved):
LGTM, just one question.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
Lukasa: Yeah, I think we probably do: this ensures that errors don't suddenly go missing when they were reported before.

glbrntt (approved):

fabianfett:
Even though, to late to the game... 👍 Great change, thanks!

PR #316, opened 2021-11-29
Add Transfer-Encoding chunked to HTTP1 requests and responses.

fabianfett:
### Motivation

When translating HTTP2 to HTTP1, developers assume HTTP1 header semantics. HTTP2 does not know the Transfer-Encoding header. For this reason, we should add the Transfer-Encoding header, when needed to HTTP2 requests and responses.

### Changes

- Adds the Transfer-Encoding to incoming HTTP requests on the server http2 to http1 translation side
- Adds the Transfer-Encoding to incoming HTTP responses on the client http2 to http1 translation side

### Result

Users get more real http1.1 semantics, when using the HTTP2 to HTTP1 translation.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2ToHTTP1Codec.swift
+++ b/Sources/NIOHTTP2/HTTP2ToHTTP1Codec.swift
@@ -23,6 +23,8 @@ fileprivate struct BaseClientCodec {
     private let normalizeHTTPHeaders: Bool
 
     private var headerStateMachine: HTTP2HeadersStateMachine = HTTP2HeadersStateMachine(mode: .client)
+    
+    private var outgoingHTTP1RequestHead: HTTPRequestHead?
 
     /// Initializes a `BaseClientCodec`.
     ///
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP2/HTTP2ToHTTP1Codec.swift
```diff
@@ -492,91 +513,11 @@ private extension HTTPMethod {
     }
 }
 
-
-internal extension String {
-    /// Create a `HTTPMethod` from the string representation of that method.
-    init(httpMethod: HTTPMethod) {
-        switch httpMethod {
```
fabianfett, line 499: This is implemented in NIOHTTP1 and can be removed for this reason.

Sources/NIOHTTP2/HTTP2ToHTTP1Codec.swift
```diff
@@ -82,7 +100,10 @@ fileprivate struct BaseClientCodec {
     mutating func processOutboundData(_ data: HTTPClientRequestPart, allocator: ByteBufferAllocator) throws -> HTTP2Frame.FramePayload {
         switch data {
         case .head(let head):
+            #warning("How we want to deal with multiple requests on the same stream?")
+            precondition(self.outgoingHTTP1RequestHead == nil)
```
fabianfett, line 104: @Lukasa, what do we want to do in this case?

Sources/NIOHTTP2/HTTP2ToHTTP1Codec.swift
```diff
@@ -613,6 +580,35 @@ internal extension HTTPResponseHead {
         let status = HTTPResponseStatus(statusCode: integerStatus)
         self.init(version: .init(major: 2, minor: 0), status: status, headers: HTTPHeaders(regularHeadersFrom: headers))
     }
+    
+    /// Create a `HTTPResponseHead` from the header block. Use this for non-informational responses
```
Lukasa, line 584:
```suggestion
    /// Create a `HTTPResponseHead` from the header block. Use this for non-informational responses.
```

Sources/NIOHTTP2/HTTP2ToHTTP1Codec.swift
```diff
@@ -613,6 +580,35 @@ internal extension HTTPResponseHead {
...
+    init(http2HeaderBlock headers: HPACKHeaders, requestHead: HTTPRequestHead, endStream: Bool) throws {
+        // A response head should have only one psuedo-header. We strip it off.
+        let statusHeader = try headers.peekPseudoHeader(name: ":status")
+        guard let integerStatus = Int(statusHeader, radix: 10) else {
+            throw NIOHTTP2Errors.invalidStatusValue(statusHeader)
+        }
+        let status = HTTPResponseStatus(statusCode: integerStatus)
```
Lukasa, line 591: Given that the first few lines here are identical to the method above, should we not just call that and assign it to `self`, and then modify `self` as needed by adding the additional headers?

dnadoba (approved):

Lukasa (approved):

PR #319, opened 2021-12-13
Protect against reentrancy inside the HTT2ChannelHandler

FranzBusch:
In rare cases, the `HTTP2ChannelHandler`s method `unbufferAndFlushAutomaticFrames` and the `HTTP2StreaMultiplexer`s `newConnectionWindowSize` method were recursively calling themselves. This was causing a stack overflow. To protect ourselves from this stack overflow, I added a variable to guard against reentrancy inside the `unbufferAndFlushAutomaticFrames` method.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
@@ -81,6 +81,10 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
     /// to determine buffering strategies.
     private var channelWritable: Bool = true
 
+    /// This variable is set to `true` when we are inside `unbufferAndFlushAutomaticFrames` and protects us from reentering the method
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -81,6 +81,10 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
     /// to determine buffering strategies.
     private var channelWritable: Bool = true
 
+    /// This variable is set to `true` when we are inside `unbufferAndFlushAutomaticFrames` and procetcs us from reentering the method
```
Lukasa, line 84:
```suggestion
    /// This variable is set to `true` when we are inside `unbufferAndFlushAutomaticFrames` and protects us from reentering the method
```

Lukasa (approved):
One minor typo otherwise this looks great.

PR #328, opened 2022-02-23
Use unchecked math in encodeInteger

Lukasa:
Motivation

Checked math is great when you aren't sure what the constraints are, but
in some places we happen to know that all the arithmetic operations
occur in a defined domain and range. In those cases, we can improve the
codegen by adopting unchecked math.

Modifications

Widely adopt unchecked math in encodeInteger.

Result

Better codegen for encodeInteger.


glbrntt (approved):

fabianfett (approved):

Lukasa:

PR #330, opened 2022-03-17
fix code soundness check

tomerd:
motivation: code should be sound!

changes:
* update one case of inappropriate langauge
* update header years check regex

PR #331, opened 2022-03-17
Update alloc limits and clean up soundness

Lukasa:
Some soundness checks were broken by our security fix (thanks to @tomerd in #330) and the allocation counts got dropped by the rewrite. This fixes both.


FranzBusch (approved):
LGTM

PR #339, opened 2022-04-25
Use const array for unrolled Huffman table

Lukasa:
Motivation

We've spent a long time with an unrolled Huffman-table to optimize the
Huffman decoding of HPACK headers. This works well, but because of some
limitations in Swift it's been impossible to embed this structure in the
binary directly: we've always had to alloc-and-copy on startup.
Additionally, due to type checker performance limitations, we've been
Base64 decoding it on startup, which isn't great either.

The introduction of const allows us to embed the knowledge of the fact
that this structure is compile-time constant in to the binary, improving
performance.

Modifications

Revert to the old layout, constify, and remove anything that blocks
that.

Result

Hopefully better performance and less dirty memory.

This is a bit of a WIP. In particular, we have a few issues to resolve. The biggest one is that this still _drastically_ slows down compiles due to the type checker overhead. The type checking here takes 25s on my machine to check this one statement, which is simply not practical for something shipped as source code.


Lukasa:

Lukasa:
Huh, no comparison there. Weird. I don't expect substantial performance improvements regardless: this only cleans up the dirty memory situation and our benchmark doesn't really test memory pressure.

PR #341, opened 2022-05-04
Remove build_podspec.sh

glbrntt:
Motivation:

We no longer support Cocoapods.

Modifications:

- Remove build_podspec.sh and associated scripts

Result:

Less unused code.


Lukasa (approved):
Love a negative diff.

PR #343, opened 2022-05-04
Adopt `Sendable` in relevant types

Davidde94:
Make `InboundEventBuffer` and `StringAndLocationStorage` conform to `@unchecked Sendable` in preparation for async/await in NIO. These are the only 2 types I could find where it's sensible to provide `Sendable` conformance. Everything else is either a channel handler or a struct.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2Error.swift
+++ b/Sources/NIOHTTP2/HTTP2Error.swift
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 import NIOHPACK
+import NIOConcurrencyHelpers
 
 public protocol NIOHTTP2Error: Equatable, Error { }
 
@@ -1454,26 +1455,80 @@ private func _location(file: String, line: UInt) -> String {
 }
 
 private final class StringAndLocationStorage: Equatable {
-    var value: String
-    var file: String
-    var line: UInt
+
+    private let lock = Lock()
+
+    private var _value: String
+    private var _file: String
+    private var _line: UInt
+
+    var value: String {
+        get {
+            self.lock.withLock {
+                self._value
+            }
+        }
+        set {
+            self.lock.withLock {
+                self._value = newValue
+            }
+        }
+    }
+    var file: String {
+        get {
+            self.lock.withLock {
+                self._file
+            }
+        }
+        set {
+            self.lock.withLock {
+                self._file = newValue
+            }
+        }
+    }
+    var line: UInt {
+        get {
+            self.lock.withLock {
+                self._line
+            }
+        }
+        set {
+            self.lock.withLock {
+                self._line = newValue
+            }
+        }
+    }
 
     var location: String {
-        return _location(file: self.file, line: self.line)
+        self.lock.withLock {
+            _location(file: self._file, line: self._line)
+        }
     }
 
     init(_ value: String, file: String, line: UInt) {
-        self.value = value
-        self.file = file
-        self.line = line
+        self._value = value
+        self._file = file
+        self._line = line
     }
 
     func copy() -> StringAndLocationStorage {
-        return StringAndLocationStorage(self.value, file: self.file, line: self.line)
+        self.lock.withLock {
+            return StringAndLocationStorage(self._value, file: self._file, line: self._line)
+        }
     }
 
     static func ==(lhs: StringAndLocationStorage, rhs: StringAndLocationStorage) -> Bool {
         // Only compare the value. The 'file' is not relevant here.
-        return lhs.value == rhs.value
+        lhs.lock.withLock {
+            rhs.lock.withLock {
+                lhs._value == rhs._value
+            }
+        }
     }
 }
+
+#if swift(>=5.5) && canImport(_Concurrency)
+extension StringAndLocationStorage: @unchecked Sendable {
+
+}
+#endif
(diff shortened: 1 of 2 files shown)


Lukasa (changes requested):
Structs need to be made sendable where they are thread-safe: `Sendable` is not assumed for public structures.

Davidde94:
@Lukasa thanks, moved to WIP

Davidde94:

Sources/NIOHPACK/IndexedHeaderTable.swift
```diff
@@ -190,3 +190,9 @@ public struct IndexedHeaderTable {
         set { self.dynamicTable.maximumTableLength = newValue }
     }
 }
+
+#if swift(>=5.5) && canImport(_Concurrency)
+extension IndexedHeaderTable: @unchecked Sendable {
```
Lukasa, line 195: Can we use `@preconcurrency` to cover this?

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -293,3 +293,15 @@ extension HTTP2Frame.FramePayload {
         }
     }
 }
+
+#if swift(>=5.5) && canImport(_Concurrency)
+extension HTTP2Frame: @unchecked Sendable {
+
+}
+extension HTTP2Frame.FramePayload: @unchecked Sendable {
```
Lukasa, line 301: I think `@preconcurrency` should handle this

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -293,3 +293,15 @@ extension HTTP2Frame.FramePayload {
         }
     }
 }
+
+#if swift(>=5.5) && canImport(_Concurrency)
+extension HTTP2Frame: @unchecked Sendable {
```
Lukasa, line 298: I don't think this should need `@unchecked`.

Sources/NIOHTTP2/InboundEventBuffer.swift
```diff
@@ -65,3 +70,9 @@ extension InboundEventBuffer: CustomStringConvertible {
         return "InboundEventBuffer { \(self.buffer) }"
     }
 }
+
+#if swift(>=5.5) && canImport(_Concurrency)
+extension InboundEventBuffer: @unchecked Sendable {
```
Lukasa, line 75: This type is very deliberately not `Sendable` and shouldn't be. Can we revert the change here?

Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -344,29 +347,72 @@ public enum NIOHTTP2Errors {
             }
         }
 
-        private final class Storage: Equatable {
-            var delta: Int
-            var currentWindowSize: Int
-            var file: String
-            var line: UInt
+        fileprivate final class Storage: Equatable {
```
Lukasa, line 350: We don't need this: the backing storage is only accessed after a call to `self.copyStorageIfNotUniquelyReferenced`, so we will never write to this across threads. This is the perfect case for making the backing storage _not_ `Sendable`, and instead having the wrapper `struct` be `@unchecked Sendable`.

Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -1201,29 +1247,72 @@ public enum NIOHTTP2Errors {
             }
         }
 
-        private final class Storage: Equatable {
-            var name: String
-            var value: String
-            var file: String
-            var line: UInt
+        fileprivate final class Storage: Equatable {
```
Lukasa, line 1250: Same note here.

Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -1375,21 +1464,51 @@ public enum NIOHTTP2Errors {
     /// As they are a wrapper error, they carry a "real" error in the `baseError`. Additionally, they cannot
     /// meaningfully be `Equatable`, so they aren't. There's also no additional location information: that's
     /// provided by the base error.
-    public struct StreamError: Error {
-        private final class Storage {
-            var streamID: HTTP2StreamID
-            var baseError: Error
+    public struct StreamError: Error, NIOSendable {
+        fileprivate final class Storage {
```
Lukasa, line 1468: Same note here

Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -1454,26 +1573,89 @@ private func _location(file: String, line: UInt) -> String {
 }
 
 private final class StringAndLocationStorage: Equatable {
-    var value: String
-    var file: String
-    var line: UInt
+
```
Lukasa, line 1576: Same note here.

Davidde94:
Discussed with Cory offline - there's no way to use `@preconcurrency` to avoid `@unchecked` at the moment, so I'll resolve those comments.

Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -1474,6 +1476,6 @@ private final class StringAndLocationStorage: Equatable {
 
     static func ==(lhs: StringAndLocationStorage, rhs: StringAndLocationStorage) -> Bool {
         // Only compare the value. The 'file' is not relevant here.
-        return lhs.value == rhs.value
+        lhs.value == rhs.value
     }
 }
```
Lukasa, line 1481: Do you mind reverting the changes in this file?

Lukasa (commented):
Shiny, one minor note and we're good.

Issue #342, opened 2022-05-04
Reconsider implementation of Equatable and Hashable for HPACKHeaders


Lukasa:
For historical reasons, HPACKHeaders does not pay attention to order or indexability when considering equality and hashability. We should consider whether this is appropriate for this type, or whether we should make the definition more strict.

dnadoba:
We could also deprecate the conformance and add two views to it, `ordered` and `unordered`, and only conform the views to Equatable/Hashable.

Lukasa:
I don't think that's really the right thing to do. We should commit the core type to one of those two positions, and where necessary we can always expose the other.

PR #344, opened 2022-05-04
Make HPACKHeaders hashable.

Lukasa:
Motivation:

HPACKHeaders is equatable, and it rarely makes sense to have a type that
is equatable but not hashable. The definition of equatability is pretty
limited here (see #342) but we can define a definition of hashability
consistent with it.

Modifications:

- Make HPACKHeaders hashable.

Result:

People can store HPACKHeaders in Sets


Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -394,6 +394,9 @@ extension HPACKHeaders: CustomStringConvertible {
     }
 }
 
+// NOTE: This is a bad definition of equatable and hashable. In particular, both order and
+// indexability are ignored. We're sort of stuck with this behaviour now and cannot change it,
+// but in the next major version we should strongly consider revising it.
```
Davidde94, line 399: Should we link a GH issue?

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -416,6 +419,22 @@ extension HPACKHeaders: Equatable {
...
+extension HPACKHeaders: Hashable {
+    @inlinable
+    public func hash(into hasher: inout Hasher) {
+        // Discriminator, to indicate that this is a collection. This improves the performance
+        // of Sets and Dictionaries that include collections of HPACKHeaders by reducing hash collisions.
+        hasher.combine(self.count)
+
+        // This emulates the logic used in equatability, but we sort it to ensure that
+        // we hash equivalently.
+        let names = Set(self.headers.map { $0.name }).sorted()
```
Davidde94, line 431: Why do we need to load into a `Set` to then immediately extract?

Davidde94 (approved):
Minor question, otherwise looks good

Sources/NIOHPACK/HPACKHeader.swift
Lukasa: Flattening uniqueness: it's possible to have multiple entries with the same name. This is what happens above in the equatable implementation.

Sources/NIOHPACK/HPACKHeader.swift
Lukasa: Oh, I think this comment is just unnecessarily aggressive.

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -416,6 +419,22 @@ extension HPACKHeaders: Equatable {
...
+        for name in names {
+            hasher.combine(self[name].sorted())
```
dnadoba, line 433: Does the order of calls to `Hasher.combine(_:)` matter?

Sources/NIOHPACK/HPACKHeader.swift
dnadoba: We can remove one allocation if we lazily map over the headers to extract the name
```suggestion
        let names = Set(self.headers.lazy.map { $0.name }).sorted()
```

dnadoba (commented):
One perf and one question, otherwise LGTM

PR #347, opened 2022-05-17
Improve HTTPHeader normalisation performance for large inputs

glbrntt:
Motivation:

Header normalisation is expensive when the input headers contain a large
number of connection header values.

Modifications:

- When the headers to normalise contain more than 32 connection header
  values use a set to provide constant time lookup, rather than linear.

Result:

Better perf for HTTPHeader normalisation with large connection header
value lists.


Lukasa (approved):

Lukasa:
Really nice change!

PR #351, opened 2022-07-29
Use Docc for documentation

Lukasa:
Motivation

Documentation is nice, and we can help support users by providing useful
clear docs.

Modifications

Add Docc to 5.6 and later builds
Make sure symbol references work
Add overview docs

Result

Nice rendering docs

### DIFF
--- a/docker/docker-compose.2004.56.yaml
+++ b/docker/docker-compose.2004.56.yaml
@@ -12,6 +12,7 @@ services:
 
   unit-tests:
     image: swift-nio-http2:20.04-5.6
+    command: /bin/bash -xcl "swift test -Xswiftc -warnings-as-errors --enable-test-discovery"
 
   integration-tests:
     image: swift-nio-http2:20.04-5.6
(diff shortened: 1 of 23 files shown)


FranzBusch (approved):
LGTM!

docker/docker-compose.2004.56.yaml
```diff
@@ -12,6 +12,7 @@ services:
 
   unit-tests:
     image: swift-nio-http2:20.04-5.6
+    command: /bin/bash -xcl "swift test -Xswiftc -warnings-as-errors --enable-test-discovery"
```
PeterAdams-A, line 15: Rather than doing this override for all the version 5.5+ why don't you change the default and override for 5.4?

Lukasa: Giving this a shot now.

Lukasa: We need test discovery disabled on 5.5 too because we have deprecated tests here. Otherwise this works fine.

PR #354, opened 2022-08-26
Better GoAWAY error

carolinacass:
Motivation:

Improve error messages

Modifications:

Replaced no such stream error with correct createdStreamAfterGoaway error

### DIFF
--- a/Sources/NIOHTTP2/ConnectionStateMachine/FrameSendingStates/SendingHeadersState.swift
+++ b/Sources/NIOHTTP2/ConnectionStateMachine/FrameSendingStates/SendingHeadersState.swift
@@ -70,8 +70,9 @@ extension SendingHeadersState where Self: RemotelyQuiescingState {
     mutating func sendHeaders(streamID: HTTP2StreamID, headers: HPACKHeaders, isEndStreamSet endStream: Bool) -> StateMachineResultWithEffect {
         let validateHeaderBlock = self.headerBlockValidation == .enabled
         let validateContentLength = self.contentLengthValidation == .enabled
-        if streamID > self.streamState.lastClientStreamID &&
-            self.role == .client && streamID.mayBeInitiatedBy(.client) {
+        if streamID.mayBeInitiatedBy(.client) &&
+            self.role == .client &&
+            streamID > self.streamState.lastClientStreamID {
             let error = NIOHTTP2Errors.createdStreamAfterGoaway()
             return StateMachineResultWithEffect(result: .connectionError(underlyingError: error, type: .protocolError), effect: nil)
         }
--- a/Sources/NIOHTTP2/ConnectionStateMachine/FrameSendingStates/SendingHeadersState.swift
+++ b/Sources/NIOHTTP2/ConnectionStateMachine/FrameSendingStates/SendingHeadersState.swift
@@ -70,7 +70,11 @@ extension SendingHeadersState where Self: RemotelyQuiescingState {
     mutating func sendHeaders(streamID: HTTP2StreamID, headers: HPACKHeaders, isEndStreamSet endStream: Bool) -> StateMachineResultWithEffect {
         let validateHeaderBlock = self.headerBlockValidation == .enabled
         let validateContentLength = self.contentLengthValidation == .enabled
-        
+        if streamID > self.streamState.lastClientStreamID &&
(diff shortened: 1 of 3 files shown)


glbrntt:
@swift-nio-bot add to allowlist

Sources/NIOHTTP2/ConnectionStateMachine/FrameSendingStates/SendingHeadersState.swift
```diff
@@ -70,7 +70,11 @@ extension SendingHeadersState where Self: RemotelyQuiescingState {
     mutating func sendHeaders(streamID: HTTP2StreamID, headers: HPACKHeaders, isEndStreamSet endStream: Bool) -> StateMachineResultWithEffect {
         let validateHeaderBlock = self.headerBlockValidation == .enabled
         let validateContentLength = self.contentLengthValidation == .enabled
-        
+        if streamID > self.streamState.lastClientStreamID &&
+            self.role == .client && streamID.mayBeInitiatedBy(.client) {
```
dnadoba, line 74: AFAIK a client can send a GOAWAY frame too in which case the server is not allowed to create new streams. We therefore need to check if we have create the stream ID or the other peer and only disallow stream creation of the stream if we try to create it. This should do the trick:
```suggestion
            streamID.mayBeInitiatedBy(self.role) {
```

dnadoba: Thinking a bit more about this, I think if we are in the `RemotelyQuiescingState` the remote peer is also not allowed to create new streams. HTTP/2 messages are guaranteed to arrive in order so if the remote initiated the shutdown it should also not send any new streams afterwards. We therefore might want to remove the check completely.

glbrntt:

Sources/NIOHTTP2/ConnectionStateMachine/FrameSendingStates/SendingHeadersState.swift
Lukasa: @dnadoba this analysis is not correct.

We transition into `RemotelyQuiescingState` because we have received a GOAWAY frame. Per [RFC 9113 § 6.8](https://httpwg.org/http2-spec/rfc9113.html#GOAWAY) (emphasis added):

> Once the GOAWAY is sent, the sender will ignore frames sent on streams initiated by the receiver if the stream has an identifier higher than the included last stream identifier. **Receivers** of a GOAWAY frame MUST NOT open additional streams on the connection, although a new connection can be established for new streams.

That is, an endpoint can send a GOAWAY frame and continue to create new requests. This is moot however: in this code we are _sending_ a HEADERS frame, not receiving one, so the peer cannot be creating a stream in this code.

What we're trying to detect is whether _we_ are trying to create a stream here. That's what's forbidden by RFC 9113.

More broadly, the original check is correct. Merely seeing `streamID.mayBeInitiatedBy(self.role) == true` is not a sufficient condition for us to create a stream, for two reasons. Firstly, if we are a server, we could attempt to create a stream using a HEADERS frame on a server-initiated stream ID, which would cause your check to pass. This is an error: servers may not create streams by sending HEADERS frames.

To that end we need to check a) that we are a client, b) that we're sending this HEADERS frame on a client-initiated stream, and c) that this stream has not yet been opened by us. That's the complete set of criteria for an error.

Lukasa: @carolinacass Can I ask us to reorder the check? It'd be better to check that we're a client _first_, that the stream ID may be initiated by a client _second_, and only then that it's larger than the last client initiated stream ID. 

There's no correctness difference between this and what you wrote, but IMO it reads more logically (because the first two questions establish that the third question is meaningful to ask).

Lukasa (changes requested):
Nice looking patch, thanks! I've left a couple of notes in the diff to be addressed when you get a moment.

Sources/NIOHTTP2/ConnectionStateMachine/FrameSendingStates/SendingHeadersState.swift
dnadoba: @Lukasa Thanks for clarifying!

PR #360, opened 2022-11-24
ContentLengthVerifier checks all header vals match

rnro:
### Motivation:

Per fetch (https://fetch.spec.whatwg.org/#content-length-header), repeated values of Content-Length are fine if they’re the same, and an error otherwise.

### Modifications:

Iterate over all supplied content-length headers and check for inconsistency.

Note that as per the linked spec, the check for matching values is done _before_ the string to integer conversion. This means for example we will throw for header values "1234" and "01234" complaining of a mismatch even though they are numerically equivalent.

### Result:

ContentLengthVerifier will throw if inconsistent content-length values are found.


Sources/NIOHTTP2/ContentLengthVerifier.swift
```diff
@@ -48,8 +48,19 @@ extension ContentLengthVerifier {
...
+        guard let first = headerIterator.next() else {
+            return
+        }
+
+        // multiple content-length headers are permitted as long as they agree
+        for header in headerIterator {
+            guard header == first else {
+                throw NIOHTTP2Errors.contentLengthHeadersMismatch()
+            }
+        }
```
dnadoba, line 62: TIL that `for ... in` works with an iterator as well 🤯
However, personally I like to use the collection algorithms instead but thats just a style preference so feel free to ignore:
```suggestion
        let contentLengths = headers[canonicalForm: "content-length"]
        guard let first = a.first else {
            return
        }
        // multiple content-length headers are permitted as long as they agree
        guard contentLengths.dropFirst().allSatisfy({ $0 == first }) else {
            throw NIOHTTP2Errors.contentLengthHeadersMismatch()
        }
```

Tests/NIOHTTP2Tests/ContentLengthVerifierTests.swift
```diff
@@ -0,0 +1,27 @@
+import NIOHPACK
+@testable import NIOHTTP2
+import XCTest
+
+class ContentLengthVerifierTests: XCTestCase {
+    func testDuplicatedLengthHeadersPermitted() throws {
+        var headers = HPACKHeaders([("Host", "apple.com"), ("content-length", "1834"), ("User-Agent", "myCoolClient/1.0")])
+        XCTAssertNoThrow(try ContentLengthVerifier(headers))
```
dnadoba, line 8: Can we test that the length is indeed parsed as `1834` and not nil or something else

Tests/NIOHTTP2Tests/ContentLengthVerifierTests.swift
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    func testDuplicatedConflictingLengthHeadersThrow() throws {
+        var headers = HPACKHeaders([("Host", "apple.com"), ("content-length", "1834"), ("User-Agent", "myCoolClient/1.0")])
+        XCTAssertNoThrow(try ContentLengthVerifier(headers))
+
+        headers.add(contentsOf: [("Content-Length", "4381")])
+        XCTAssertThrowsError(try ContentLengthVerifier(headers))
+    }
+}
```
dnadoba, line 27: Can we add a test with a negative length, invalid length and a length above `Int.max`?

Lukasa:

dnadoba (changes requested):

Tests/NIOHTTP2Tests/ContentLengthVerifierTests.swift
```diff
@@ -0,0 +1,27 @@
...
+        XCTAssertNoThrow(try ContentLengthVerifier(headers))
+
+        headers.add(contentsOf: [("content-length", "1834")])
+        XCTAssertNoThrow(try ContentLengthVerifier(headers))
+
+        headers.add(contentsOf: [("Content-Length", "1834")])
+        XCTAssertNoThrow(try ContentLengthVerifier(headers))
+    }
+
+    func testDuplicatedConflictingLengthHeadersThrow() throws {
```
Lukasa, line 20: Can you add a test for the case you've indicated in your PR comment, where the two numerical values match but the string values are different?

Lukasa (changes requested):
Generally looks good! Let's just add one extra test.

Tests/NIOHTTP2Tests/ContentLengthVerifierTests.swift
rnro: As far as I can tell this isn't easy with the currently exposed public information. I initially had a check which parsed the description but this seemed like a bad idea.

Lukasa: You can consume 1384 bytes (which won't throw), and then consume 1 more (which will).

dnadoba: We can make `expectedContentLength` `internal` instead of `private` and afterwards access it:
```swift
XCTAssertEqual(try ContentLengthVerifier(headers).expectedContentLength, 1834)
```

Sources/NIOHTTP2/ContentLengthVerifier.swift
rnro: I like your version, it's more succinct 🙏

Tests/NIOHTTP2Tests/ContentLengthVerifierTests.swift
rnro: I'd rather add those tests in https://github.com/apple/swift-nio-http2/pull/361 which changes the negative value behaviour.

Lukasa (approved):
This LGTM

dnadoba (approved):
Great! Thanks @rnro

dnadoba:
@rnro soundness fails because you need to run ` ./scripts/generate_linux_tests.rb`

Tests/NIOHTTP2Tests/ContentLengthVerifierTests.swift
```diff
@@ -0,0 +1,58 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
```
dnadoba, line 5: Nit
```suggestion
// Copyright (c) 2022 Apple Inc. and the SwiftNIO project authors
```

PR #361, opened 2022-11-24
Content length verifier negative check

rnro:
`ContentLengthVerifier` `init` throws on negative values

### Motivation:

We discovered that this code path can result in an arithmetic overflow in the case where the content-length header has `INT_MIN` value. Since negative content-lengths don't have any meaning anyway we may as well throw if we encounter one so the decoding can be aborted.

### Modifications:

* ContentLengthVerifier init throws a `ContentLengthHeaderNegative` error when the decoded value is < 0.
* ContentLengthVerifier init throws a `ContentLengthHeaderMalformedValue` error when the value cannot be decoded, e.g. it is out of bounds for Int or is an invalid string.

### Result:

We will no longer panic for `INT_MIN` content-lengths and will fail decoding if one of various invalid values is encountered.

### DIFF
--- a/Sources/NIOHTTP2/ContentLengthVerifier.swift
+++ b/Sources/NIOHTTP2/ContentLengthVerifier.swift
@@ -57,7 +57,15 @@ extension ContentLengthVerifier {
         guard contentLengths.dropFirst().allSatisfy({ $0 == first }) else {
             throw NIOHTTP2Errors.contentLengthHeadersMismatch()
         }
+
         self.expectedContentLength = Int(first, radix: 10)
+
+        guard let expectedLength = self.expectedContentLength else {
+            throw NIOHTTP2Errors.contentLengthHeaderMalformedValue()
+        }
+        if expectedLength < 0 {
+            throw NIOHTTP2Errors.contentLengthHeaderNegative()
+        }
     }
 
     /// The verifier for use when content length verification is disabled.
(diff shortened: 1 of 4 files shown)


Lukasa (approved):

Sources/NIOHTTP2/ContentLengthVerifier.swift
```diff
@@ -57,7 +57,15 @@ extension ContentLengthVerifier {
...
         }
+
         self.expectedContentLength = Int(first, radix: 10)
+
+        guard let expectedLength = self.expectedContentLength else {
+            throw NIOHTTP2Errors.contentLengthHeaderMalformedValue()
+        }
+        if expectedLength < 0 {
+            throw NIOHTTP2Errors.contentLengthHeaderNegative()
+        }
```
dnadoba, line 68: Note: more of a felling then an actual rule I follow

I usually only assign after the value is fully validated. It doesn't really matter but it signals to me at least that once it is assigned I'm done with it.
```suggestion
        guard let expectedLength = Int(first, radix: 10) else {
            throw NIOHTTP2Errors.contentLengthHeaderMalformedValue()
        }
        if expectedLength < 0 {
            throw NIOHTTP2Errors.contentLengthHeaderNegative()
        }
        self.expectedContentLength = expectedLength
```

PR #362, opened 2022-11-25
Provide an iterator for header values

glbrntt:
Motivation:

We don't always require an array of header values to be extracted. Sometimes we just need to be able to iterate through them. For 'regular' header values we incur an allocation for the array, for values in their canonical form we incure at least two allocations. We can do better than this by providing a header value sequence.

Modifications:

- Add new API to `HPACKHeaders` for getting a sequence of header values which can optionally be in their canonical form.
- Update the API of the two existing subscripts to use this new API. For `subscript(canonicalForm:)` we save an allocation by not going via `subscript(name:)`.

Result:

- Fewer allocations in `subscript(canonicalForm:)`
- Header values can be iterated without allocating


Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -276,42 +265,159 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral, Sendable {
...
             return []
         }
 
-        // It's not safe to split Set-Cookie on comma.
-        if name.isEqualCaseInsensitiveASCIIBytes(to: "set-cookie") {
-            return result
+        // Estimate the size of the array at the cost of a second scan through the header table.
+        // This allows us to avoid the incremental allocations we would otherwise have without
+        // reserving capacity in the array.
+        let count = self._countValues(forHeader: name)
```
Lukasa, line 307: This seems somewhat needlessly expensive: is there any reason not to just overcommit the array? We could default this size to `self.count`, for example.

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -276,42 +265,159 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral, Sendable {
...
+        if canonicalForm, name.isEqualCaseInsensitiveASCIIBytes(to: "set-cookie") {
+            // Not safe to split 'set-cookie'.
+            return Values(headers: self, name: name, canonicalize: false)
+        } else {
+            return Values(headers: self, name: name, canonicalize: canonicalForm)
+        }
+    }
+
+    /// Special internal function for use by tests.
+    internal subscript(position: Int) -> (String, String) {
```
Lukasa, line 292: Can we put a leading underscore in the label?

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -276,42 +265,159 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral, Sendable {
...
+                count &+= 1
+            }
+        }
+        return count
+    }
+}
+
+extension HPACKHeaders {
+    public struct Values: Sequence {
+        public typealias Element = String
```
Lukasa, line 333: Given that we're currently trying to reduce allocations, should we take this opportunity to return `Substring` instead of `String`?

Sources/NIOHPACK/HPACKHeader.swift
glbrntt: This isn't new API, GitHub did a bad job with the diff. Happy to make this change in a followup PR though.

Sources/NIOHPACK/HPACKHeader.swift
glbrntt: I wasn't sure which way to lean with the expense: counting values means we don't allocate in the case where there are no values for the given header (this is the behaviour we have now, at least for `subscript(name:)`). Allocating `self.count` also seems excessive. Don't mind which way we go here.

Sources/NIOHPACK/HPACKHeader.swift
glbrntt: Yeah, good idea.

Sources/NIOHPACK/HPACKHeader.swift
Lukasa: Oh interesting, good spot. If we abandon using `append(contentsOf:)` and instead use a `for in` loop then we can lazily reserve the capacity.

glbrntt: We can, but then we get the intermediate (re)allocations which is also a bit sad... 🤷‍♂️

glbrntt: Oh, or do you mean just the first iteration?

Lukasa: Just the first iteration.

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -276,42 +265,159 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral, Sendable {
...
-                let trimmed = entry._trimWhitespace()
-                if trimmed.isEmpty {
+        public mutating func next() -> String? {
+            while true {
+                switch self.process() {
+                case .emit(let value):
+                    return value
+                case .continueProcessing:
                     continue
                 }
```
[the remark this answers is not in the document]
glbrntt: It is a bit of an odd pattern at first sight but I think it makes the code simpler. FWIW: we use this pattern in a few other places too.

Sources/NIOHPACK/HPACKHeader.swift
glbrntt: FWIW: there is no external label at the moment

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -276,42 +265,149 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral, Sendable {
...
             return []
         }
 
-        // It's not safe to split Set-Cookie on comma.
-        if name.isEqualCaseInsensitiveASCIIBytes(to: "set-cookie") {
-            return result
+        var headerValues = [String]()
+        // Avoid intermediate allocations by overcomitting on capacity. Note: may be an undercommit
+        // if the canonical form is being fetched.
+        headerValues.reserveCapacity(self.headers.count)
```
glbrntt, line 311: Previously when getting headers with the `canonicalForm` we overcommitted by 4x the number of headers for the given name. Overcommitting by some multiple of `self.headers.count` feels like excessive resource usage. 

However this does mean we pay for intermediate reallocations if `canonicalForm: true` and the headers are dominated by the name we're looking for (which is why the alloc counting tests show no improvements.) I think this is totally fine though.

Lukasa: Yeah I mean the only way we can fix this is to do the computation pre-pass, but having to scan the headers _twice_ to find the field names is extremely painful. The performance of that scan is really not good.

Lukasa (approved):
Nice, LGTM.

PR #365, opened 2022-12-08
Make FramePayload.Data class-backed

Lukasa:
Motivation:

Using indirect cases makes unwrapping and rewrapping HTTP/2 frames more expensive than it needs to be. As FramePayload.Data is essentially always stuffed into a FramePayload enum, we may as well just class-back it and avoid the extra allocs.

Modifications:

Make FramePayload.Data class-backed

Result:

Fewer allocations in many use-cases

### DIFF
--- a/Sources/NIOHTTP2/HTTP2Frame.swift
+++ b/Sources/NIOHTTP2/HTTP2Frame.swift
@@ -125,8 +125,8 @@ public struct HTTP2Frame: Sendable {
         /// The payload of a `DATA` frame.
         public struct Data {
             /// The application data carried within the `DATA` frame.
+            @inlinable
             public var data: IOData {
-                @inlinable
                 get {
                     return self._backing.data
                 }
@@ -137,8 +137,8 @@ public struct HTTP2Frame: Sendable {
             }
 
             /// The value of the `END_STREAM` flag on this frame.
+            @inlinable
             public var endStream: Bool {
-                @inlinable
                 get {
                     return self._backing.endStream
                 }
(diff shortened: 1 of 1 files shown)


Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -125,33 +125,86 @@ public struct HTTP2Frame: Sendable {
...
+            public var data: IOData {
+                @inlinable
+                get {
+                    return self._backing.data
+                }
+                set {
+                    self._copyIfNeeded()
+                    self._backing.data = newValue
+                }
+            }
```
dnadoba, line 137: why is only the getter `@inlinable`?

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -125,33 +125,86 @@ public struct HTTP2Frame: Sendable {
...
+            init(_ backing: _Backing) {
+                self._backing = backing
+            }
+
+            @inlinable
+            mutating func _copyIfNeeded() {
+                if !isKnownUniquelyReferenced(&self._backing) {
+                    self._backing = self._backing.copy()
+                }
+            }
```
dnadoba, line 184: Does this need to be `@inlinable` if it is not public and never called from an `@inlinable` declaration? If not, we could make it `fileprivate`

dnadoba (approved):
Only small nits and comments out of interest but LGTM.

glbrntt (approved):
Nothing beyond David's comments

PR #372, opened 2023-02-10
Better handle writes issued with already completed promises

glbrntt:
Motivation:

The stream channel updates its writability manager when issuing a write to the parent channel and when the write promise has been succeeded. If the write promise has already been succeeded then the writability manager can be updated in the wrong order which can lead to underflow.

Modifications:

- Use signed arithmetic in the writability manager and allow pending
  bytes to become (briefly) negative.

Result:

The writability manager's pending bytes shouldn't underflow.

### DIFF
--- a/Sources/NIOHTTP2/WatermarkedFlowController.swift
+++ b/Sources/NIOHTTP2/WatermarkedFlowController.swift
@@ -36,23 +36,24 @@
 struct WatermarkedFlowController {
     /// The "high" water mark. If the number of pending bytes exceeds this number, the
     /// writability state will change to "false".
-    private let highWatermark: UInt
+    private let highWatermark: Int
 
     /// The "low" watermark. If the number of pending bytes is lower than this number, the
     /// writability state will change to "true".
-    private let lowWatermark: UInt
+    private let lowWatermark: Int
 
     /// The number of pending bytes waiting to be written to the network.
-    private var pendingBytes: UInt
+    private var pendingBytes: Int
 
     /// Whether the `Channel` should consider itself writable or not.
     internal private(set) var isWritable: Bool
 
     internal init(highWatermark: Int, lowWatermark: Int) {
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTP2/WatermarkedFlowController.swift
```diff
@@ -62,15 +63,18 @@ struct WatermarkedFlowController {
...
             self.isWritable = false
         }
     }
 
     /// Notifies the flow controller that we have successfully written some bytes to the network.
     mutating func wroteBytes(_ writtenBytes: Int) {
-        self.pendingBytes -= UInt(writtenBytes)
+        // pendingBytes can become negative if the ordering of calls to wroteBytes(_:) and bufferedBytes(_:)
+        // is incorrect. This can happen if a completed promise is attached to a write in the HTTP2StreamChannel and
+        // the callback to wroteBytes(_:) happens before bufferedBytes(_:).
```
Lukasa, line 76: Can you elaborate on this a bit to explain why we think it's fine to allow this to happen?

Lukasa: Additionally, can you add a `precondition` that `writtenBytes` is positive?

Sources/NIOHTTP2/WatermarkedFlowController.swift
```diff
@@ -62,15 +63,18 @@ struct WatermarkedFlowController {
 extension WatermarkedFlowController {
     /// Notifies the flow controller that we have buffered some bytes to send to the network.
     mutating func bufferedBytes(_ bufferedBytes: Int) {
-        self.pendingBytes += UInt(bufferedBytes)
+        self.pendingBytes += bufferedBytes
```
Lukasa, line 66: Can you add a precondition that `bufferedBytes` is positive?

Lukasa (approved):

PR #373, opened 2023-02-15
Ensure that we fire channelActive and channelInactive in order

Lukasa:
Motivation:

channelActive must always precede channelInactive. Unfortunately, this isn't guaranteed right now: if we happen to get channelInactive as a result of our writing the preamble we'll fire them in the wrong order. Many systems can tolerate this, but we should avoid doing this anyway.

Modifications:

- Added a state enum to keep track of the state we're in
- Use that state enum to ensure we deliver active/inactive in the right order
- Additionally police other unexpected active/inactive behaviours

Result:

We'll appropriately police the state.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
@@ -92,6 +92,13 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
     /// which can cause an infinite recursion.
     private var isUnbufferingAndFlushingAutomaticFrames = false
 
+    /// In some cases channelInactive can be fired while channelActive is still running.
+    private var activationState = ActivationState.idle
+
+    /// Whether we should tolerate "impossible" state transitions in debug mode. Only true in tests specifically trying to
+    /// trigger them.
+    private let tolerateImpossibleStateTransitionsInDebugMode: Bool
+
     /// The mode for this parser to operate in: client or server.
     public enum ParserMode {
         /// Client mode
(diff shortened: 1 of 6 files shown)


Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -699,8 +807,51 @@ NIOHTTP2Handler(
...
+    }
+
+    fileprivate func impossibleActivationStateTransition(
+        state: ActivationState, activating: Bool, context: ChannelHandlerContext
+    ) {
+        debugOnly {
+            if !self.tolerateImpossibleStateTransitionsInDebugMode {
+                fatalError("Unexpected channelActive in state \(state)")
+            }
+        }
```
glbrntt, line 844: Isn't this just `assert(self.tolerateImpossibleStateTransitionsInDebugMode, "Unexpected channelActive in state \(state)")`?

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -160,6 +197,10 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
         self.writeBuffer = context.channel.allocator.buffer(capacity: 128)
 
         if context.channel.isActive {
+            // handlerAdded is always first, so we can confidently assert our initial state.
+            // We jump immediately to activated here, as channelActive has probably already passed.
+            precondition(self.activationState == .idle)
```
glbrntt, line 202: I understand the reasoning here but where do we draw the line between using `precondition` and `impossibleActivationStateTransition`?

glbrntt (approved):

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
Lukasa: Yes, it is. I had done this more awkward spelling because _originally_ the code would always fatalError, so I had nothing to assert on. When I realised some of these code paths could actually be tested I refactored and didn't change the spelling.

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
Lukasa: Good question. In this instance my view was that it isn't "network-reachable": this can only be hit if the user actively uses the `NIOHTTP2Handler` twice. However, we can still error here and so we probably should.

PR #377, opened 2023-03-03
Add no-op NIOHTTP2ConnectionDemultiplexer

rnro:
Motivation:

This change prepares for future work to embed the connection demultiplexer within the `HTTP2ChannelHandler` which will improve performance by removing the need for passing expensive `InboundUserEvent`s.

Modifications:

This first step defines a new protocol (`NIOHTTP2ConnectionDemultiplexer`) which will later be used to abstract over the embedded and legacy separate demultiplexer code but at the moment should not change the logic.

Result:

Behaviour should be unchanged.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
@@ -99,8 +99,7 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
     /// trigger them.
     private let tolerateImpossibleStateTransitionsInDebugMode: Bool
 
-    /// The delegate for demultiplexing streams. Set on `handlerAdded` and `nil`-ed out on
-    /// `channelInactive` and `handlerRemoved`.
+    /// The delegate for demultiplexing streams. Set on `handlerAdded` and `nil`-ed out on `handlerRemoved`.
     private var demultiplexer: ConnectionDemultiplexer?
 
     /// The mode for this parser to operate in: client or server.
@@ -283,7 +282,6 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
             // This is the easy one. We were active, now we aren't.
             self.activationState = .inactive
             self.channelClosed = true
-            self.demultiplexer = nil
             context.fireChannelInactive()
         case .activating:
             // Huh, we got channelInactive during activation. We need to maintain
@@ -498,16 +496,7 @@ extension NIOHTTP2Handler {
     /// Emit any pending user events.
     private func processPendingUserEvents(context: ChannelHandlerContext) {
         for event in self.inboundEventBuffer {
-            switch event {
-            case .streamCreated(let event):
-                self.demultiplexer?.streamCreated(event.streamID, localInitialWindowSize: event.localInitialWindowSize, remoteInitialWindowSize: event.remoteInitialWindowSize)
-            case .initialStreamWindowChanged(let event):
-                self.demultiplexer?.initialStreamWindowChanged(by: event.delta)
-            case .streamWindowUpdated(let event):
-                self.demultiplexer?.streamWindowUpdated(event.streamID, inboundWindowSize: event.inboundWindowSize, outboundWindowSize: event.outboundWindowSize)
-            case .streamClosed(let event):
-                self.demultiplexer?.streamClosed(event.streamID, reason: event.reason)
-            }
+            self.demultiplexer?.process(event: event)
         }
     }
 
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -272,6 +278,7 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
             // This is the easy one. We were active, now we aren't.
             self.activationState = .inactive
             self.channelClosed = true
+            self.demultiplexer = nil
```
Lukasa, line 281: This is a bit aggressive: we don't need to throw this away right now.

Sources/NIOHTTP2/InboundEventBuffer.swift
```diff
@@ -24,17 +24,27 @@ import NIOCore
...
+    fileprivate var buffer: CircularBuffer<BufferedHTTP2UserEvent> = CircularBuffer(initialCapacity: 8)
 
-    func pendingUserEvent(_ event: Any) {
+    func pendingUserEvent(_ event: BufferedHTTP2UserEvent) {
         self.buffer.append(event)
     }
+
+    /// Wraps user event types.
+    ///
+    /// This allows us to buffer and pass around the events without making use of a generic.
```
Lukasa, line 35: Strictly this was an existential.

Sources/NIOHTTP2/InboundEventBuffer.swift
```diff
@@ -24,17 +24,27 @@ import NIOCore
 /// have always delivered all pending user events before we deliver a frame.
 /// Deliberately not threadsafe or `Sendable`.
 class InboundEventBuffer {
-    fileprivate var buffer: CircularBuffer<Any> = CircularBuffer(initialCapacity: 8)
+    fileprivate var buffer: CircularBuffer<BufferedHTTP2UserEvent> = CircularBuffer(initialCapacity: 8)
 
-    func pendingUserEvent(_ event: Any) {
+    func pendingUserEvent(_ event: BufferedHTTP2UserEvent) {
```
Lukasa, line 29: To reduce the noise of this patch somewhat, it might be nicer to add individual `pendingUserEvent` functions for each of the types in the enum cases, and then the call sites don't have to change.

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -486,7 +493,16 @@ extension NIOHTTP2Handler {
...
+            switch event {
+            case .streamCreated(let event):
+                self.demultiplexer?.streamCreated(event.streamID, localInitialWindowSize: event.localInitialWindowSize, remoteInitialWindowSize: event.remoteInitialWindowSize)
+            case .initialStreamWindowChanged(let event):
+                self.demultiplexer?.initialStreamWindowChanged(by: event.delta)
+            case .streamWindowUpdated(let event):
+                self.demultiplexer?.streamWindowUpdated(event.streamID, inboundWindowSize: event.inboundWindowSize, outboundWindowSize: event.outboundWindowSize)
+            case .streamClosed(let event):
+                self.demultiplexer?.streamClosed(event.streamID, reason: event.reason)
+            }
```
Lukasa, line 505: Can we hide this switch in an extension on the demultiplexer? It'll again let us clean up this call site somewhat.

rnro: I have moved this to:
```
extension NIOHTTP2Handler.ConnectionDemultiplexer {
    func process(event: InboundEventBuffer.BufferedHTTP2UserEvent) {
 ```

File: Sources/NIOHTTP2/HTTP2ChannelHandler.swift [hunk 5]
```diff
@@ -672,9 +692,10 @@ extension NIOHTTP2Handler {
...
         case .streamCreated(let streamCreatedData):
             self.outboundBuffer.streamCreated(streamCreatedData.streamID, initialWindowSize: streamCreatedData.localStreamWindowSize.map(UInt32.init) ?? 0)
-            self.inboundEventBuffer.pendingUserEvent(NIOHTTP2StreamCreatedEvent(streamID: streamCreatedData.streamID,
-                                                                                localInitialWindowSize: streamCreatedData.localStreamWindowSize.map(UInt32.init),
-                                                                                remoteInitialWindowSize: streamCreatedData.remoteStreamWindowSize.map(UInt32.init)))
+            self.inboundEventBuffer.pendingUserEvent(
+                NIOHTTP2StreamCreatedEvent(streamID: streamCreatedData.streamID,
+                                                          localInitialWindowSize: streamCreatedData.localStreamWindowSize.map(UInt32.init),
+                                                          remoteInitialWindowSize: streamCreatedData.remoteStreamWindowSize.map(UInt32.init)))
```
Lukasa, line 698: Can you clean up the indentation here?

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -505,6 +510,21 @@ extension NIOHTTP2Handler {
...
+    func process(event: InboundEventBuffer.BufferedHTTP2UserEvent) {
+        switch event {
+        case .streamCreated(let event):
+            self.streamCreated(event.streamID, localInitialWindowSize: event.localInitialWindowSize, remoteInitialWindowSize: event.remoteInitialWindowSize)
+        case .initialStreamWindowChanged(let event):
+            self.initialStreamWindowChanged(by: event.delta)
+        case .streamWindowUpdated(let event):
+            self.streamWindowUpdated(event.streamID, inboundWindowSize: event.inboundWindowSize, outboundWindowSize: event.outboundWindowSize)
+        case .streamClosed(let event):
+            self.streamClosed(event.streamID, reason: event.reason)
```
Lukasa, line 523: Seeing it like this makes me ask the question: is there a good reason these functions aren't defined to simply take the event directly?

glbrntt: When I sketched out the API I didn't consider what was already there (i.e. the inbound event buffer). So no, no good reason to not use events directly.

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
rnro: I _think_ I have done this now. I have certainly changed it.

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
rnro: Changed to use the events directly.

Lukasa (approved):

PR #379, opened 2023-03-06
Add multiplexer protocol abstraction in http2 stream channel

rnro:
### Motivation:

This change prepares for future work to embed the connection multiplexer within the `HTTP2ChannelHandler` which will improve performance by removing the need for passing expensive `InboundUserEvent`s.

### Modifications:

* Define a new protocol (`OutboundStreamMultiplexer `; named such to avoid confusion with the legacy handler, `HTTP2StreamMultiplexer`) which will later be used to abstract over the inline and legacy multiplexer code but at the moment should not change the logic.
* Rename `NIOHTTP2ConnectionDemultiplexer` -> `InboundStreamMultiplexer`

### Result:

Behaviour should be unchanged.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamChannel+OutboundStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamChannel+OutboundStreamMultiplexer.swift
@@ -18,7 +18,7 @@ import NIOCore
 internal protocol HTTP2OutboundStreamMultiplexer {
   /// Write the frame into the HTTP/2 connection. Stream ID is included in the
   /// frame.
-  func writeStream(_ frame: HTTP2Frame, promise: EventLoopPromise<Void>?)
+  func writeFrame(_ frame: HTTP2Frame, promise: EventLoopPromise<Void>?)
 
   /// Flush the stream with the given ID.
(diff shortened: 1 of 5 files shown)


Sources/NIOHTTP2/HTTP2StreamChannel+OutboundStreamMultiplexer.swift
```diff
@@ -0,0 +1,123 @@
...
+//
+//===----------------------------------------------------------------------===//
+import NIOCore
+
+/// Multiplexes outbound HTTP/2 frames from an HTTP/2 stream into an HTTP/2
+/// connection.
+internal protocol HTTP2OutboundStreamMultiplexer {
+  /// Write the frame into the HTTP/2 connection. Stream ID is included in the
+  /// frame.
+  func writeStream(_ frame: HTTP2Frame, promise: EventLoopPromise<Void>?)
```
Lukasa, line 21:
```suggestion
  func writeFrame(_ frame: HTTP2Frame, promise: EventLoopPromise<Void>?)
```

Lukasa (changes requested):

Lukasa (approved):

PR #380, opened 2023-03-29
Implement direct stream (de)multiplexing in NIOHTTP2ChannelHandler

rnro:
Motivation:

Less expensive stream multiplexing. To implement direct stream multiplexing (with an incomplete public API) to break up delivery and allow testing.

Modifications:

* Factor out common inbound stream multiplexer functionality from `HTTP2StreamMultiplexer` into `CommonInboundStreamMultiplexer` (first commit).
* Direct stream multiplexing is implemented as `InlineStreamMultiplexer`. It implements both the `HTTP2InboundStreamMultiplexer` and `HTTP2OutboundStreamMultiplexer` protocols.
* We expose `NIOHTTP2Handler.StreamMultiplexer` as a `public` view over the new multiplexer to allow stream creation.
* Adds various `propagate...` functions to relay signals such as read complete to the child channels held by the inline stream multiplexer.
* Adds a stub for `HTTP2StreamDelegate`  which will be completed in an upcoming PR. It will be responsible for surfacing stream open/close information.
* Adds conveniences to `NIOHTTP2Handler` for setting up a handler using the inline multiplexer
* Duplicate existing tests and modify to use the inline stream multiplexer where appropriate. `SimpleClientServerFramePayloadStreamTests.swift` and `HTTP2FramePayloadStreamMultiplexerTests.swift` were copied and modified

Result:

Default behavior should be unchanged at this time.


Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -0,0 +1,205 @@
...
+        public func createStreamChannel(_ streamStateInitializer: @escaping StreamInitializer) -> EventLoopFuture<Channel> {
+            self.inlineStreamMultiplexer.createStreamChannel(streamStateInitializer)
+        }
+    }
+
+    /// A ``StreamMultiplexer`` which can be used to create new outbound HTTP/2 streams.
+    ///
+    /// > Note: The ``NIOHTTP2Handler`` must have been initialized with a ``StreamConfiguration``
+    /// for this value to be non-nil.
+    public var multiplexer: StreamMultiplexer? {
```
Lukasa, line 197: Can we document that this value is only non-`nil` after the handler has been added to the pipeline?

glbrntt: This is slightly awkward as we don't have an `EventLoop` on `init`. Adding one to the new inits is fine but we can't add them to the existing inits without breaking API. Rick and I discussed a couple of options. These are the most promising ones:

One is to have a future-based API which bangs the event loop since it will only ever be `nil` when using the old `init` (where we will never have a `StreamMultiplexer`).

The other is having a promise-based API which fails the promise if `self.eventLoop` is `nil`.

We think the first is fine since it'll only fail if the old soon-to-be-deprecated `init` is used and we can document it appropriately. WDYT @Lukasa?

rnro: I've gone ahead and implemented the first one as a basis of discussion.

Lukasa: I think that's fine by me. We may want to keep a non-future-based one as well, perhaps behind a `preconditionOnEventLoop` flag or similar.

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -869,3 +930,105 @@ extension NIOHTTP2Handler {
...
+        public var outboundBufferSizeHighWatermark: Int = 8196
+        public var outboundBufferSizeLowWatermark: Int = 4092
+        public init() {}
+    }
+
+    /// An `EventLoopFuture` which returns a ``StreamMultiplexer`` which can be used to create new outbound HTTP/2 streams.
+    ///
+    /// > Note: This is only safe to call if the ``NIOHTTP2Handler`` uses a local multiplexer,
+    /// i.e. it was initialized with an `inboundStreamInitializer`.
+    public func multiplexer() -> EventLoopFuture<StreamMultiplexer> {
```
Lukasa, line 1012: This can still be `var`. I think we should also keep a sync version around.

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -869,3 +930,111 @@ extension NIOHTTP2Handler {
...
+    public typealias StreamInitializer = @Sendable (Channel) -> EventLoopFuture<Void>
+#else
+    /// The type of all `inboundStreamInitializer` callbacks.
+    public typealias StreamInitializer = (Channel) -> EventLoopFuture<Void>
+#endif
+
+    /// Creates a new ``NIOHTTP2Handler`` with a local multiplexer. (i.e. using
+    /// ``StreamMultiplexer``.)
+    ///
+    /// Frames on the root stream will continue to be passed down the main channel, whereas those intended for other streams will be forwarded to the appropriate child channel.
```
glbrntt, line 967: Can you mention how to get the multiplexer to create streams here as well?

glbrntt (approved):

PR #384, opened 2023-04-05
Remove NIOHTTP2StreamDelegate Sendable requirement

rnro:
Motivation:

Remove `NIOHTTP2StreamDelegate` `Sendable` conformance requirement. This wasn't needed in the first place because `NIOHTTP2Handler` is not `Sendable`, therefore it never leaves the thread on which it is created.

The delegate may still be `Sendable` depending on the implementation but it is not a strict requirement.

Modifications:

Remove `NIOHTTP2StreamDelegate` `Sendable` requirement

Result:

Sendable is no longer required for `NIOHTTP2StreamDelegate` conformance


rnro:
API breakage is expected and is okay because we haven't cut a release since this API was added.

Lukasa (approved):

PR #385, opened 2023-04-06
Avoid allocation in outbound frame buffer

glbrntt:
Motivation:

CiruclarBuffer always allocates on init as it never has a zero capacity. It is implemented in such a way that changing this is not trivial. The compound outbound buffer creates a marked circular buffer unconditionally on each call but only uses it if there are streams to drop.

Modifications:

Add a static empty marked circular buffer to the compound outbound buffer and rely on CoW semantics to avoid the allocation for each call in cases where there are no frames to drop.

Result:

Fewer allocations.


Lukasa (approved):
Nice, great improvement!

PR #386, opened 2023-04-06
Inline multiplexer calls writability mgr directly

rnro:
Motivation:

The `HTTP2StreamChannel` keeps track of how many bytes have been written and how many are buffered for use in evaluating the 'writability' of the stream channel. Bytes are considered 'written' or at least no longer buffered when the write is completed (successfully or errored). In order to make the determination of when writes are complete across channel handlers we attach a promise to all writes and deduct adjust the byte counts on completion. Creating and attaching these promises requires costly allocations.

Modifications:

We can remove the need to create these promises for the new inline stream multiplexer because the multiplexer and man HTTP2 channel operations are now all within the `HTTP2ChannelHandler`. This change calls the `writabilityManager` directly when writes are completed.

Result:

* Fewer allocations are required for HTTP2 writes.
* Writes are considered 'complete' when they leave the `HTTP2ChannelHandler`

### DIFF
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler+InboundStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler+InboundStreamMultiplexer.swift
@@ -135,6 +135,15 @@ extension NIOHTTP2Handler.InboundStreamMultiplexer {
             break // do nothing
         }
     }
+
+    func processedFrame(streamID: HTTP2StreamID, size: Int) {
+        switch self {
+        case .inline(let inlineStreamMultiplexer):
+            inlineStreamMultiplexer.processedFrame(streamID: streamID, size: size)
+        case .legacy:
+            break // do nothing
+        }
+    }
 }
 
 /// Provides an inbound stream multiplexer interface for legacy compatibility.
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler+InboundStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler+InboundStreamMultiplexer.swift
@@ -136,10 +136,10 @@ extension NIOHTTP2Handler.InboundStreamMultiplexer {
         }
     }
 
-    func processedFrame(streamID: HTTP2StreamID, size: Int) {
+    func processedFrame(_ frame: HTTP2Frame) {
         switch self {
         case .inline(let inlineStreamMultiplexer):
-            inlineStreamMultiplexer.processedFrame(streamID: streamID, size: size)
+            inlineStreamMultiplexer.processedFrame(frame: frame)
         case .legacy:
             break // do nothing
         }
--- a/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
@@ -309,6 +309,12 @@ extension HTTP2CommonInboundStreamMultiplexer {
 }
 
 extension HTTP2CommonInboundStreamMultiplexer {
+    /// Mark the bytes as written
+    ///
+    /// > This is only to be called from the inline multiplexer
+    ///
+    /// Mark bytes as written in the `HTTP2StreamChannel` writability manager directly, as used by the inline stream multiplexer.
+    /// This is taken care of separately via a promise in the legacy case.
     internal func processedFrame(streamID: HTTP2StreamID, size: Int) {
         if let channel = self.streams[streamID] {
             channel.wroteBytes(size)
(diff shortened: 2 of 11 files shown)


rnro:
This change reduces the allocations for the `test_client_server_request_response` many (1000) requests allocation test from 1.3M to 1.0M.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -490,19 +496,30 @@ final class HTTP2StreamChannel: Channel, ChannelCore {
...
+        switch self.multiplexer {
+        case .inline:
+            promise = userPromise
+            break // do nothing at this stage, we'll be called back
+        case .legacy:
+            promise = userPromise ?? self.eventLoop.makePromise()
+            promise!.futureResult.hop(to: self.eventLoop).whenComplete { (_: Result<Void, Error>) in
+                if case .changed(newValue: let value) = self.writabilityManager.wroteBytes(writeSize) {
+                    self.changeWritability(to: value)
+                }
```
glbrntt, line 519: This can just be a call to `wroteBytes` now

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -490,19 +496,30 @@ final class HTTP2StreamChannel: Channel, ChannelCore {
             streamData = .framePayload(self.unwrapData(data))
         }
 
-        // We need a promise to attach our flow control callback to.
+        // For the legacy case we need a promise to attach our flow control callback to.
+        //
+        // This is no longer needed for the `.inline` stream multiplexer case for which this
+        // is all dealt with within the ```NIOHTTP2Handler``` so we can just
+        // call the `writabilityManager` directly.
```
glbrntt, line 503:
```suggestion
        // is all dealt with within the `NIOHTTP2Handler` which calls us back directly
        // via `wroteBytes`.
```

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -476,9 +476,15 @@ final class HTTP2StreamChannel: Channel, ChannelCore {
         fatalError("not implemented \(#function)")
     }
 
-    public func write0(_ data: NIOAny, promise: EventLoopPromise<Void>?) {
+    func wroteBytes(_ writeSize: Int) {
```
glbrntt, line 479: We should assert that we're in the right event loop when we enter this function.

glbrntt (commented):
I think we might have an issue we didn't consider before. In the stream channel we account for every frame which is sent to the parent channel. The estimated frame sizes include the 9-byte header of each frame.

However, in the HTTP2Handler DATA frames can be split up we would notify the stream channel about header bytes it knew nothing and our account would be wrong.

Lukasa:
It should be noted that the 9 byte frame header does not count against the flow control windows, so we do not need to consider it here.

glbrntt:
> It should be noted that the 9 byte frame header does not count against the flow control windows, so we do not need to consider it here.
> 
> [RFC 9113 § 6.9.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-6.9.1):
> 
> > For flow-control calculations, the 9-octet frame header is not counted.

Oh great! We just need to update `estimatedFrameSize` or the value we provide to `wroteBytes` to ignore that then.

Lukasa:
Yeah I think the fix is to drop the use of `estimatedFrameSize` altogether, tbh. I don't really know what I was thinking when I proposed it. Might be worth writing a separate patch to replace this with a `flowControlledSize` that correctly only covers that question.

rnro:
I've opened a PR for changing that `estimatedFrameSize` https://github.com/apple/swift-nio-http2/pull/387

Sources/NIOHTTP2/HTTP2ChannelHandler+InboundStreamMultiplexer.swift
```diff
@@ -135,6 +135,15 @@ extension NIOHTTP2Handler.InboundStreamMultiplexer {
             break // do nothing
         }
     }
+
+    func processedFrame(streamID: HTTP2StreamID, size: Int) {
```
Lukasa, line 139: I think this spelling is a bit weird, and it feels like a bit of a layering violation to me. I'm inclined to say we should pass the entire frame in here, and only do the size computation once we get inside the `inlineStreamMultiplexer`.

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -307,3 +307,11 @@ extension HTTP2CommonInboundStreamMultiplexer {
         }
     }
 }
+
+extension HTTP2CommonInboundStreamMultiplexer {
+    internal func processedFrame(streamID: HTTP2StreamID, size: Int) {
```
Lukasa, line 312: Hrm: this doesn't really feel like a "common" function at all. Can we avoid making it seem like one?

rnro: As discussed I've added a comment to explain that this should only be used from the inline multiplexer. Anything more substantial that I can think of either blurs responsibilities or adds a lot of complexity of abstractions.

Lukasa (approved):

PR #387, opened 2023-04-06
Replace estimatedFrameSize with flowControlledSize

rnro:
Motivation:

`estimatedFrameSize` aims to estimate the number of bytes a frame will use on the wire. It includes the 9-octed frame header and is currently used for flow control estimates but according to spec it shouldn't include those 9 bytes https://www.rfc-editor.org/rfc/rfc9113.html#section-6.9.1

Modifications:

Replace `estimatedFrameSize` with `flowControlledSize` which ignores the 9-octed frame header and update tests.

Result:

Flow control calculations should be more accurate

### DIFF
--- a/Sources/NIOHTTP2/HTTP2Frame.swift
+++ b/Sources/NIOHTTP2/HTTP2Frame.swift
@@ -315,37 +315,36 @@ public struct HTTP2Frame: Sendable {
 }
 
 extension HTTP2Frame.FramePayload {
-    /// A shorthand heuristic for how many bytes we assume a frame consumes on the wire.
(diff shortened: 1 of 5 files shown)


Lukasa:
Too late here, but we need to fix this up. `flowControlledSize` is a bad name for what this calculates. More broadly, I don't think we should calculate in this way at all: I don't think counting control frames against our window here makes any sense.

The only frames in HTTP/2 that are actually flow controlled are DATA frames. I think either the channels should only count DATA frames, _or_ we should rename this variable to something else.

rnro:
See https://github.com/apple/swift-nio-http2/pull/388

PR #393, opened 2023-05-17
Add `HPACKHeader.removeAll(keepingCapacity:)`

fabianfett:
In scenarios, in which we want to send lots of small HTTP/2 requests, we want to reuse a single HPACKHeaders object as often as we can. To support this we should have a `removeAll(keepingCapacity:)` method.

### DIFF
--- a/Sources/NIOHPACK/HPACKHeader.swift
+++ b/Sources/NIOHPACK/HPACKHeader.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2023 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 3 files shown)


Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -425,6 +425,15 @@ extension HPACKHeaders {
     public mutating func reserveCapacity(_ minimumCapacity: Int) {
         self.headers.reserveCapacity(minimumCapacity)
     }
+
+    /// Removes all headers whist optionally keeping the capacity.
```
Lukasa, line 429:
```suggestion
    /// Removes all headers whilst optionally keeping the capacity.
```

Sources/NIOHPACK/HPACKHeader.swift
```diff
@@ -425,6 +425,15 @@ extension HPACKHeaders {
...
+    ///
+    /// - Parameter keepingCapacity: Wether to keep the underlying memory or not. Use this flag to
```
Lukasa, line 431:
```suggestion
    /// - Parameter keepingCapacity: Whether to keep the underlying memory or not. Use this flag to
```

Lukasa (approved):

PR #395, opened 2023-05-18
Fix test for `HPACKHeader.removeAll(keepingCapacity:)` on Android

### LINKED ISSUE #394: The new HPACK header test for RemoveAll() fails on Android
#393 broke [my Android CI, which runs that test in an Android x86_64 emulator](https://github.com/finagolfin/swift-android-sdk/actions/runs/5001352241/jobs/8961762304#step:3:9221), with these 4 test failures:
```
/home/runner/work/swift-android-sdk/swift-android-sdk/snh/Tests/NIOHPACKTests/HPACKHeadersTests.swift:194: error: HPACKHeadersTests.testRemoveAll : XCTAssertEqual failed: ("7") is not equal to ("6") - 
/home/runner/work/swift-android-sdk/swift-android-sdk/snh/Tests/NIOHPACKTests/HPACKHeadersTests.swift:197: error: HPACKHeadersTests.testRemoveAll : XCTAssertEqual failed: ("7") is not equal to ("6") - 
/home/runner/work/swift-android-sdk/swift-android-sdk/snh/Tests/NIOHPACKTests/HPACKHeadersTests.swift:200: error: HPACKHeadersTests.testRemoveAll : XCTAssertEqual failed: ("7") is not equal to ("6") - 
/home/runner/work/swift-android-sdk/swift-android-sdk/snh/Tests/NIOHPACKTests/HPACKHeadersTests.swift:204: error: HPACKHeadersTests.testRemoveAll : XCTAssertEqual failed: ("7") is not equal to ("6") - 
```
All four are related to checking the capacity of the header array and not getting the expected size.

I can reproduce on my Android AArch64 phone, so I stuck these `print()`s in the initializer and found the offending line when the test is run:
```
diff --git a/Sources/NIOHPACK/HPACKHeader.swift b/Sources/NIOHPACK/HPACKHeader.swift
index 866025c..bd26df9 100644
--- a/Sources/NIOHPACK/HPACKHeader.swift
+++ b/Sources/NIOHPACK/HPACKHeader.swift
@@ -85,7 +85,11 @@ public struct HPACKHeaders: ExpressibleByDictionaryLiteral, Sendable {
     ///     - headers: An initial set of headers to use to populate the header block.
     @inlinable
     public init(_ headers: [(String, String)] = []) {
+        print(headers.count)
+        print(headers.capacity)
         self.headers = headers.map { HPACKHeader(name: $0.0, value: $0.1) }
+        print(self.headers.count)
+        print(self.headers.capacity)
     }
 
     /// Construct a `HPACKHeaders` structure.
```
That prints the following for this test:
```
Test Case 'HPACKHeadersTests.testRemoveAll' started at 2023-05-18 21:31:43.616
6
6
6
7
```
I have no idea if this `map` should be allocating a larger capacity or if this even matters: shouldn't the array capacity just be an implementation detail?

Opening this to see what you all think of this strange failure, let me know.

finagolfin:
Motivation:

Get this test passing on Android

Modifications:

Change the capacity to whatever it originally was, rather than a hard-coded value.

Result:

Resolves #394 and the test now passes

I could not reproduce on Android when assigning to a String array like this, implying this overcapacity is specific to this HPACK header array:
```
var headerArray : [String] = []

func assign(_ headers: [(String, String)] = []) {
   print(headers.count)
   print(headers.capacity)
   headerArray = headers.map { "headerKey : \($0.0), headerValue: \($0.1)" }
   print(headerArray.count)
   print(headerArray.capacity)
}
        let original = [
            ("foo", "bar"),
            ("bar", "foo"),
            ("foo", "baz"),
            ("foo", "bar,baz"),
            ("foo", " bar ,baz ,,"),
            ("bar", "foo")
        ]
assign(original)
```

### DIFF
--- a/Tests/NIOHPACKTests/HPACKHeadersTests.swift
+++ b/Tests/NIOHPACKTests/HPACKHeadersTests.swift
@@ -191,17 +191,17 @@ final class HPACKHeadersTests: XCTestCase {
             "bar": "foo"
         ]
 
-        XCTAssertEqual(original.capacity, 6)
+        let originalCapacity = original.capacity
         var keepCapacity = original
         XCTAssertEqual(keepCapacity.count, 6)
-        XCTAssertEqual(keepCapacity.capacity, 6)
+        XCTAssertEqual(keepCapacity.capacity, originalCapacity)
         keepCapacity.removeAll(keepingCapacity: true)
         XCTAssertEqual(keepCapacity.count, 0)
-        XCTAssertEqual(keepCapacity.capacity, 6)
+        XCTAssertEqual(keepCapacity.capacity, originalCapacity)
 
         var lowerCapacity = original
         XCTAssertEqual(lowerCapacity.count, 6)
-        XCTAssertEqual(lowerCapacity.capacity, 6)
+        XCTAssertEqual(lowerCapacity.capacity, originalCapacity)
         lowerCapacity.removeAll(keepingCapacity: false)
         XCTAssertEqual(lowerCapacity.count, 0)
         XCTAssertEqual(lowerCapacity.capacity, 0)


Lukasa (approved):
Love it, thanks for the fix!

PR #396, opened 2023-05-30
use synchronous H2 multiplexer accessor

rnro:
Motivation:

The `ChannelPipeline.SynchronousOperations.configureHTTP2Pipeline(...)` method added to support synchronously setting up a pipeline to handle HTTP2 with the `StreamMultiplexer` embedded within the `NIOHTTP2Handler` erroneously retrieved the multiplexer via a `.wait()`. This is not permitted and will cause any code which executes it to deadlock.

Modifications:

The synchrounous operation now obtains the multiplexer via `handler.syncMultiplexer()`

Result:

The code should not deadlock.


glbrntt (approved):

Lukasa (approved):

Lukasa:
Gotta fix the allocation counts!

PR #399, opened 2023-06-22
Replace os() with canImport(Darwin)

brennobemoura:
This will enable compiling the package to `xrOS` using the right code.

### DIFF
--- a/Sources/NIOHPACK/HuffmanCoding.swift
+++ b/Sources/NIOHPACK/HuffmanCoding.swift
@@ -268,7 +268,7 @@ extension String {
 // this declaration on having the 5.3 compiler. This has caused a number of build issues. While updating
 // to newer Xcodes does work, we can save ourselves some hassle and just wait until 5.4 to get this
 // enhancement on Apple platforms.
-#if (compiler(>=5.3) && !(os(macOS) || os(iOS) || os(tvOS) || os(watchOS))) || compiler(>=5.4)
+#if (compiler(>=5.3) && !(canImport(Darwin))) || compiler(>=5.4)
 extension String {
     init(customUnsafeUninitializedCapacity capacity: Int,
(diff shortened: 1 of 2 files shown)


FranzBusch (approved):

Lukasa:

FranzBusch:

brennobemoura:
@FranzBusch is this test failure related to the modification I've made?

FranzBusch:
No this looks like our allocation counters need to be updated because we regressed somewhere. We should check where that allocation regression comes from.

Sources/NIOHPACK/HuffmanCoding.swift
```diff
@@ -268,7 +268,7 @@ extension String {
 // this declaration on having the 5.3 compiler. This has caused a number of build issues. While updating
 // to newer Xcodes does work, we can save ourselves some hassle and just wait until 5.4 to get this
 // enhancement on Apple platforms.
-#if (compiler(>=5.3) && !(os(macOS) || os(iOS) || os(tvOS) || os(watchOS))) || compiler(>=5.4)
+#if (compiler(>=5.3) && !(canImport(Darwin))) || compiler(>=5.4)
```
Lukasa, line 271: We can drop this condition entirely now, as we no longer support Swift versions older than 5.6, so we will always enter this branch.

Lukasa (changes requested):

Lukasa:
Alloc limits fix in #400

Sources/NIOHPACK/HuffmanCoding.swift
brennobemoura: Fixing this right now.

brennobemoura:
@Lukasa great catch! You nailed it.

Lukasa (approved):
Love it, negative diffs are the best. :sparkles:

PR #401, opened 2023-06-26
Generic helpers for HTTP/2 async pipelines

rnro:
Expose (as spi) generic helpers for HTTP/2 async pipelines

Motivation:

This PR is the first step in exposing APIs which surface HTTP/2 connections/streams using structured concurrency. This PR exposes the most abstract spelling of this concept, assuming no particular types/forms for the types involved in the stream channel types.

Modifications:

* Define types and methods for storing generic types which wrap `Channels` corresponding to inbound HTTP/2 streams in an async sequence.
* Store a generic `streamInitializer` on the common inbound initializer which is used to initialize inbound streams before yielding them to the continuation of streams.
* Expose pipeline configuration functions which assume HTTP/2 but nothing else about sream channel types.
* Provide internal functions for creating streams without configuring them. Configuring leads to activation so it can be helpful to allow that step to be performed manually once any provided initialization closures have beenn executed.

Result:
 
Adopters of the new SPI should be able to create outbound and deal with inbound HTTP/2 stream channels using async streams.

Outside of SPI there should be no changes.


Package.swift
```diff
@@ -17,6 +17,9 @@ import PackageDescription
 
 let package = Package(
     name: "swift-nio-http2",
+    platforms: [
+        .macOS("10.15")
+    ],
```
glbrntt, line 22: We can't add this; it's an API breaking change. You can use `@available` guards in the code though.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -194,6 +194,39 @@ extension Channel {
...
+    ///     - position: The position in the pipeline into which to insert this handler.
+    ///     - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    /// - returns: An `EventLoopFuture` containing the `AsyncThrowingStream` of inbound HTTP/2 stream channels and  the `AsyncStreamMultiplexer` inserted into this pipeline, which can be used to initiate new streams.
+    @_spi(AsyncChannel)
+    public func configureHTTP2PipelineAsync<InboundStreamChannelType>(mode: NIOHTTP2Handler.ParserMode,
+                                       connectionConfiguration: NIOHTTP2Handler.ConnectionConfiguration,
+                                       streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
+                                       streamDelegate: NIOHTTP2StreamDelegate? = nil,
+                                       position: ChannelPipeline.Position = .last,
+                                       inboundStreamInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<InboundStreamChannelType>) -> EventLoopFuture<(AsyncThrowingStream<InboundStreamChannelType, any Error>, NIOHTTP2Handler.AsyncStreamMultiplexer)> {
```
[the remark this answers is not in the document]
FranzBusch: I was kind of expecting a different return type here.
1. We should create our own concrete type like `NIOHTTP2InboundStreamChannels` which holds the `AsyncThrowingStream` under the hood. This allows us to change the underlying root async sequence in the future.
2. We should create a concrete type that holds both the `NIOHTTP2InboundStreamChannels` and the `AsyncStreamMultiplexer`
3. I am wondering if we want to expose the `AsyncStreamMultiplexer` at all or rather just a method on the new type that allows you to create a new outbound stream channel
4. We probably have to return the connection channel itself as well since we are adding handlers to it and changing the inbound and outbound types there. We might want to wrap that into a `NIOAsyncChannel` unconditionally.

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,150 @@
...
+import XCTest
+
+import NIOCore
+import NIOEmbedded
+import NIOHPACK
+import NIOHTTP1
+@_spi(AsyncChannel) import NIOHTTP2
+import NIOTLS
+
+class ConfiguringPipelineAsyncMultiplexerTests: XCTestCase {
```
FranzBusch, line 24:
```suggestion
final class ConfiguringPipelineAsyncMultiplexerTests: XCTestCase {
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -377,4 +410,51 @@ extension ChannelPipeline.SynchronousOperations {
...
+                                                  mode: NIOHTTP2Handler.ParserMode,
+                                                  connectionConfiguration: NIOHTTP2Handler.ConnectionConfiguration,
+                                                  streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
+                                                  streamDelegate: NIOHTTP2StreamDelegate? = nil,
+                                                  position: ChannelPipeline.Position = .last,
+                                                  streamChannelType: InboundStreamChannelType.Type = InboundStreamChannelType.self,
+                                                  inboundStreamInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<InboundStreamChannelType>
+    ) throws -> (AsyncThrowingStream<InboundStreamChannelType, Error>, NIOHTTP2Handler.AsyncStreamMultiplexer) {
+
+        let handler = NIOHTTP2Handler(
```
FranzBusch, line 442: Should we precondition here that we are on the EL?

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -201,3 +205,50 @@ extension NIOHTTP2Handler {
...
+extension InlineStreamMultiplexer {
+    func setStreamChannels(_ streamChannels: any ContinuationSanitizer) {
+        self.commonStreamMultiplexer.setStreamChannels(streamChannels)
+    }
+}
+
+extension NIOHTTP2Handler {
+    /// A variant of `NIOHTTP2Handler.StreamMultiplexer` which creates a child channel for each HTTP/2 stream and yields it to the supplied async continuation.
+
+    /// In general in NIO applications it is helpful to consider each HTTP/2 stream as an
```
glbrntt, line 218: I assume these are meant to be joined:

```suggestion
    /// A variant of `NIOHTTP2Handler.StreamMultiplexer` which creates a child channel for each HTTP/2 stream and yields it to the supplied async continuation.
    ///
    /// In general in NIO applications it is helpful to consider each HTTP/2 stream as an
```

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,150 @@
...
+                fatalError("unexpected frame type: \(frame)")
+            }
+            context.fireChannelRead(data)
+        }
+    }
+
+    func testBasicPipelineCommunicates() async throws {
+        let requestCount = 100
+
+        let serverRecorder = InboundFramePayloadRecorder()
```
FranzBusch, line 89: This is not safe to do. You cannot construct the type and hand it to the initializer since the `InboundFramePayloadRecorder` is not `Sendable`. I would recommend creating an `AsyncStream` and handing the continuation to the `InboundFramePayloadRecorder` inside the `streamChannelInitializer`. You can then await on the stream itself in your test.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -201,3 +205,50 @@ extension NIOHTTP2Handler {
...
 }
+
+extension InlineStreamMultiplexer {
+    func setStreamChannels(_ streamChannels: any ContinuationSanitizer) {
+        self.commonStreamMultiplexer.setStreamChannels(streamChannels)
+    }
+}
+
+extension NIOHTTP2Handler {
+    /// A variant of `NIOHTTP2Handler.StreamMultiplexer` which creates a child channel for each HTTP/2 stream and yields it to the supplied async continuation.
```
glbrntt, line 216:
```suggestion
    /// A variant of `NIOHTTP2Handler.StreamMultiplexer` which creates a child channel for each HTTP/2 stream and yields it to the supplied async continuation.
```

We don't need to mention continuations here: it's an implementation details and this documentation should be aimed at users. We should only talk about creating new streams here.

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,150 @@
...
+            // client
+            for _ in 0 ..< requestCount {
+                // Let's try sending some requests
+                let streamChannel = try await clientMultiplexer.createStreamChannel() { channel in
+                    return channel.pipeline.addHandlers([SimpleRequest(), InboundFramePayloadRecorder()]).map {
+                        return channel
+                    }
+                }
+
+                let clientRecorder = try await streamChannel.pipeline.handler(type: InboundFramePayloadRecorder.self).get()
```
FranzBusch, line 124: That's also not safe since the handler is not `Sendable`. You should be able to do the same trick here with the `AsyncStream`

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -201,3 +205,50 @@ extension NIOHTTP2Handler {
...
+
+    /// In general in NIO applications it is helpful to consider each HTTP/2 stream as an
+    /// independent stream of HTTP/2 frames. This multiplexer achieves this by creating a
+    /// number of in-memory `HTTP2StreamChannel` objects, one for each stream. These operate
+    /// on ``HTTP2Frame/FramePayload`` objects as their base communication
+    /// atom, as opposed to the regular NIO `SelectableChannel` objects which use `ByteBuffer`
+    /// and `IOData`.
+    ///
+    /// The stream channel objects are initialized upon creation using the supplied `streamStateInitializer` which returns a type
+    /// `StreamChannelType` which must match the communication atom (final handler's InboundOut/OutboundIn types). This type may
```
glbrntt, line 226: I think we should rename `StreamChannelType` to be `InitializationResult`. `StreamChannelType` suggests it has to be a `Channel` which isn't the case; it can be anything.

I also don't think this documentation is accurate anymore: it it doesn't have to match the final handlers types.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -201,3 +205,50 @@ extension NIOHTTP2Handler {
...
+    /// still be `HTTP2Frame` or changed to any other type.
+    @_spi(AsyncChannel)
+    public struct AsyncStreamMultiplexer {
+        private let inlineStreamMultiplexer: InlineStreamMultiplexer
+
+        /// Cannot be created by users.
```
glbrntt, line 232: This seems like a code comment (`//`) rather than a doc comment (`///`).

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,150 @@
...
+                try await self.interactInMemory(self.clientChannel, self.serverChannel)
+
+                await self.clientChannel.testingEventLoop.run()
+                try await streamChannel.closeFuture.get()
+
+                clientRecorder.receivedFrames.assertFramePayloadsMatch([ConfiguringPipelineAsyncMultiplexerTests.responseFramePayload])
+            }
+
+            async let closeFuture: Void = self.clientChannel.close()
```
FranzBusch, line 133: Do you have a particular reason for using an `async let` here? If the reason is because you need to run the EL to make the close actually close then just use the `close(promise: nil)` variant instead.

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,150 @@
...
+    override func tearDown() {
+        self.clientChannel = nil
+        self.serverChannel = nil
+    }
+
+
+    static let requestFramePayload = HTTP2Frame.FramePayload.headers(.init(headers: HPACKHeaders([(":method", "GET"), (":authority", "localhost"), (":scheme", "https"), (":path", "/")]), endStream: true))
+    static let responseFramePayload = HTTP2Frame.FramePayload.headers(.init(headers: HPACKHeaders([(":status", "200")]), endStream: true))
+
+    class OKResponder: ChannelInboundHandler {
```
FranzBusch, line 44: Same for the rest here
```suggestion
    final class OKResponder: ChannelInboundHandler {
```

FranzBusch (changes requested):
Great start! Left some inline comments here and there

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -201,3 +205,50 @@ extension NIOHTTP2Handler {
...
+        internal init(_ inlineStreamMultiplexer: InlineStreamMultiplexer, continuation: any ContinuationSanitizer) {
+            self.inlineStreamMultiplexer = inlineStreamMultiplexer
+            self.inlineStreamMultiplexer.setStreamChannels(continuation)
+        }
+
+        public func createStreamChannel<StreamChannelType>(_ streamStateInitializer: @escaping (Channel) -> EventLoopFuture<StreamChannelType>) async throws -> StreamChannelType {
+            return try await self.inlineStreamMultiplexer.createStreamChannel().flatMap { channel in
+                streamStateInitializer(channel).flatMap { streamChannel in
```
glbrntt, line 240: I don't think `streamChannel` is a `Channel` here, it's a `StreamChannelType`, right?

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -201,3 +205,50 @@ extension NIOHTTP2Handler {
...
+                    let promise = channel.eventLoop.makePromise(of: Channel.self)
+                    if let http2StreamChannel = channel as? HTTP2StreamChannel {
```
glbrntt, line 242: I think we can `as!` this. If it isn't an `HTTP2StreamChannel` something's gone quite wrong, I think.

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +362,80 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+
+extension HTTP2CommonInboundStreamMultiplexer {
+    func setStreamChannels(_ streamChannels: any ContinuationSanitizer) {
+        self.streamChannels = streamChannels
+    }
+}
+
+/// `ContinuationSanitizer` is used to type-erase generic async-sequence-like objects. This is so that they may be held
+/// by the `HTTP2ChannelHandler` without causing it to become generic itself.
+internal protocol ContinuationSanitizer {
```
glbrntt, line 374: Let's call this `ChannelContinuation`

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -93,7 +102,14 @@ extension HTTP2CommonInboundStreamMultiplexer {
             )
 
             self.streams[streamID] = channel
+
+            // If we have an async sequence of inbound stream channels yield the channel to it
+            // This also implicitly performs the stream initialization step
+            self.streamChannels?.yield(channel: channel.baseChannel)
```
glbrntt, line 108: What does `channel.configureInboundStream` do  below? (Just checking we don't init the channel twice...)

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -93,7 +102,14 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+
             channel.configureInboundStream(initializer: self.inboundStreamStateInitializer)
+
+
```
glbrntt, line 112: Let's kill these new blank lines

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +362,80 @@ extension HTTP2CommonInboundStreamMultiplexer {
         }
     }
 }
+
+extension HTTP2CommonInboundStreamMultiplexer {
+    func setStreamChannels(_ streamChannels: any ContinuationSanitizer) {
```
glbrntt, line 367: Let's call this `setChannelContinuation` (and in other places)

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +362,80 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+/// `ContinuationSanitizer` is used to type-erase generic async-sequence-like objects. This is so that they may be held
+/// by the `HTTP2ChannelHandler` without causing it to become generic itself.
+internal protocol ContinuationSanitizer {
+    func yield(channel: Channel)
+    func finish()
+    func finish(throwing error: Error)
+}
+
+/// `StreamChannels` is a wrapper for a generic `AsyncThrowingStream` which holds the inbound HTTP2 stream channels.
+struct StreamChannels<StreamChannelType>: ContinuationSanitizer {
```
glbrntt, line 381: Let's call this `StreamChannelContinuation`

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +362,80 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+    var continuation: AsyncThrowingStream<StreamChannelType, Error>.Continuation
```
glbrntt, line 382: should this be private?

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -194,6 +194,39 @@ extension Channel {
         }
     }
 
+    /// Configures a `ChannelPipeline` to speak HTTP/2 and sets up mapping functions so that it may be interacted with from concurrent code.
+    ///
+    /// In general this is not entirely useful by itself, as HTTP/2 is a negotiated protocol. This helper does not handle negotiation.
+    /// Instead, this simply adds the handler required to speak HTTP/2 after negotiation has completed, or when agreed by prior knowledge.
+    /// Use this function to setup a HTTP/2 server pipeline if you wish to use async sequence abstractions over inbound and outbound streams, as it allows that pipeline to evolve without breaking your code.
```
glbrntt, line 201:
```suggestion
    /// Use this function to setup a HTTP/2 pipeline if you wish to use async sequence abstractions over inbound and outbound streams, as it allows that pipeline to evolve without breaking your code.
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
glbrntt, line 218: This formatting is ... interesting. The parameters are aligned in two different places (`mode` vs. the-rest) and the final line is super long.

```suggestion
    public func configureHTTP2PipelineAsync<InboundStreamChannelType>(
        mode: NIOHTTP2Handler.ParserMode,
        connectionConfiguration: NIOHTTP2Handler.ConnectionConfiguration,
        streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
        streamDelegate: NIOHTTP2StreamDelegate? = nil,
        position: ChannelPipeline.Position = .last,
        inboundStreamInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<InboundStreamChannelType>
    ) -> EventLoopFuture<(AsyncThrowingStream<InboundStreamChannelType, any Error>, NIOHTTP2Handler.AsyncStreamMultiplexer)> {
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -377,4 +410,51 @@ extension ChannelPipeline.SynchronousOperations {
...
+    @_spi(AsyncChannel)
+    public func configureHTTP2PipelineAsync<InboundStreamChannelType>(
+                                                  mode: NIOHTTP2Handler.ParserMode,
+                                                  connectionConfiguration: NIOHTTP2Handler.ConnectionConfiguration,
+                                                  streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
+                                                  streamDelegate: NIOHTTP2StreamDelegate? = nil,
+                                                  position: ChannelPipeline.Position = .last,
+                                                  streamChannelType: InboundStreamChannelType.Type = InboundStreamChannelType.self,
+                                                  inboundStreamInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<InboundStreamChannelType>
+    ) throws -> (AsyncThrowingStream<InboundStreamChannelType, Error>, NIOHTTP2Handler.AsyncStreamMultiplexer) {
```
glbrntt, line 440:
```suggestion
    public func configureHTTP2PipelineAsync<InboundStreamChannelType>(
        mode: NIOHTTP2Handler.ParserMode,
        connectionConfiguration: NIOHTTP2Handler.ConnectionConfiguration,
        streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
        streamDelegate: NIOHTTP2StreamDelegate? = nil,
        position: ChannelPipeline.Position = .last,
        streamChannelType: InboundStreamChannelType.Type = InboundStreamChannelType.self,
        inboundStreamInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<InboundStreamChannelType>
    ) throws -> (AsyncThrowingStream<InboundStreamChannelType, Error>, NIOHTTP2Handler.AsyncStreamMultiplexer) {
```

glbrntt (changes requested):

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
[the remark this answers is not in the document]
glbrntt: > We can't use `any` here since we have to support older versions. Just drop the `any` for now

We can use `any`: it was added in 5.6 which is our minimum supported version. (https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md)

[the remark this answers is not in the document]
glbrntt: > I was kind of expecting a different return type here.
> 
> 1. We should create our own concrete type like `NIOHTTP2InboundStreamChannels` which holds the `AsyncThrowingStream` under the hood. This allows us to change the underlying root async sequence in the future.
> 2. We should create a concrete type that holds both the `NIOHTTP2InboundStreamChannels` and the `AsyncStreamMultiplexer`
> 3. I am wondering if we want to expose the `AsyncStreamMultiplexer` at all or rather just a method on the new type that allows you to create a new outbound stream channel

Agree.

> 4. We probably have to return the connection channel itself as well since we are adding handlers to it and changing the inbound and outbound types there. We might want to wrap that into a `NIOAsyncChannel` unconditionally.

We are already holding the connection channel (this is an extension on `Channel`) so we don't need to return it. I also don't think we should return `NIOAsyncChannel` _here_, that will come later. This is a lower-level building block which most users will never use.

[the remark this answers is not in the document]
FranzBusch: Makes sense!

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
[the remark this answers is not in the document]
rnro: I think we are type-erasing, we're erasing the generic `Output` type are we not?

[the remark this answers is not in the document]
FranzBusch: Not really. In this type, we are specialising the generic type of `AsyncThrowingStream<Element>.Continuation` to be `Channel`. We do this so that we can avoid spelling out the `Output` type inside the multiplexer/handler. While it has similar properties has type erasure I wouldn't really call it this here :D
Overall I think we should just change the name of the protocol `ContinuationSanitizer` seems very vague and the type does something different that I expected from the name.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -201,3 +205,50 @@ extension NIOHTTP2Handler {
...
+    public struct AsyncStreamMultiplexer {
+        private let inlineStreamMultiplexer: InlineStreamMultiplexer
+
+        /// Cannot be created by users.
+        internal init(_ inlineStreamMultiplexer: InlineStreamMultiplexer, continuation: any ContinuationSanitizer) {
+            self.inlineStreamMultiplexer = inlineStreamMultiplexer
+            self.inlineStreamMultiplexer.setStreamChannels(continuation)
+        }
+
+        public func createStreamChannel<StreamChannelType>(_ streamStateInitializer: @escaping (Channel) -> EventLoopFuture<StreamChannelType>) async throws -> StreamChannelType {
```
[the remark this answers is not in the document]
rnro: Yeah `streamStateInitializer` is established in the existing similar functions. I made the other changes.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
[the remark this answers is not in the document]
rnro: Not with the current code, but I can imagine them falling out of step. We could put an assertion in here to protect against it rather than silently failing gracefully.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -201,3 +205,50 @@ extension NIOHTTP2Handler {
...
+            return try await self.inlineStreamMultiplexer.createStreamChannel().flatMap { channel in
+                streamStateInitializer(channel).flatMap { streamChannel in
+                    let promise = channel.eventLoop.makePromise(of: Channel.self)
+                    if let http2StreamChannel = channel as? HTTP2StreamChannel {
+                        http2StreamChannel.configure(initializer: { channel in channel.eventLoop.makeSucceededVoidFuture() }, userPromise: promise)
```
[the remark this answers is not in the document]
rnro: Yeah. I didn't do that initially because I was trying to avoid changing too much existing infrastructure and generics leading to API changes - but that said I think we can do better here. I've made a duplicate of the `configure` function which I think makes it more readable and rolls in your suggestion.

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +362,80 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+    private let streamInitializer: (Channel) -> EventLoopFuture<StreamChannelType>
```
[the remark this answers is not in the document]
rnro: Aligned with `streamStateInitializer`

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
rnro: It _used_ to be required for the test to pass but changing the recorders to use streams seems to have fixed that issue

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
rnro: I changed the generic type to `Output` to match the server bootstrap bind method.

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
rnro: `channel.configureInboundStream` uses the `inboundStreamStateInitializer` `from the HTTP2Handler` which simply returns a void successful future.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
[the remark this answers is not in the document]
rnro: I have made these changes but had the `AsyncStreamMultiplexer` hold the `NIOHTTP2InboundStreamChannels` so that we continue with the tradition of being the object with which we create outbound channels.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -194,6 +194,42 @@ extension Channel {
...
+    ///     - connectionConfiguration: The settings that will be used when establishing the connection. These will be sent to the peer as part of the
+    ///         handshake.
+    ///     - streamConfiguration: The settings that will be used when establishing new streams. These mainly pertain to flow control.
+    ///     - streamDelegate: The delegate to be notified in the event of stream creation and close.
+    ///     - position: The position in the pipeline into which to insert this handler.
+    ///     - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    /// - returns: An `EventLoopFuture` containing the `AsyncThrowingStream` of inbound HTTP/2 stream channels and  the `AsyncStreamMultiplexer` inserted into this pipeline, which can be used to initiate new streams.
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    @_spi(AsyncChannel)
+    public func configureHTTP2PipelineAsync<Output>(
```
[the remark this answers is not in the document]
rnro: I went with `configureAstncHTTP2Pipeline` because we're configuring an async pipeline but the configuration itself isn't async - which I think the previous one implied.

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +380,113 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+    func finish(throwing error: Error)
+}
+
+
+/// `StreamChannelContinuation` is a wrapper for a generic `AsyncThrowingStream` which holds the inbound HTTP2 stream channels.
+///
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+struct StreamChannelContinuation<Output>: ChannelContinuation {
+    private var continuation: AsyncThrowingStream<Output, Error>.Continuation
+    private let streamStateInitializer: (Channel) -> EventLoopFuture<Output>
```
glbrntt, line 405: We should be explicit in that this is for _inbound_ streams: `inboundStreamInititializer`

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +380,113 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+
+    private init(
+        continuation: AsyncThrowingStream<Output, Error>.Continuation,
+        streamStateInitializer: @escaping (Channel) -> EventLoopFuture<Output>
+    ) {
+        self.continuation = continuation
+        self.streamStateInitializer = streamStateInitializer
+    }
+
+    /// `initialize` creates a new `StreamChannels` object and returns it along with its backing `AsyncThrowingStream`.
```
glbrntt, line 415:
```suggestion
    /// `initialize` creates a new `StreamChannelContinuation` object and returns it along with its backing `AsyncThrowingStream`.
```

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +380,113 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+    ///
+    /// - Parameters:
+    ///   - streamInitializer: A closure which initializes the newly-created inbound stream channel and returns a generic.
+    ///   The returned type corresponds to the output of the channel once the operations in the initializer have been performed.
+    ///   For example a `streamStateInitializer` which inserts handlers before wrapping the channel in a `NIOAsyncChannel` would
+    ///   have a `Output` corresponding to that `NIOAsyncChannel` type. Another example is in cases where there is
+    ///   per-stream protocol negotiation where `Output` would be some form of `NIOProtocolNegotiationResult`.
+    static func initialize(
+        with streamStateInitializer: @escaping (Channel) -> EventLoopFuture<Output>
```
glbrntt, line 424: lets make it explicit that this is for inbound streams

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +380,113 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+        var continuation: AsyncThrowingStream<Output, Error>.Continuation? = nil
+        let stream = AsyncThrowingStream { continuation = $0 }
+        return (StreamChannelContinuation(continuation: continuation!, streamStateInitializer: streamStateInitializer), NIOHTTP2InboundStreamChannels(stream))
+    }
+
+    /// `yield` takes a channel, executes the stored `streamInitializer` upon it and then yields the *derived* type to
+    /// the wrapped `AsyncThrowingStream`.
+    func yield(channel: Channel) {
+        channel.eventLoop.assertInEventLoop()
+        self.streamStateInitializer(channel).whenSuccess{ streamChannel in
```
glbrntt, line 435:
```suggestion
        self.streamStateInitializer(channel).whenSuccess { streamChannel in
```

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +380,113 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+    /// `finish` marks the continuation as finished with the supplied error.
+    func finish(throwing error: Error) {
+        self.continuation.finish(throwing: error)
+    }
+}
+
+/// `NIOHTTP2InboundStreamChannels` provides access to inbound stream channels as an `AsyncSequence`.
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+@_spi(AsyncChannel)
+public struct NIOHTTP2InboundStreamChannels<Output>: AsyncSequence {
```
glbrntt, line 466: We should mark this as `Sendable` when `Output` is also `Sendable`:

```swift
extension NIOHTTP2InboundStreamChannels: Sendable where Output: Sendable {}
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
glbrntt: This one hasn't been fixed.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -224,22 +219,47 @@ final class HTTP2StreamChannel: Channel, ChannelCore {
...
+            case .success(let autoRead):
+                self.autoRead = autoRead
+                initializer(self).whenComplete { result in
+                    self.onInitializationResult(result, promise: promise)
+                }
+            case .failure(let error):
+                self.configurationFailed(withError: error, promise: promise)
+            }
+        }
+
```
glbrntt, line 251:
```suggestion
```

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
glbrntt: this one hasn't been done

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    func testBasicPipelineCommunicates() async throws {
+        let requestCount = 100
+
+
+        let (serverRecorderStream, serverRecorderContinuation) = AsyncStream<InboundRecorder<HTTP2Frame.FramePayload>>.makeStream()
+        var serverRecorderIterator = serverRecorderStream.makeAsyncIterator()
+
+        let clientMultiplexer = try await assertNoThrowWithValue(
```
glbrntt, line 93: FWIW: `XCTAssertNoThrow` was valuable because Xcode provided pretty bad errors when something was thrown; the errors are much better now so I tend to think `XCTAssertNoThrow` (and by extension `assertNoThrowWithValue`) aren't really necessary. There's no need to remove it but I won't ask you to use it if you don't!

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,175 @@
...
+                for try await _ in serverMultiplexer.inbound {
+                    serverInboundChannelCount += 1
+                }
+                return serverInboundChannelCount
+            }
+
+            // client
+            for _ in 0 ..< requestCount {
+                // Let's try sending some requests
+                let streamChannel = try await clientMultiplexer.createStreamChannel() { channel -> EventLoopFuture<Channel> in
```
glbrntt, line 128:
```suggestion
                let streamChannel = try await clientMultiplexer.createStreamChannel { channel -> EventLoopFuture<Channel> in
```

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,175 @@
...
+                    let clientRecorder = InboundFramePayloadRecorder()
+                    clientRecorderContinuation.yield(clientRecorder)
+                    return channel.pipeline.addHandlers([SimpleRequest(), clientRecorder]).map {
+                        return channel
+                    }
+                }
+
+                try await self.interactInMemory(self.clientChannel, self.serverChannel)
+
+                await self.clientChannel.testingEventLoop.run()
```
glbrntt, line 138: I know we put this in while debugging, is it still necessary?

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,175 @@
...
+
+                let clientRecorder = await clientRecorderIterator.next()!
+                clientRecorder.receivedFrames.assertFramePayloadsMatch([ConfiguringPipelineAsyncMultiplexerTests.responseFramePayload])
+            }
+
+            self.clientChannel.close(promise: nil)
+            await self.clientChannel.testingEventLoop.run()
+            group.cancelAll()
+
+            let serverInboundChannelCount =  try await assertNoThrowWithValue(try await group.next()!)
```
glbrntt, line 149:
```suggestion
            let serverInboundChannelCount = try await assertNoThrowWithValue(try await group.next()!)
```

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -62,6 +62,32 @@ extension XCTestCase {
...
+    /// Have two `NIOAsyncTestingChannel` objects send and receive data from each other until
+    /// they make no forward progress.
+    func interactInMemory(_ first: NIOAsyncTestingChannel, _ second: NIOAsyncTestingChannel, file: StaticString = #filePath, line: UInt = #line) async throws {
+        var operated: Bool
+
+        func readBytesFromChannel(_ channel: NIOAsyncTestingChannel) async -> ByteBuffer? {
+            guard let data = try? await assertNoThrowWithValue(await channel.readOutbound(as: ByteBuffer.self)) else {
+                return nil
+            }
+            return data
```
glbrntt, line 74: Isn't this just:
```suggestion
            return try? await assertNoThrowWithValue(await channel.readOutbound(as: ByteBuffer.self))
```

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -62,6 +62,32 @@ extension XCTestCase {
...
+        repeat {
+            operated = false
+
+            if let data = await readBytesFromChannel(first) {
+                operated = true
+                try await assertNoThrow(try await second.writeInbound(data), file: (file), line: line)
+            }
+            if let data = await readBytesFromChannel(second) {
+                operated = true
+                try await assertNoThrow(try await first.writeInbound(data), file: (file), line: line)
```
glbrntt, line 86: The parentheses around `file` were to squash a warning a long time ago, they're not necessary anymore.

```suggestion
                try await assertNoThrow(try await second.writeInbound(data), file: file, line: line)
            }
            if let data = await readBytesFromChannel(second) {
                operated = true
                try await assertNoThrow(try await first.writeInbound(data), file: file, line: line)
```

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -110,6 +136,43 @@ extension XCTestCase {
...
+    func assertDoHandshake(client: NIOAsyncTestingChannel, server: NIOAsyncTestingChannel,
+                           clientSettings: [HTTP2Setting] = nioDefaultSettings, serverSettings: [HTTP2Setting] = nioDefaultSettings,
+                           file: StaticString = #filePath, line: UInt = #line) async throws {
+        // This connects are not semantically right, but are required in order to activate the
+        // channels.
+        //! FIXME: Replace with registerAlreadyConfigured0 once EmbeddedChannel propagates this
+        //         call to its channelcore.
+        let socket = try SocketAddress(unixDomainSocketPath: "/fake")
+        _ = try await client.connect(to: socket).get()
+        _ = try await server.connect(to: socket).get()
```
glbrntt, line 153: I think we do this in our test already? We should avoid doing it twice.

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -110,6 +136,43 @@ extension XCTestCase {
...
+        let serverReceivedSettingsAck = try await server.assertReceivedFrame(file: (file), line: line)
+
+        // Check that these SETTINGS frames are ok.
+        clientReceivedSettings.assertSettingsFrame(expectedSettings: serverSettings, ack: false, file: (file), line: line)
+        serverReceivedSettings.assertSettingsFrame(expectedSettings: clientSettings, ack: false, file: (file), line: line)
+        clientReceivedSettingsAck.assertSettingsFrame(expectedSettings: [], ack: true, file: (file), line: line)
+        serverReceivedSettingsAck.assertSettingsFrame(expectedSettings: [], ack: true, file: (file), line: line)
+
+        await client.assertNoFramesReceived(file: (file), line: line)
+        await server.assertNoFramesReceived(file: (file), line: line)
```
glbrntt, line 173: We can remove the parens around `file`

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -193,6 +256,53 @@ extension EmbeddedChannel {
...
+        var frameDecoder = HTTP2FrameDecoder(allocator: self.allocator, expectClientMagic: false)
+        while let buffer = try await assertNoThrowWithValue(await self.readOutbound(as: ByteBuffer.self), file: (file), line: line) {
+            frameDecoder.append(bytes: buffer)
+            if let (frame, _) = try frameDecoder.nextFrame() {
+                receivedFrames.append(frame)
+            }
+        }
+
+        return receivedFrames
+    }
```
glbrntt, line 303: Are these used anywhere?

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -790,3 +900,50 @@ func assertNoThrowWithValue<T>(_ body: @autoclosure () throws -> T, defaultValue
     }
 }
 
+func assertNoThrowWithValue<T>(_ body: @autoclosure () async throws -> T, defaultValue: T? = nil, message: String? = nil, file: StaticString = #filePath, line: UInt = #line) async throws -> T {
```
glbrntt, line 903: nit: long lines!

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
rnro: It's not - removed

Tests/NIOHTTP2Tests/TestUtilities.swift
rnro: Not yet, I just ported them over when I did the others in case they were useful but I'll delete them for now and pull them over when needed

Tests/NIOHTTP2Tests/TestUtilities.swift
rnro: It's not not that

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -201,3 +205,41 @@ extension NIOHTTP2Handler {
...
+        public let inbound: NIOHTTP2InboundStreamChannels<InboundStreamOutput>
+
+        // Cannot be created by users.
+        internal init(_ inlineStreamMultiplexer: InlineStreamMultiplexer, continuation: any ChannelContinuation, inboundStreamChannels: NIOHTTP2InboundStreamChannels<InboundStreamOutput>) {
+            self.inlineStreamMultiplexer = inlineStreamMultiplexer
+            self.inlineStreamMultiplexer.setChannelContinuation(continuation)
+            self.inbound = inboundStreamChannels
+        }
+
+        public func createStreamChannel<OutboundStreamOutput>(_ streamStateInitializer: @escaping NIOHTTP2Handler.StreamInitializerWithOutput<OutboundStreamOutput>) async throws -> OutboundStreamOutput {
```
FranzBusch, line 241: I know that `streamStateInitializer` is aligned with the previous APIs but I just want to take one more stab at this discussion. Should we align this would our bootstrap init namings and other packages so it is called `streamChannelInitializer`? cc @glbrntt @Lukasa

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -201,3 +205,41 @@ extension NIOHTTP2Handler {
...
+    /// atom, as opposed to the regular NIO `SelectableChannel` objects which use `ByteBuffer`
+    /// and `IOData`.
+    ///
+    /// Outbound stream channel objects are initialized upon creation using the supplied `streamStateInitializer` which returns a type
+    /// `OutboundStreamOutput`. This type may be `HTTP2Frame` or changed to any other type.
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    @_spi(AsyncChannel)
+    public struct AsyncStreamMultiplexer<InboundStreamOutput> {
+        private let inlineStreamMultiplexer: InlineStreamMultiplexer
+        public let inbound: NIOHTTP2InboundStreamChannels<InboundStreamOutput>
```
FranzBusch, line 232: One potential problem I see with this is: If you are an HTTP server and only expect to create outbound streams on any incoming connection so you are not consuming the `inbound` stream here at all. However, some malicious clients will still create inbound streams we will accumulate them in the `NIOHTTP2InboundStreamChannels`.

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -990,9 +990,13 @@ extension NIOHTTP2Handler {
 #if swift(>=5.7)
     /// The type of all `inboundStreamInitializer` callbacks.
     public typealias StreamInitializer = @Sendable (Channel) -> EventLoopFuture<Void>
+    /// The type of `inboundStreamInitializer` callbacks which return non-void results.
+    public typealias StreamInitializerWithOutput<Output> = @Sendable (Channel) -> EventLoopFuture<Output>
```
FranzBusch, line 994: Normally we try to avoid type aliases because they become quite hard to discover. Personally I would prefer if we actually spell this out all the time.

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +406,116 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+    ///   - inboundStreamInititializer: A closure which initializes the newly-created inbound stream channel and returns a generic.
+    ///   The returned type corresponds to the output of the channel once the operations in the initializer have been performed.
+    ///   For example an `inboundStreamInititializer` which inserts handlers before wrapping the channel in a `NIOAsyncChannel` would
+    ///   have a `Output` corresponding to that `NIOAsyncChannel` type. Another example is in cases where there is
+    ///   per-stream protocol negotiation where `Output` would be some form of `NIOProtocolNegotiationResult`.
+    static func initialize(
+        with inboundStreamInititializer: @escaping (Channel) -> EventLoopFuture<Output>
+    ) -> (StreamChannelContinuation<Output>, NIOHTTP2InboundStreamChannels<Output>) {
+        var continuation: AsyncThrowingStream<Output, Error>.Continuation? = nil
+        let stream = AsyncThrowingStream { continuation = $0 }
```
FranzBusch, line 453: NIT: In Swift 5.9+ we are going to have a method on `AsyncStream` which makes this easier called `makeStream`. This method returns the stream and the continuation. Instead of doing this here could we add this extension here ourselves surrounded with `#if` that guard the Swift version so that we can remove it in the future.

The extensions looks like this
```
extension AsyncThrowingStream {
  public static func makeStream(
      of elementType: Element.Type = Element.self,
      throwing failureType: Failure.Type = Failure.self,
      bufferingPolicy limit: Continuation.BufferingPolicy = .unbounded
  ) -> (stream: AsyncThrowingStream<Element, Failure>, continuation: AsyncThrowingStream<Element, Failure>.Continuation) where Failure == Error {
    var continuation: AsyncThrowingStream<Element, Failure>.Continuation!
    let stream = AsyncThrowingStream<Element, Failure>(bufferingPolicy: limit) { continuation = $0 }
    return (stream: stream, continuation: continuation!)
  }
}
```

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +406,116 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+        return (StreamChannelContinuation(continuation: continuation!, inboundStreamInititializer: inboundStreamInititializer), NIOHTTP2InboundStreamChannels(stream))
+    }
+
+    /// `yield` takes a channel, executes the stored `streamInitializer` upon it and then yields the *derived* type to
+    /// the wrapped `AsyncThrowingStream`.
+    func yield(channel: Channel) {
+        channel.eventLoop.assertInEventLoop()
+        self.inboundStreamInititializer(channel).whenSuccess { streamChannel in
```
FranzBusch, line 461: I think the parameter here should be called `output`
```suggestion
        self.inboundStreamInititializer(channel).whenSuccess { output in
```

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +406,116 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+            case .enqueued:
+                break // success, nothing to do
+            case .dropped:
+                preconditionFailure("Attempted to yield channel when AsyncThrowingStream is over capacity. This shouldn't be possible for an unbounded stream.")
+            case .terminated:
+                channel.close(mode: .all, promise: nil)
+                assertionFailure("Attempted to yield channel to AsyncThrowingStream in terminated state.")
+            default:
+                channel.close(mode: .all, promise: nil)
+                assertionFailure("Attempt to yield channel to AsyncThrowingStream failed for unhandled reason.")
```
FranzBusch, line 473: We should probably precondition in both cases here. From the code above we should never hit those without it being an error on our side.

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +406,116 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+
+    init(_ asyncThrowingStream: AsyncThrowingStream<Output, Error>) {
+        self.asyncThrowingStream = asyncThrowingStream
+    }
+
+    public func makeAsyncIterator() -> AsyncIterator {
+        AsyncIterator(self.asyncThrowingStream.makeAsyncIterator())
+    }
+}
+
```
FranzBusch, line 519: Can we add an explicit `Sendable` unavailability on the iterator. We do this to make sure that nobody is using this across tasks.

```
@available(*, unavailable)
extension NIOHTTP2InboundStreamChannels.AsyncIterator: Sendable {}
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -194,6 +194,56 @@ extension Channel {
...
+    ///     - connectionConfiguration: The settings that will be used when establishing the connection. These will be sent to the peer as part of the
+    ///         handshake.
+    ///     - streamConfiguration: The settings that will be used when establishing new streams. These mainly pertain to flow control.
+    ///     - streamDelegate: The delegate to be notified in the event of stream creation and close.
+    ///     - position: The position in the pipeline into which to insert this handler.
+    ///     - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    /// - returns: An `EventLoopFuture` containing the `AsyncStreamMultiplexer` inserted into this pipeline, which can be used to initiate new streams and iterate over inbound HTTP/2 stream channels.
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    @_spi(AsyncChannel)
+    public func configureAsyncHTTP2Pipeline<Output>(
```
FranzBusch, line 214: Do we have to add `Async` in the name here? For the bootstrap methods we also just overloaded the `bind` and `connect` names. (Don't remember if I brought this up previously already)

Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -224,22 +219,46 @@ final class HTTP2StreamChannel: Channel, ChannelCore {
...
-                    self.postInitializerActivate(promise: promise)
+                    self.postInitializerActivate(promise: promise, output: self)
                 }
             case .failure(let error):
                 self.configurationFailed(withError: error, promise: promise)
             }
         }
     }
 
+    internal func configure<Output>(initializer: @escaping NIOHTTP2Handler.StreamInitializerWithOutput<Output>, userPromise promise: EventLoopPromise<Output>?) {
```
FranzBusch, line 233: We might one to add `@inlinable` to all the generic methods here now so that they get properly specialised. This needs to go up the whole call stack. Might be something for the future though once we drop the SPI.

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,175 @@
...
+
+        await self.clientChannel.assertNoFramesReceived()
+        await self.serverChannel.assertNoFramesReceived()
+
+        try await assertNoThrow(try await self.clientChannel.finish(acceptAlreadyClosed: true))
+        try await assertNoThrow(try await self.serverChannel.finish())
+    }
+}
+
+extension AsyncStream {
```
FranzBusch, line 166: Let's guard this to below 5.9 since we ship it with 5.9 in the toolchains

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,175 @@
...
+                mode: .client, connectionConfiguration: .init(), streamConfiguration: .init()) { channel -> EventLoopFuture<Channel> in
+                    self.serverChannel.eventLoop.makeSucceededFuture(channel)
+                }.get()
+        )
+
+        let serverMultiplexer = try await assertNoThrowWithValue(
+            try await self.serverChannel.configureAsyncHTTP2Pipeline(
+                mode: .server, connectionConfiguration: .init(), streamConfiguration: .init()) { channel -> EventLoopFuture<Channel> in
+                    let serverRecorder = InboundFramePayloadRecorder()
+                    serverRecorderContinuation.yield(serverRecorder)
```
FranzBusch, line 102: I don't think this is safe. The `InboundFramePayloadRecorder` is probably not `Sendable`. What I meant with my previous comment was actually passing the continuation into this recorder so that the recorder is yielding whatever it records to it. (We might want to create a complete new type for this instead of botching the `InboundFramePayloadRecorder`)

FranzBusch (commented):
This looks super good already and I think we are getting quite close here!

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
glbrntt: I prefer `streamChannelInitializer` over `streamStateInitializer`. 

Just throwing it out there... in this instance we have the context of creating a stream channel so maybe `initializer` is sufficient?

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -265,31 +283,98 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
         return promise.futureResult
     }
 
     @available(*, deprecated, message: "The signature of 'streamStateInitializer' has changed to '(Channel) -> EventLoopFuture<Void>'")
-    internal func createStreamChannel(multiplexer: HTTP2StreamChannel.OutboundStreamMultiplexer, promise: EventLoopPromise<Channel>?, _ streamStateInitializer: @escaping (Channel, HTTP2StreamID) -> EventLoopFuture<Void>) {
+    internal func createStreamChannel(
+        multiplexer: HTTP2StreamChannel.OutboundStreamMultiplexer,
+        promise: EventLoopPromise<Channel>?,
+        _ streamStateInitializer: @escaping (Channel, HTTP2StreamID
+        ) -> EventLoopFuture<Void>) {
```
glbrntt, line 377:
```suggestion
    internal func createStreamChannel(
        multiplexer: HTTP2StreamChannel.OutboundStreamMultiplexer,
        promise: EventLoopPromise<Channel>?,
        _ streamStateInitializer: @escaping (Channel, HTTP2StreamID) -> EventLoopFuture<Void>
    ) {
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
glbrntt: I think we end up being overloaded on just the initialiser and return type otherwise. We couldn't come up with a better name... :/

Tests/NIOHTTP2Tests/HTTP2StreamMultiplexerTests.swift
```diff
@@ -91,13 +92,16 @@ typealias FramePayloadWriteRecorder = WriteRecorder<HTTP2Frame.FramePayload>
...
     typealias InboundIn = Frame
 
+    let framesLock = NIOLock()
     var receivedFrames: [Frame] = []
 
     func channelRead(context: ChannelHandlerContext, data: NIOAny) {
-        self.receivedFrames.append(self.unwrapInboundIn(data))
+        self.framesLock.withLock {
+            self.receivedFrames.append(self.unwrapInboundIn(data))
+        }
```
glbrntt, line 104: This is not _quite_ right. 

`receivedFrames` is still accessible. Our normal convention is to rename the protected state to be prefixed with `_` and then have a whatever-access-level-is-approproate computed property which acquires and releases the lock when accessing the protected state. Within the object holding the state and the lock you can manually acquire it if, for example, you need finer grained lock control (i.e. it protects more than one piece of state and you need to modify both).

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,175 @@
...
+                fatalError("unexpected frame type: \(frame)")
+            }
+
+            context.channel.writeAndFlush(responseFramePayload, promise: nil)
+
+            context.fireChannelRead(data)
+        }
+    }
+
+    class SimpleRequest: ChannelInboundHandler, ChannelOutboundHandler {
```
glbrntt, line 59:
```suggestion
    final class SimpleRequest: ChannelInboundHandler, ChannelOutboundHandler {
```

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,175 @@
...
+        func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+            let frame = self.unwrapInboundIn(data)
+            switch frame {
+            case .headers:
+                break
+            default:
+                fatalError("unexpected frame type: \(frame)")
+            }
+            context.fireChannelRead(data)
+        }
```
glbrntt, line 81: Do we need to bother with this? Our test should assert on the frames it receives anyway.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
glbrntt: Yeah this is a bit of an awkward one. I'm inclined to think that if you don't expect any inbound streams the you should set max concurrent streams to zero so that the other side can't open any streams then this is safe to ignore.

FranzBusch: It is a reoccurring pattern w.r.t. async sequences where if you store them as a state of something it might just buffer indefinitely. I still haven't made my mind up about this. This same is true for `NIOAsyncChannel` itself.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
rnro: `initializer` 🚀

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
rnro: In general I agree but the increased code duplication for the Sendable and non-Sendable versions in already duplicate-heavy code I think would be worse in this case.

FranzBusch: Fair enough!

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: Yep, what @glbrntt said.

Sources/NIOHTTP2/HTTP2StreamChannel.swift
rnro: Yeah I think we should do this once the API has settled down and we remove SPI

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
rnro: Yeah I realized afterwards that this is probably not what you meant however I think it works as a convenient way to get the recorders out of the initializers and indicate when they're 'done'. I have made `InboundFramePayloadRecorder` `Sendable` to make it safe.

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
rnro: Removed

Tests/NIOHTTP2Tests/HTTP2StreamMultiplexerTests.swift
```diff
@@ -91,10 +92,23 @@ typealias FramePayloadWriteRecorder = WriteRecorder<HTTP2Frame.FramePayload>
 
 
 /// A handler that keeps track of all reads made on a channel.
-final class InboundRecorder<Frame>: ChannelInboundHandler {
+final class InboundRecorder<Frame>: ChannelInboundHandler, @unchecked Sendable {
     typealias InboundIn = Frame
 
-    var receivedFrames: [Frame] = []
+    let framesLock = NIOLock()
```
glbrntt, line 98:
```suggestion
    private let framesLock = NIOLock()
```

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,167 @@
...
+
+        let (serverRecorderStream, serverRecorderContinuation) = AsyncStream<InboundRecorder<HTTP2Frame.FramePayload>>.makeStream()
+        var serverRecorderIterator = serverRecorderStream.makeAsyncIterator()
+
+        let clientMultiplexer = try await assertNoThrowWithValue(
+            try await self.clientChannel.configureAsyncHTTP2Pipeline(
+                mode: .client, connectionConfiguration: .init(), streamConfiguration: .init()) { channel -> EventLoopFuture<Channel> in
+                    self.serverChannel.eventLoop.makeSucceededFuture(channel)
+                }.get()
+        )
```
glbrntt, line 85: Note also that this fixes the event loop used in the init:

```suggestion
            try await self.clientChannel.configureAsyncHTTP2Pipeline(
                mode: .client, 
                connectionConfiguration: .init(), 
                streamConfiguration: .init()
            ) { channel -> EventLoopFuture<Channel> in
                channel.eventLoop.makeSucceededFuture(channel)
            }.get()
        )
```

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,167 @@
...
+
+        let serverMultiplexer = try await assertNoThrowWithValue(
+            try await self.serverChannel.configureAsyncHTTP2Pipeline(
+                mode: .server, connectionConfiguration: .init(), streamConfiguration: .init()) { channel -> EventLoopFuture<Channel> in
+                    let serverRecorder = InboundFramePayloadRecorder()
+                    serverRecorderContinuation.yield(serverRecorder)
+                    return channel.pipeline.addHandlers([OKResponder(), serverRecorder]).map { _ in return channel }
+                }.get()
```
glbrntt, line 93:
```suggestion
            try await self.serverChannel.configureAsyncHTTP2Pipeline(
                mode: .server, 
                connectionConfiguration: .init(), 
                streamConfiguration: .init()
            ) { channel -> EventLoopFuture<Channel> in
                let serverRecorder = InboundFramePayloadRecorder()
                serverRecorderContinuation.yield(serverRecorder)
                return channel.pipeline.addHandlers([OKResponder(), serverRecorder]).map { _ in channel }
            }.get()
```

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,167 @@
...
+                    }
+                }
+
+                try await self.interactInMemory(self.clientChannel, self.serverChannel)
+                try await streamChannel.closeFuture.get()
+
+                let clientRecorder = await clientRecorderIterator.next()!
+                clientRecorder.framesLock.withLock {
+                    clientRecorder.receivedFrames.assertFramePayloadsMatch([ConfiguringPipelineAsyncMultiplexerTests.responseFramePayload])
+                }
```
glbrntt, line 129: We don't need an async stream here, we can just return the recorder from the init with the channel (or recover it from the channel pipeline) and then check the appropriate frames are there after the stream has been closed.

Note also that you don't need the lock here: `clientRecorder.receviedFrames` acquires the lock for you.

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,167 @@
...
+            let serverInboundChannelCount = try await assertNoThrowWithValue(try await group.next()!)
+            XCTAssertEqual(serverInboundChannelCount, requestCount, "We should have created one server-side channel as a result of the one HTTP/2 stream used.")
+        }
+
+        for _ in 0 ..< requestCount {
+            let serverRecorder = await serverRecorderIterator.next()!
+            serverRecorder.framesLock.withLock {
+                serverRecorder.receivedFrames.assertFramePayloadsMatch([ConfiguringPipelineAsyncMultiplexerTests.requestFramePayload])
+            }
+        }
```
glbrntt, line 145: Because the recorder is now `Sendable` you don't need an async stream. You can create the recorder at the start of the test and insert it into each server handler (sharing handler's isn't usually safe but in this case it's fine). We can then just assert it holds the correct frames here.

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -265,31 +283,97 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+        } else {
+            self.channel.eventLoop.execute {
+                self._createStreamChannel(multiplexer, promise, streamStateInitializer)
+            }
+        }
+    }
+
+    internal func createStreamChannel(
+        multiplexer: HTTP2StreamChannel.OutboundStreamMultiplexer,
+        _ streamStateInitializer: @escaping (Channel) -> EventLoopFuture<Void>) -> EventLoopFuture<Channel> {
```
glbrntt, line 365:
```suggestion
        _ streamStateInitializer: @escaping (Channel) -> EventLoopFuture<Void>
    ) -> EventLoopFuture<Channel> {
```

glbrntt: Does this need to be a `NIOHTTP2Handler.StreamInitializer` so we get appropriate `Sendable` annotations?

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +405,142 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+    /// The `StreamChannelContinuation` provides access to the inbound HTTP2 stream channels.
+    ///
+    /// - Parameters:
+    ///   - inboundStreamInititializer: A closure which initializes the newly-created inbound stream channel and returns a generic.
+    ///   The returned type corresponds to the output of the channel once the operations in the initializer have been performed.
+    ///   For example an `inboundStreamInititializer` which inserts handlers before wrapping the channel in a `NIOAsyncChannel` would
+    ///   have a `Output` corresponding to that `NIOAsyncChannel` type. Another example is in cases where there is
+    ///   per-stream protocol negotiation where `Output` would be some form of `NIOProtocolNegotiationResult`.
+    static func initialize(
+        with inboundStreamInititializer: @escaping (Channel) -> EventLoopFuture<Output>
```
glbrntt, line 449: should this be `NIOHTTP2Handler.StreamInitializerWithOutput<Output>`?

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -321,3 +405,142 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+    func finish()
+    func finish(throwing error: Error)
+}
+
+
+/// `StreamChannelContinuation` is a wrapper for a generic `AsyncThrowingStream` which holds the inbound HTTP2 stream channels.
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+struct StreamChannelContinuation<Output>: ChannelContinuation {
+    private var continuation: AsyncThrowingStream<Output, Error>.Continuation
+    private let inboundStreamInititializer: (Channel) -> EventLoopFuture<Output>
```
glbrntt, line 429: should this be `NIOHTTP2Handler.StreamInitializerWithOutput<Output>`?

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -0,0 +1,148 @@
...
+    final class SimpleRequest: ChannelInboundHandler, ChannelOutboundHandler {
+        typealias InboundIn = HTTP2Frame.FramePayload
+        typealias OutboundIn = HTTP2Frame.FramePayload
+
+        func writeRequest(context: ChannelHandlerContext) {
+            context.channel.writeAndFlush(requestFramePayload, promise: nil)
+        }
+
+        func channelActive(context: ChannelHandlerContext) {
+            self.writeRequest(context: context)
```
glbrntt, line 68: nit: this should be `context.writeAndFlush`, `context.channel.writeAndFlush` will write into the handler at the further point from the network which is rarely what you want from inside a channel handler.

```suggestion
        func channelActive(context: ChannelHandlerContext) {
            context.writeAndFlush(requestFramePayload, promise: nil)
```

glbrntt (commented):
A couple of nits, but very nearly there!

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
rnro: I don’t think that the common one can be `NIOHTTP2Handler.StreamInitializer` because it bubbles up to the legacy `StreamMultiplexer` create stream which is public and doesn’t have `Sendable` annotations.

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
rnro: Yes, good catch

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -61,7 +61,7 @@ final class ConfiguringPipelineAsyncMultiplexerTests: XCTestCase {
         typealias OutboundIn = HTTP2Frame.FramePayload
 
         func writeRequest(context: ChannelHandlerContext) {
-            context.channel.writeAndFlush(requestFramePayload, promise: nil)
+            context.writeAndFlush(NIOAny(requestFramePayload), promise: nil)
```
glbrntt, line 64: Avoid using `NIOAny` within a handler, prefer using `self.wrapOutboundOut` when writing (and `self.wrapInboundOut` for `fireChannelRead`)

glbrntt (approved):

FranzBusch (approved):
LGTM!

PR #404, opened 2023-07-04
HTTP/2 NIOAsyncChannel standard types pipeline config

rnro:
Motivation:

This continues the work to expose functionality which allows users to interact with HTTP/2 connections via async abstractions and using structured concurrency.

We build atop of previous work to configure pipelines to deal with HTTP/2 (with no protocol negotiation) and to wrap connection/stream channels with `NIOAsyncChannel`s. This will allow users to iterate over streams and HTTP2Frames.

This PR exposes conveniences for configuring HTTP/2 pipelines in the simplest case where the user knows that any inserted channel handlers will not change the `InboundOut`/`OutboundIn` types of the pipeline handlers.

Modifications:

Provide functions which specialize configuration functions which exposed generic `NIOAsyncChannel`s with standard `HTTP2Frame` types.

Result:

Users will be able to create and interact with HTTP/2 connections via `NIOAsyncChannel`s by using simpler configuration functions. Because HTTP/2 is a negotiated protocol and we do not yet handle it, this is of limited utility.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
+++ b/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
@@ -327,6 +327,65 @@ extension Channel {
         }
     }
 
+    /// Configures a `ChannelPipeline` to speak HTTP/2 and wraps any created inbound stream channels in `NIOAsyncChannel`s
+    /// expecting standard HTTP/2 frame types.
+    ///
+    /// In general this is not entirely useful by itself, as HTTP/2 is a negotiated protocol. This helper does not handle negotiation.
+    /// Instead, this simply adds the handler required to speak HTTP/2 after negotiation has completed, or when agreed by prior knowledge.
+    /// Use this function to setup a HTTP/2 pipeline if you wish to use `NIOAsyncChannel`s to provide async sequence abstractions
+    /// over inbound and outbound streams whilst handling back-pressure and you have no need to change the inbound/outbound types of the stream channel pipelines.
+    ///
+    /// This convenience adds the restriction that the final handler added to the stream channel by the `inboundStreamInitializer`
+    /// must have `HTTP2Frame.Payload` as its `InboundOut` and `OutboundIn` types.
+    ///
+    /// Using this rather than implementing a similar function yourself allows that pipeline to evolve without breaking your code.
+    ///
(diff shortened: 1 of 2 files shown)


Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -327,6 +327,65 @@ extension Channel {
...
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    @_spi(AsyncChannel)
+    public func configureAsyncHTTP2Pipeline(
+        mode: NIOHTTP2Handler.ParserMode,
+        connectionConfiguration: NIOHTTP2Handler.ConnectionConfiguration,
+        streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
+        streamDelegate: NIOHTTP2StreamDelegate? = nil,
+        position: ChannelPipeline.Position = .last,
+        inboundStreamInitializer: @escaping NIOHTTP2Handler.StreamInitializer
+    ) throws -> EventLoopFuture<NIOHTTP2Handler.AsyncStreamMultiplexer<NIOAsyncChannel<HTTP2Frame.FramePayload, HTTP2Frame.FramePayload>>> {
```
glbrntt, line 363: I'm not sure this adds any value over this: https://github.com/apple/swift-nio-http2/blob/1ba485c5507a9c807cca0cf9432547ebca3a51ab/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift#L291-L300 

It's effectively the same but assumes the stream types to be `HTTP2Frame.FramePayload` but the user can still mess up the types with the initialiser. There's maybe a small amount of value in offering this _without_ the stream initialiser but I'm actually not convinced it's worth having.

rnro: I could be convinced otherwise but I thought that these would be useful for people who wanted to just go straight into async-land without having to deal with any of the complexity. To that end I wondered about removing the stream initializer too. Maybe such a user doesn't exist though.

FranzBusch: Yeah I agree with George. I would leave this out for now and see if this is requested in the future.

PR #403, opened 2023-07-03
HTTP/2 pipeline NIOAsyncChannel pipeline config

rnro:
Motivation:

This continues the work to expose functionality which allows users to interact with HTTP/2 connections via async abstractions and using structured concurrency.

We build atop of previous work to configure pipelines to deal with HTTP/2 (with no protocol negotiation) and to wrap connection/stream channels with `NIOAsyncChannel`s. This will allow users to iterate over streams and HTTP2Frames.

Modifications:

Provide functions which configure channels and pipelines with the HTTP2 handler, expose a multiplexer for dealing with streams and wrap connection and stream channels with `NIOAsyncChannel`s.

Result:

Users will be able to create and interact with HTTP/2 connections via `NIOAsyncChannel`s. Because HTTP/2 is a negotiated protocol and we do not yet handle it, this is of limited utility.

PR #406, opened 2023-07-04
Add additional NIOAsyncChannel config

rnro:
Motivation:

Recent additions to pipeline configuration did not expose some of the `NIOAsyncChannel` configuration.

Modifications:

Expose `backpressureStrategy`, `isOutboundHalfClosureEnabled`.

I think specific naming and object encapsulation for these parameters should be discussed later before SPI is removed and we review the API as a whole.

Result:

Pipelie helpers for `NIOAsyncChannel` expose `backpressureStrategy`, `isOutboundHalfClosureEnabled`.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
@@ -258,8 +258,8 @@ extension NIOHTTP2Handler {
         @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
         @_spi(AsyncChannel)
         public func createStreamChannel<Inbound, Outbound>(
-            inboundType: Inbound.Type,
-            outboundType: Outbound.Type,
+            inboundType: Inbound.Type = Inbound.self,
+            outboundType: Outbound.Type = Inbound.self,
             backpressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark? = nil,
             isOutboundHalfClosureEnabled: Bool = false,
             initializer: @escaping NIOHTTP2Handler.StreamInitializer
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
@@ -247,6 +247,8 @@ extension NIOHTTP2Handler {
         /// Create a stream channel initialized with the provided closure and return it wrapped within a `NIOAsyncChannel`.
         ///
         /// - Parameters:
+        ///     - backpressureStrategy: The backpressure strategy of the ``NIOAsyncChannel`` wrapping the HTTP/2 stream channel.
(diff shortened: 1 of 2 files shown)


Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -255,7 +255,11 @@ extension NIOHTTP2Handler {
         ///       or ``HTTP2Frame/FramePayload`` if there are none.
         ///     - initializer: A callback that will be invoked to allow you to configure the
         ///         `ChannelPipeline` for the newly created channel.
+        @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+        @_spi(AsyncChannel)
         public func createStreamChannel<Inbound, Outbound>(
+            backpressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark? = nil,
+            isOutboundHalfClosureEnabled: Bool = false,
             inboundType: Inbound.Type,
             outboundType: Outbound.Type,
```
glbrntt, line 264: can we reorder these so `{in,out}boundType` come first? Required parameters should come first.

FranzBusch: We should add defaults to the types as well like this `= Inbound.self`. Then the order is the same as in our bootstrap

rnro: I don't think we want to add defaults here to force people to consider what types are used. This was also mentioned in an earlier PR https://github.com/apple/swift-nio-http2/pull/403#discussion_r1250706741

FranzBusch: This is not really a default. If you write `inboundType: Inbound.Type = Inbound.self` it will allow the user to either write out the generic return type which will get inferred or pass parameters. If you don't add this default value then they always have to write out the parameters even though type inference doesn't need it.

rnro: Yep, good point. I've added these in the latest commit and put the parameters back in the order to match `NIOAsyncChannel`.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -258,10 +258,10 @@ extension NIOHTTP2Handler {
         @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
         @_spi(AsyncChannel)
         public func createStreamChannel<Inbound, Outbound>(
+            inboundType: Inbound.Type = Inbound.self,
```
FranzBusch, line 261: This order seems different than the other methods

FranzBusch (approved):
LGTM except one minor nit

glbrntt (approved):

rnro:
The API breakages are safe because these changes were added since the last release.

rnro:
The API breakages are safe because these changes were added since the last release.

PR #408, opened 2023-07-05
HTTP/2 negotiated connection pipeline config

rnro:
Motivation:

This continues the work to expose functionality which allows users to interact with HTTP/2 connections via async abstractions and using structured concurrency.

This work enables protocol negotiation between HTTP/1.1 and HTTP/2 in its most generic form.

Modifications:

Enable protocol negotiation between HTTP/1.1 and HTTP/2 using typed negotiation results (`NIOProtocolNegotiationResult`) from the `NIOTypedApplicationProtocolNegotiationHandler`.
In the HTTP/2 case the negotiation result will return the HTTP/2 connection channel and the `NIOHTTP2Handler.AsyncStreamMultiplexer` which exposes inbound streams as an iterable async stream.

Result:

Users will be able to set up negotiated HTTP/1.1 / HTTP/2 connections and iterate over inbound streams.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
+++ b/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
@@ -660,10 +660,9 @@ extension ChannelPipeline.SynchronousOperations {
             eventLoop: self.eventLoop,
             connectionConfiguration: connectionConfiguration,
             streamConfiguration: streamConfiguration,
-            streamDelegate: streamDelegate
-        ) { channel in
-            inboundStreamInitializer(channel)
-        }
+            streamDelegate: streamDelegate,
+            inboundStreamInitializer: inboundStreamInitializer
+        )
 
         try self.addHandler(handler, position: position)
 
@@ -772,7 +771,7 @@ public enum NIONegotiatedHTTPVersion<HTTP1Output: Sendable, HTTP2Output: Sendabl
 /// `NIOHTTP2AsyncConfiguration` contains all configuration required for setting up an HTTP/2 async connection.
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
 @_spi(AsyncChannel)
-public struct NIOHTTP2AsyncConfiguration<HTTP2ConnectionInbound: Sendable, HTTP2ConnectionOutbound: Sendable, HTTP2StreamInbound: Sendable, HTTP2StreamOutbound: Sendable>: Sendable {
+public struct NIOHTTP2AsyncConfiguration<HTTP2ConnectionInbound: Sendable, HTTP2ConnectionOutbound: Sendable, HTTP2StreamInbound: Sendable, HTTP2StreamOutbound: Sendable> {
     /// The settings that will be used when establishing new streams. These mainly pertain to flow control.
     public var connection: NIOHTTP2Handler.ConnectionConfiguration
     /// The settings that will be used when establishing the connection. These will be sent to the peer as part of the
--- a/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
+++ b/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
@@ -274,7 +274,7 @@ extension Channel {
     ///
     /// - Parameters:
     ///   - mode: The mode this pipeline will operate in, server or client.
-    ///   - inboundStreamAsyncChannelConfiguration: Settings relating to `NIOAsyncChannel`s wrapping internal stream channels
+    ///   - inboundStreamAsyncChannelConfiguration: Settings relating to `NIOAsyncChannel`s wrapping internal stream channels.
     ///   - connectionConfiguration: The settings that will be used when establishing the connection. These will be sent to the peer as part of the
     ///         handshake.
     ///   - streamConfiguration: The settings that will be used when establishing new streams. These mainly pertain to flow control.
@@ -389,7 +389,7 @@ extension Channel {
     ///         channel has been fully mutated.
     /// - Returns: An `EventLoopFuture` containing a ``NIOTypedApplicationProtocolNegotiationHandler`` that completes when the channel
     ///     is ready to negotiate. This can then be used to access the ``NIOProtocolNegotiationResult`` which may itself
-    ///     be waited on to retrieve the result of the negotiation
+    ///     be waited on to retrieve the result of the negotiation.
     internal func configureHTTP2AsyncSecureUpgrade<HTTP1Output, HTTP2Output>(
         http1ConnectionInitializer: @escaping NIOChannelInitializerWithOutput<HTTP1Output>,
         http2ConnectionInitializer: @escaping NIOChannelInitializerWithOutput<HTTP2Output>
@@ -549,7 +549,7 @@ extension Channel {
     ///   - http2InboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
     /// - Returns: An `EventLoopFuture` containing a ``NIOTypedApplicationProtocolNegotiationHandler`` that completes when the channel
     ///     is ready to negotiate. This can then be used to access the ``NIOProtocolNegotiationResult`` which may itself
-    ///     be waited on to retrieve the result of the negotiation
+    ///     be waited on to retrieve the result of the negotiation.
     @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
     @_spi(AsyncChannel)
     public func configureAsyncHTTPServerPipeline<HTTP1ConnectionOutput, HTTP2ConnectionOutput, HTTP2StreamOutput>(
@@ -681,10 +681,7 @@ extension ChannelPipeline.SynchronousOperations {
     ///
     /// - Parameters:
     ///   - mode: The mode this pipeline will operate in, server or client.
-    ///   - connectionConfiguration: The settings that will be used when establishing the connection. These will be sent to the peer as part of the
-    ///         handshake.
-    ///   - streamConfiguration: The settings that will be used when establishing new streams. These mainly pertain to flow control.
-    ///   - streamDelegate: The delegate to be notified in the event of stream creation and close.
+    ///   - configuration: The settings that will be used when establishing the connection and new streams.
     ///   - position: The position in the pipeline into which to insert this handler.
     ///   - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
     /// - Returns: An `EventLoopFuture` containing the `AsyncStreamMultiplexer` inserted into this pipeline, which can
@@ -725,7 +722,7 @@ extension ChannelPipeline.SynchronousOperations {
     ///
     /// - Parameters:
     ///   - mode: The mode this pipeline will operate in, server or client.
-    ///   - inboundStreamAsyncChannelConfiguration: Settings relating to `NIOAsyncChannel`s wrapping internal stream channels
+    ///   - inboundStreamAsyncChannelConfiguration: Settings relating to `NIOAsyncChannel`s wrapping internal stream channels.
     ///   - connectionConfiguration: The settings that will be used when establishing the connection. These will be sent to the peer as part of the
     ///         handshake.
     ///   - streamConfiguration: The settings that will be used when establishing new streams. These mainly pertain to flow control.
@@ -806,6 +803,7 @@ public struct NIOHTTP2AsyncConfiguration<HTTP2ConnectionInbound: Sendable, HTTP2
 }
 
 #if swift(>=5.7)
+// Unchecked sendable because `NIOAsyncChannel.Configuration` is actually sendable but not yet marked as such
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
 @_spi(AsyncChannel)
 extension NIOHTTP2AsyncConfiguration: @unchecked Sendable {}
--- a/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
+++ b/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
@@ -434,29 +434,6 @@ extension Channel {
         }
     }
 
-    private func _commonHTTPServerPipeline(
-        configurator: @escaping (Channel) -> EventLoopFuture<Void>,
-        h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)?,
-        configureHTTP2Pipeline: @escaping (Channel) -> EventLoopFuture<Void>
-    ) -> EventLoopFuture<Void> {
-        let h2ChannelConfigurator = { (channel: Channel) -> EventLoopFuture<Void> in
-            configureHTTP2Pipeline(channel).flatMap { _ in
-                if let h2ConnectionChannelConfigurator = h2ConnectionChannelConfigurator {
-                    return h2ConnectionChannelConfigurator(channel)
-                } else {
-                    return channel.eventLoop.makeSucceededFuture(())
-                }
-            }
-        }
-        let http1ChannelConfigurator = { (channel: Channel) -> EventLoopFuture<Void> in
-            channel.pipeline.configureHTTPServerPipeline().flatMap { _ in
-                configurator(channel)
-            }
-        }
-        return self.configureHTTP2SecureUpgrade(h2ChannelConfigurator: h2ChannelConfigurator,
-                                                http1ChannelConfigurator: http1ChannelConfigurator)
-    }
-
     /// Configures a `ChannelPipeline` to speak either HTTP or HTTP/2 according to what can be negotiated with the client.
     ///
     /// This helper takes care of configuring the server pipeline such that it negotiates whether to
@@ -494,6 +471,29 @@ extension Channel {
                 }
             }.map { _ in () }
         }
+    }
+
+    private func _commonHTTPServerPipeline(
+        configurator: @escaping (Channel) -> EventLoopFuture<Void>,
+        h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)?,
+        configureHTTP2Pipeline: @escaping (Channel) -> EventLoopFuture<Void>
+    ) -> EventLoopFuture<Void> {
+        let h2ChannelConfigurator = { (channel: Channel) -> EventLoopFuture<Void> in
+            configureHTTP2Pipeline(channel).flatMap { _ in
+                if let h2ConnectionChannelConfigurator = h2ConnectionChannelConfigurator {
+                    return h2ConnectionChannelConfigurator(channel)
+                } else {
+                    return channel.eventLoop.makeSucceededFuture(())
+                }
+            }
+        }
+        let http1ChannelConfigurator = { (channel: Channel) -> EventLoopFuture<Void> in
+            channel.pipeline.configureHTTPServerPipeline().flatMap { _ in
+                configurator(channel)
+            }
+        }
+        return self.configureHTTP2SecureUpgrade(h2ChannelConfigurator: h2ChannelConfigurator,
+                                                http1ChannelConfigurator: http1ChannelConfigurator)
     }
 
     /// Configures a `ChannelPipeline` to speak HTTP/2 and wraps both the connection channel and any
--- a/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
+++ b/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
@@ -804,3 +804,9 @@ public struct NIOHTTP2AsyncConfiguration<HTTP2ConnectionInbound: Sendable, HTTP2
         self.inboundStreamInitializer = inboundStreamInitializer
     }
 }
+
+#if swift(>=5.7)
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+@_spi(AsyncChannel)
+extension NIOHTTP2AsyncConfiguration: @unchecked Sendable {}
+#endif
--- a/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
+++ b/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 @_spi(AsyncChannel) import NIOCore
-import NIOTLS
+@_spi(AsyncChannel) import NIOTLS
 
 /// The supported ALPN protocol tokens for NIO's HTTP/2 abstraction layer.
 ///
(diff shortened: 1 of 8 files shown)


Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -380,6 +380,63 @@ extension Channel {
         return self.pipeline.addHandler(alpnHandler)
     }
 
+
+    /// Configures a channel to perform a HTTP/2 secure upgrade with typed negotiation results.
```
FranzBusch, line 384: NIT
```suggestion
    /// Configures a channel to perform an HTTP/2 secure upgrade with typed negotiation results.
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -380,6 +380,63 @@ extension Channel {
...
+    ///         channel has been fully mutated.
+    ///   - http1ChannelConfigurator: A callback that will be invoked if HTTP/1.1 has been explicitly
+    ///         negotiated, or if no protocol was negotiated. Must return a future that completes when the
+    ///         channel has been fully mutated.
+    /// - Returns: An `EventLoopFuture` containing a ``NIOProtocolNegotiationResult`` that completes when the channel
+    ///     is ready to negotiate. The ``NIOProtocolNegotiationResult`` may itself be waited on to retrieve the result of
+    ///     the negotiation
+    @_spi(AsyncChannel)
+    public func configureHTTP2AsyncSecureUpgrade<HTTP1Output, HTTP2Output>(
+        h2ChannelConfigurator: @escaping (Channel) throws -> EventLoopFuture<HTTP2Output>,
```
FranzBusch, line 413: Can we call this `http2ChannelConfigurator` to align with the other parameter. I think these closures also need to be `@Sendable`.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -725,3 +839,10 @@ extension ChannelPipeline.SynchronousOperations {
         }
     }
 }
+
+/// `NegotiationResult` is a generic negotiation result holder for HTTP/1.1 and HTTP/2
+@_spi(AsyncChannel)
+public enum NegotiationResult<HTTP1Output, HTTP2Output> {
+    case http1_1(HTTP1Output)
+    case http2(HTTP2Output)
+}
```
FranzBusch, line 848: I think we need to be a bit more specific here. Especially if we want to have something that does H1, H2, H3 in the future. What about naming this `HTTP1And2NegotiationResult`? Also open for better names cc @Lukasa @glbrntt

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -460,40 +556,58 @@ extension Channel {
...
+    public func configureAsyncHTTPServerPipeline<HTTP1Output, HTTP2Output, HTTP2StreamOutput>(
         connectionConfiguration: NIOHTTP2Handler.ConnectionConfiguration,
         streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
         streamDelegate: NIOHTTP2StreamDelegate? = nil,
-        h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)? = nil,
-        configurator: @escaping NIOHTTP2Handler.StreamInitializer
-    ) -> EventLoopFuture<Void> {
-        return self._commonHTTPServerPipeline(configurator: configurator, h2ConnectionChannelConfigurator: h2ConnectionChannelConfigurator) { channel in
-            channel.configureHTTP2Pipeline(
+        h1ConnectionChannelConfigurator: @escaping NIOHTTP2Handler.ConnectionInitializerWithOutput<HTTP1Output>,
```
FranzBusch, line 584: This uses a type on the H2 handler which is a bit confusing.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -460,40 +556,58 @@ extension Channel {
...
+    ///         These will be sent to the peer as part of the handshake.
+    ///   - streamConfiguration: The settings that will be used when establishing new HTTP/2 streams.
+    ///         These mainly pertain to flow control.
     ///   - streamDelegate: The delegate to be notified in the event of stream creation and close.
+    ///   - h1ConnectionChannelConfigurator: An optional callback that will be invoked only
+    ///         when the negotiated protocol is HTTP/1.1 to configure the connection channel.
     ///   - h2ConnectionChannelConfigurator: An optional callback that will be invoked only
-    ///         when the negotiated protocol is H2 to configure the connection channel.
+    ///         when the negotiated protocol is HTTP/2 to configure the connection channel.
     ///   - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
```
FranzBusch, line 574: This doc comment doesn't match with the parameter names

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -460,40 +556,58 @@ extension Channel {
...
         streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
         streamDelegate: NIOHTTP2StreamDelegate? = nil,
-        h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)? = nil,
-        configurator: @escaping NIOHTTP2Handler.StreamInitializer
-    ) -> EventLoopFuture<Void> {
-        return self._commonHTTPServerPipeline(configurator: configurator, h2ConnectionChannelConfigurator: h2ConnectionChannelConfigurator) { channel in
-            channel.configureHTTP2Pipeline(
+        h1ConnectionChannelConfigurator: @escaping NIOHTTP2Handler.ConnectionInitializerWithOutput<HTTP1Output>,
+        h2ConnectionChannelConfigurator: @escaping NIOHTTP2Handler.ConnectionInitializerWithOutput<HTTP2Output>,
+        streamInitializer: @escaping NIOHTTP2Handler.StreamInitializerWithOutput<HTTP2StreamOutput>
```
FranzBusch, line 586: I think we should holistically prefix the parameters with either h1/2 or http1/http2 if they are for one specific protocol. I got quite confused trying to understand what parameter applies to what

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -460,40 +556,58 @@ extension Channel {
...
+    ) throws -> EventLoopFuture<NIOProtocolNegotiationResult<NegotiationResult<HTTP1Output, (HTTP2Output, NIOHTTP2Handler.AsyncStreamMultiplexer<HTTP2StreamOutput>)>>> {
```
FranzBusch, line 587: I am a bit surprised by this return type. I rather expected us to return the handler instead of the result here. The reason for this is that our bootstrap methods require the channel initialiser to return a `Handler` which we then use to await for the result.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: I _think_ they do but that the diff is presenting them in a confusing way

rnro: I restructured the original commit to split out the simple moving of a function to group like functions. Hopefully that makes the main diff easier to understand

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -434,6 +491,47 @@ extension Channel {
         }
     }
 
+    /// Configures a `ChannelPipeline` to speak either HTTP or HTTP/2 according to what can be negotiated with the client.
```
glbrntt, line 494:
```suggestion
    /// Configures a `ChannelPipeline` to speak either HTTP/1.1 or HTTP/2 according to what can be negotiated with the client.
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -434,6 +491,47 @@ extension Channel {
...
+    ///         handshake.
+    ///   - streamConfiguration: The settings that will be used when establishing new HTTP/2 streams. These mainly pertain to flow control.
+    ///   - streamDelegate: The delegate to be notified in the event of stream creation and close.
+    ///   - h2ConnectionChannelConfigurator: An optional callback that will be invoked only
+    ///         when the negotiated protocol is H2 to configure the connection channel.
+    ///   - configurator: A callback that will be invoked after a protocol has been negotiated.
+    ///         The callback only needs to add application-specific handlers and must return a future
+    ///         that completes when the channel has been fully mutated.
+    /// - Returns: `EventLoopFuture<Void>` that completes when the channel is ready.
+    public func configureCommonHTTPServerPipeline(
```
glbrntt, line 514: Why are we adding this? It doesn't seem related to the async work.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
glbrntt: Yup, agree, `streamInitializer` is not obviously http2

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: 🤦 we aren't - it's just that git is having difficulty with the diff. Before my change we had three functions with `func configureCommonHTTPServerPipeline` and one `func _common` which two of the three called. I re-ordered the functions so that the three similarly named functions were all together and followed by the common code. This apparently baffles git even when it's split out into its own commit.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: I've gone with `NIONegotiatedHTTPVersion`

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: I've created a pair of simplified global types which are used everywhere except for where the existing `NIOHTTP2Handler.StreamInitializer` was already used.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: I've changed this to match the bootstrap methods.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -380,7 +392,60 @@ extension Channel {
...
+    ///         negotiated, or if no protocol was negotiated. Must return a future that completes when the
+    ///         channel has been fully mutated.
+    ///   - http2ChannelConfigurator: A callback that will be invoked if HTTP/2 has been negogiated, and that
+    ///         should configure the channel for HTTP/2 use. Must return a future that completes when the
+    ///         channel has been fully mutated.
+    /// - Returns: An `EventLoopFuture` containing a ``NIOTypedApplicationProtocolNegotiationHandler`` that completes when the channel
+    ///     is ready to negotiate. This can then be used to access the ``NIOProtocolNegotiationResult`` which may itself
+    ///     be waited on to retrieve the result of the negotiation
+    @_spi(AsyncChannel)
+    public func configureHTTP2AsyncSecureUpgrade<HTTP1Output, HTTP2Output>(
```
FranzBusch, line 424: Are we intending to make this public in the end? I don't see a lot of usage for this without adding the protocol handlers in there so maybe we should make it non public in the end and right now add an `_` to the method name?

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -457,43 +563,62 @@ extension Channel {
...
+    ///   - http2StreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    /// - Returns: An `EventLoopFuture` containing a ``NIOTypedApplicationProtocolNegotiationHandler`` that completes when the channel
+    ///     is ready to negotiate. This can then be used to access the ``NIOProtocolNegotiationResult`` which may itself
+    ///     be waited on to retrieve the result of the negotiation
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    @_spi(AsyncChannel)
+    public func configureAsyncHTTPServerPipeline<HTTP1Output, HTTP2Output, HTTP2StreamOutput>(
         connectionConfiguration: NIOHTTP2Handler.ConnectionConfiguration,
         streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
         streamDelegate: NIOHTTP2StreamDelegate? = nil,
```
FranzBusch, line 593: IMO we should prefix those methods with `http2` as well. They only apply to it and right now you can only discover this by reading the docs.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -457,43 +563,62 @@ extension Channel {
...
+    ///         when the negotiated protocol is HTTP/1.1 to configure the connection channel.
+    ///   - http2ChannelConfigurator: An optional callback that will be invoked only
+    ///         when the negotiated protocol is HTTP/2 to configure the connection channel.
+    ///   - http2StreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    /// - Returns: An `EventLoopFuture` containing a ``NIOTypedApplicationProtocolNegotiationHandler`` that completes when the channel
+    ///     is ready to negotiate. This can then be used to access the ``NIOProtocolNegotiationResult`` which may itself
+    ///     be waited on to retrieve the result of the negotiation
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    @_spi(AsyncChannel)
+    public func configureAsyncHTTPServerPipeline<HTTP1Output, HTTP2Output, HTTP2StreamOutput>(
```
FranzBusch, line 590: What about naming the `HTTP2Ouput` generic type as `HTTP2ConnectionOutput` to better differentiate the two H2 types. (Maybe also add it for the H1 type

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -457,43 +563,62 @@ extension Channel {
...
         streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
         streamDelegate: NIOHTTP2StreamDelegate? = nil,
-        h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)? = nil,
-        configurator: @escaping NIOHTTP2Handler.StreamInitializer
-    ) -> EventLoopFuture<Void> {
-        return self._commonHTTPServerPipeline(configurator: configurator, h2ConnectionChannelConfigurator: h2ConnectionChannelConfigurator) { channel in
-            channel.configureHTTP2Pipeline(
+        http1ChannelConfigurator: @escaping NIOChannelInitializerWithOutput<HTTP1Output>,
+        http2ChannelConfigurator: @escaping NIOChannelInitializerWithOutput<HTTP2Output>,
+        http2StreamInitializer: @escaping NIOChannelInitializerWithOutput<HTTP2StreamOutput>
```
FranzBusch, line 596: Should we add `inbound` in this parameters name?

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -291,6 +316,201 @@ final class ConfiguringPipelineAsyncMultiplexerTests: XCTestCase {
...
+        var receivedParts: [HTTPServerRequestPart] = []
+
+        func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+            self.receivedParts.append(self.unwrapInboundIn(data))
+            context.fireChannelRead(data)
+        }
+    }
+
+    /// A simple channel handler that records inbound frames.
+    class HTTP1ClientResponseRecorderHandler: ChannelInboundHandler {
```
FranzBusch, line 504:
```suggestion
    final class HTTP1ServerRequestRecorderHandler: ChannelInboundHandler {
        typealias InboundIn = HTTPServerRequestPart

        var receivedParts: [HTTPServerRequestPart] = []

        func channelRead(context: ChannelHandlerContext, data: NIOAny) {
            self.receivedParts.append(self.unwrapInboundIn(data))
            context.fireChannelRead(data)
        }
    }

    /// A simple channel handler that records inbound frames.
    final class HTTP1ClientResponseRecorderHandler: ChannelInboundHandler {
```

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -291,6 +316,201 @@ final class ConfiguringPipelineAsyncMultiplexerTests: XCTestCase {
...
+            // Let's try sending some http/1.1 requests.
+            // we need to put these through a mapping to remove references to `IOData` which isn't Sendable
+            try await self.clientChannel.writeOutbound(HTTP1ClientSendability.RequestPart.head(ConfiguringPipelineAsyncMultiplexerTests.requestHead))
+            try await self.clientChannel.writeAndFlush(HTTP1ClientSendability.RequestPart.end(nil)).get()
+            try await self.interactInMemory(self.clientChannel, self.serverChannel)
+        }
+
+        // check expectations
+        let clientRecorder = try await self.clientChannel.pipeline.handler(type: HTTP1ClientResponseRecorderHandler.self).get()
+        let serverRecorder = try await self.serverChannel.pipeline.handler(type: HTTP1ServerRequestRecorderHandler.self).get()
```
FranzBusch, line 453: This is not safe. Those two types are not `Sendable`

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: Configuration has been largely refactored.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: Configuration has been largely refactored.

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
rnro: I've made these recorder types `Sendable`

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -434,6 +468,47 @@ extension Channel {
...
+    ///   - configurator: A callback that will be invoked after a protocol has been negotiated.
+    ///         The callback only needs to add application-specific handlers and must return a future
+    ///         that completes when the channel has been fully mutated.
+    /// - Returns: `EventLoopFuture<Void>` that completes when the channel is ready.
+    public func configureCommonHTTPServerPipeline(
+        connectionConfiguration: NIOHTTP2Handler.ConnectionConfiguration,
+        streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
+        streamDelegate: NIOHTTP2StreamDelegate? = nil,
+        h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)? = nil,
+        configurator: @escaping NIOChannelInitializer
```
glbrntt, line 496: 's/configurator/initializer'

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -434,6 +468,47 @@ extension Channel {
...
+    ///         when the negotiated protocol is H2 to configure the connection channel.
+    ///   - configurator: A callback that will be invoked after a protocol has been negotiated.
+    ///         The callback only needs to add application-specific handlers and must return a future
+    ///         that completes when the channel has been fully mutated.
+    /// - Returns: `EventLoopFuture<Void>` that completes when the channel is ready.
+    public func configureCommonHTTPServerPipeline(
+        connectionConfiguration: NIOHTTP2Handler.ConnectionConfiguration,
+        streamConfiguration: NIOHTTP2Handler.StreamConfiguration,
+        streamDelegate: NIOHTTP2StreamDelegate? = nil,
+        h2ConnectionChannelConfigurator: ((Channel) -> EventLoopFuture<Void>)? = nil,
```
glbrntt, line 495: can this be `NIOChannelInitializer? = nil`

glbrntt: also 's/h2/http2'

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -457,43 +532,55 @@ extension Channel {
...
+    ///     is HTTP/1.1 to configure the connection channel.
+    ///   - http2ConnectionInitializer: An optional callback that will be invoked only when the negotiated protocol
+    ///     is HTTP/2 to configure the connection channel.
+    ///   - http2InboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    /// - Returns: An `EventLoopFuture` containing a ``NIOTypedApplicationProtocolNegotiationHandler`` that completes when the channel
+    ///     is ready to negotiate. This can then be used to access the ``NIOProtocolNegotiationResult`` which may itself
+    ///     be waited on to retrieve the result of the negotiation
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    @_spi(AsyncChannel)
+    public func configureAsyncHTTPServerPipeline<HTTP1ConnectionOutput, HTTP2ConnectionOutput, HTTP2StreamOutput>(
```
glbrntt, line 555: I think the existing version of this has a notion of "upgrade" in its name

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -673,55 +725,112 @@ extension ChannelPipeline.SynchronousOperations {
     /// 
     /// Using this rather than implementing a similar function yourself allows that pipeline to evolve without breaking your code.
     ///
+
```
glbrntt, line 728:
```suggestion
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -673,55 +725,112 @@ extension ChannelPipeline.SynchronousOperations {
...
+                    isOutboundHalfClosureEnabled: inboundStreamAsyncChannelConfiguration.isOutboundHalfClosureEnabled
                 )
             }
         }
     }
 }
+
+/// `NIONegotiatedHTTPVersion` is a generic negotiation result holder for HTTP/1.1 and HTTP/2
+@_spi(AsyncChannel)
+public enum NIONegotiatedHTTPVersion<HTTP1Output, HTTP2Output> {
```
glbrntt, line 770: I think we require this:

```suggestion
public enum NIONegotiatedHTTPVersion<HTTP1Output: Sendable, HTTP2Output: Sendable>: Sendable {
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -673,55 +725,112 @@ extension ChannelPipeline.SynchronousOperations {
...
+    case http1_1(HTTP1Output)
+    case http2(HTTP2Output)
+}
+
+/// `HTTP2AsyncConfiguration` contains all configuration required for setting up an HTTP/2 async connection.
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+@_spi(AsyncChannel)
+public struct HTTP2AsyncConfiguration<HTTP2ConnectionInbound, HTTP2ConnectionOutbound, HTTP2StreamInbound, HTTP2StreamOutbound> {
```
glbrntt, line 778:
```suggestion
public struct HTTP2AsyncConfiguration<HTTP2ConnectionInbound, HTTP2ConnectionOutbound, HTTP2StreamInbound, HTTP2StreamOutbound>: Sendable {
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -673,55 +725,112 @@ extension ChannelPipeline.SynchronousOperations {
...
+        let inboundType: Inbound.Type
+        let outboundType: Outbound.Type
+        let backpressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark?
+        let isOutboundHalfClosureEnabled: Bool
+
+        public init(
+            inboundType: Inbound.Type = Inbound.self,
+            outboundType: Outbound.Type = Outbound.self,
+            backpressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark? = nil,
+            isOutboundHalfClosureEnabled: Bool? = nil
```
glbrntt, line 827: I think these were added in a different order in NIO: `inboundType` and `outboundType` came at the end

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -57,6 +60,23 @@ final class ConfiguringPipelineAsyncMultiplexerTests: XCTestCase {
         }
     }
 
+    final class HTTP1OKResponder: ChannelInboundHandler {
+        typealias InboundIn = HTTPServerRequestPart
+        typealias OutboundOut = HTTPServerResponsePart
+
+        func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+            guard case .end = self.unwrapInboundIn(data) else {
```
glbrntt, line 68: nit: prefer switching over all cases

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -218,77 +231,222 @@ final class ConfiguringPipelineAsyncMultiplexerTests: XCTestCase {
...
+            context.write(self.wrapOutboundOut(httpServerResponsePart), promise: promise)
+        }
+    }
+
+    /// A simple channel handler that records inbound messages.
+    internal final class InboundRecorderHandler<message>: ChannelInboundHandler, @unchecked Sendable {
+        typealias InboundIn = message
+
+        private let partsLock = NIOLock()
+        var _receivedParts: [message] = []
```
glbrntt, line 432:
```suggestion
        private var _receivedParts: [message] = []
```

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -64,6 +64,8 @@ extension XCTestCase {
 
     /// Have two `NIOAsyncTestingChannel` objects send and receive data from each other until
     /// they make no forward progress.
+    ///
+    /// ** This function is not thread safe and can lead to deadlocks, prefer the one-way variant which is less error-prone**
```
glbrntt, line 68: It _is_ thread-safe, the problem is that calling it from multiple threads is racy.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: The method which mentions "upgrade" is one layer down, the analogue of this is `configureCommonHTTPServerPipeline` - but we don't have any common configuration because we don't translate H2 into H1 so that's not relevant.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: This code is existing API which has been moved so I can't rename it 😭

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: This code is existing API which has been moved so I can't rename it 😭

glbrntt:
oh, looks like you regressed allocations @rnro

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -675,53 +727,109 @@ extension ChannelPipeline.SynchronousOperations {
...
+@_spi(AsyncChannel)
+public enum NIONegotiatedHTTPVersion<HTTP1Output: Sendable, HTTP2Output: Sendable> {
+    case http1_1(HTTP1Output)
+    case http2(HTTP2Output)
+}
+
+/// `HTTP2AsyncConfiguration` contains all configuration required for setting up an HTTP/2 async connection.
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+@_spi(AsyncChannel)
+public struct HTTP2AsyncConfiguration<HTTP2ConnectionInbound: Sendable, HTTP2ConnectionOutbound: Sendable, HTTP2StreamInbound: Sendable, HTTP2StreamOutbound: Sendable>: Sendable {
```
FranzBusch, line 777:
```suggestion
public struct NIOHTTP2AsyncConfiguration<HTTP2ConnectionInbound: Sendable, HTTP2ConnectionOutbound: Sendable, HTTP2StreamInbound: Sendable, HTTP2StreamOutbound: Sendable>: Sendable {
```

rnro:

rnro:
Rebased on top of https://github.com/apple/swift-nio-http2/pull/409

rnro:
All of the API breakage is in code added since the last release. Almost all of it is `Async...` configuration functions, there are also some closure type aliases which they used but we've simplified and done away with.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -677,55 +725,88 @@ extension ChannelPipeline.SynchronousOperations {
...
+        self.stream = stream
+        self.connectionInitializer = connectionInitializer
+        self.inboundStreamInitializer = inboundStreamInitializer
+    }
+}
+
+#if swift(>=5.7)
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+@_spi(AsyncChannel)
+extension NIOHTTP2AsyncConfiguration: @unchecked Sendable {}
```
glbrntt, line 811: nit: add a comment as to why this is `@unchecked`

glbrntt (approved):

PR #411, opened 2023-07-21
Adopt latest AsyncChannel changes

FranzBusch:
# Motivation
In my latest PR in NIO, I removed a bunch of API from the bootstrap methods to make the API easier to digest and reduce a number of duplicated code issues.

# Modification
This PR adopts the latest NIO changes and adapts the APIs in H2. More specifically, it removes the `NIOAsyncChannel` convenience methods and makes sure that everything aligns nicely with the proposes bootstrap APIs.

# Result
We now have the APIs in place to bootstrap a protocol negotiated H1/H2 server.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,7 @@ let package = Package(
         .library(name: "NIOHTTP2", targets: ["NIOHTTP2"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.56.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.57.0"),
         .package(url: "https://github.com/apple/swift-atomics.git", from: "1.0.2"),
         .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
     ],
--- a/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
+++ b/Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
@@ -220,103 +220,6 @@ extension Channel {
         }
     }
 
-    /// Configures a `ChannelPipeline` to speak HTTP/2 and sets up mapping functions so that it may be interacted with from concurrent code.
-    ///
-    /// In general this is not entirely useful by itself, as HTTP/2 is a negotiated protocol. This helper does not handle negotiation.
-    /// Instead, this simply adds the handler required to speak HTTP/2 after negotiation has completed, or when agreed by prior knowledge.
-    /// Use this function to setup a HTTP/2 pipeline if you wish to use async sequence abstractions over inbound and outbound streams.
-    /// Using this rather than implementing a similar function yourself allows that pipeline to evolve without breaking your code.
-    ///
-    /// - Parameters:
-    ///   - mode: The mode this pipeline will operate in, server or client.
-    ///   - configuration: The settings that will be used when establishing the connection and new streams.
-    ///   - position: The position in the pipeline into which to insert this handler.
-    ///   - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
-    /// - Returns: An `EventLoopFuture` containing the `AsyncStreamMultiplexer` inserted into this pipeline, which can
-    ///     be used to initiate new streams and iterate over inbound HTTP/2 stream channels.
-    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
-    @_spi(AsyncChannel)
-    public func configureAsyncHTTP2Pipeline<Output>(
-        mode: NIOHTTP2Handler.ParserMode,
-        configuration: NIOHTTP2Handler.Configuration = .init(),
-        position: ChannelPipeline.Position = .last,
-        inboundStreamInitializer: @escaping NIOChannelInitializerWithOutput<Output>
-    ) -> EventLoopFuture<NIOHTTP2Handler.AsyncStreamMultiplexer<Output>> {
-        if self.eventLoop.inEventLoop {
-            return self.eventLoop.makeCompletedFuture {
-                return try self.pipeline.syncOperations.configureAsyncHTTP2Pipeline(
-                    mode: mode,
-                    configuration: configuration,
-                    position: position,
-                    inboundStreamInitializer: inboundStreamInitializer
-                )
-            }
-        } else {
-            return self.eventLoop.submit {
-                return try self.pipeline.syncOperations.configureAsyncHTTP2Pipeline(
-                    mode: mode,
-                    configuration: configuration,
-                    position: position,
-                    inboundStreamInitializer: inboundStreamInitializer
-                )
-            }
-        }
-    }
-
-    /// Configures a `ChannelPipeline` to speak HTTP/2 and wraps any created inbound stream channels in `NIOAsyncChannel`s.
-    ///
-    /// In general this is not entirely useful by itself, as HTTP/2 is a negotiated protocol. This helper does not handle negotiation.
-    /// Instead, this simply adds the handler required to speak HTTP/2 after negotiation has completed, or when agreed by prior knowledge.
-    /// Use this function to setup a HTTP/2 pipeline if you wish to use `NIOAsyncChannel`s to provide async sequence abstractions
-    /// over inbound and outbound streams whilst handling back-pressure.
-    ///
-    /// Using this rather than implementing a similar function yourself allows that pipeline to evolve without breaking your code.
-    ///
-    /// - Parameters:
-    ///   - mode: The mode this pipeline will operate in, server or client.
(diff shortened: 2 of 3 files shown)


Package.swift
```diff
@@ -21,7 +21,7 @@ let package = Package(
         .library(name: "NIOHTTP2", targets: ["NIOHTTP2"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.56.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", branch: "main"),
```
FranzBusch, line 24: Will change this once our next release is out

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -531,6 +382,146 @@ extension Channel {
...
+    ///   - http2ConnectionInitializer: A callback that will be invoked if HTTP/2 has been negotiated, and that
+    ///         should configure the channel for HTTP/2 use. Must return a future that completes when the
+    ///         channel has been fully mutated.
+    /// - Returns: An `EventLoopFuture` containing a `NIOTypedApplicationProtocolNegotiationHandler` that completes when the channel
+    ///     is ready to negotiate. This can then be used to access the `NIOProtocolNegotiationResult` which may itself
+    ///     be waited on to retrieve the result of the negotiation.
+    internal func configureHTTP2AsyncSecureUpgrade<HTTP1Output, HTTP2Output>(
+        http1ConnectionInitializer: @escaping NIOChannelInitializerWithOutput<HTTP1Output>,
+        http2ConnectionInitializer: @escaping NIOChannelInitializerWithOutput<HTTP2Output>
+    ) -> EventLoopFuture<EventLoopFuture<NIOProtocolNegotiationResult<NIONegotiatedHTTPVersion<HTTP1Output, HTTP2Output>>>> {
```
glbrntt, line 503: Oof, `ELF<ELF<...>>` -- can we remove a level here?

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -531,6 +382,146 @@ extension Channel {
...
+    /// - Parameters:
+    ///   - mode: The mode this pipeline will operate in, server or client.
+    ///   - connectionConfiguration: The settings that will be used when establishing the connection. These will be sent to the peer as part of the
+    ///         handshake.
+    ///   - streamConfiguration: The settings that will be used when establishing new streams. These mainly pertain to flow control.
+    ///   - streamDelegate: The delegate to be notified in the event of stream creation and close.
+    ///   - position: The position in the pipeline into which to insert this handler.
+    ///   - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    /// - Returns: The `StreamMultiplexer` inserted into this pipeline, which can be used to initiate new streams.
+    public func configureHTTP2Pipeline(mode: NIOHTTP2Handler.ParserMode,
```
glbrntt, line 405: This isn't actually new is it? The diff is just confusing here I think...

glbrntt (changes requested):

Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
```diff
@@ -257,21 +254,16 @@ final class ConfiguringPipelineAsyncMultiplexerTests: XCTestCase {
...
+        let negotiationResult = try await nioProtocolNegotiationResult.get().waitForFinalResult()
 
         try await assertNoThrow(try await self.assertDoHandshake(client: self.clientChannel, server: self.serverChannel))
 
         let serverMultiplexer: NIOHTTP2Handler.AsyncStreamMultiplexer<Channel>
-        switch nioProtocolNegotiationResult {
-        case .deferredResult:
+        switch negotiationResult {
+        case .http1_1:
             preconditionFailure("Negotiation result must be ready")
```
rnro, line 264:
```suggestion
            preconditionFailure("Negotiation result must be HTTP/2")
```
I was going to fix this in a future PR but we may as well do it now

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
FranzBusch: Yeah I moved moved arounds so that all async methods are now in one place in the file. I confused me a lot that they were scattered all around.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -531,6 +382,146 @@ extension Channel {
...
+    /// This configuration is acceptable for use on both client and server channel pipelines.
+    ///
+    /// - Parameters:
+    ///   - http1ConnectionInitializer: A callback that will be invoked if HTTP/1.1 has been explicitly
+    ///         negotiated, or if no protocol was negotiated. Must return a future that completes when the
+    ///         channel has been fully mutated.
+    ///   - http2ConnectionInitializer: A callback that will be invoked if HTTP/2 has been negotiated, and that
+    ///         should configure the channel for HTTP/2 use. Must return a future that completes when the
+    ///         channel has been fully mutated.
+    /// - Returns: An `EventLoopFuture` containing a `NIOTypedApplicationProtocolNegotiationHandler` that completes when the channel
```
rnro, line 497: This is no longer true, we no longer return the handler.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -582,94 +573,9 @@ extension Channel {
...
-    /// - Parameters:
-    ///   - mode: The mode this pipeline will operate in, server or client.
-    ///   - configuration: The settings and initialization configuration for both connections and streams, and their wrapping
-    ///     ``NIOAsyncChannel``s.
-    /// - Returns: An `EventLoopFuture` containing the `AsyncStreamMultiplexer` inserted into this pipeline which wraps
-    ///     inbound streams as `NIOAsyncChannels` after initialization. The multiplexer can be used to initiate new streams
-    ///     and iterate over inbound HTTP/2 stream channels.
-    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
-    @_spi(AsyncChannel)
-    public func configureAsyncHTTP2Pipeline<ConnectionInbound, ConnectionOutbound, StreamInbound, StreamOutbound>(
```
rnro, line 605: Why are we removing this variant with this change? It's not related to protocol negotiation.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
FranzBusch: I know awkward but we really can't. The `configureHTTP2AsyncSecureUpgrade` is called in the `childChannelInitializer` of a `ServerBootstrap` for example. We have to return an `ELF` from the `childChannelInitializer` which is then awaited for the result. If we would not nest the ELFs here we would wait for the protocol negotiation result in the `childChannelInitializer` but that is not really what we want. The result of the channel initialiser is an ELF<NIOProtocolNegotiationResult>

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -220,103 +220,6 @@ extension Channel {
...
-                    mode: mode,
-                    configuration: configuration,
-                    position: position,
-                    inboundStreamInitializer: inboundStreamInitializer
-                )
-            }
-        }
-    }
-
-    /// Configures a `ChannelPipeline` to speak HTTP/2 and wraps any created inbound stream channels in `NIOAsyncChannel`s.
```
rnro, line 266: Why are we removing this variant with this change? It's not related to protocol negotiation.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -708,54 +614,6 @@ extension ChannelPipeline.SynchronousOperations {
 
         return try handler.syncAsyncStreamMultiplexer(continuation: continuation, inboundStreamChannels: inboundStreamChannels)
     }
-
-    /// Configures a `ChannelPipeline` to speak HTTP/2 and wraps any created inbound stream channels in `NIOAsyncChannel`s.
```
rnro, line 712: Why are we removing this variant with this change? It's not related to protocol negotiation.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
FranzBusch: It is not removed. Just moved if further down.

FranzBusch: This was intentional. While they are not protocol negotiation specific they were conveniences for `NIOAsyncChannel`. For now we have taken all of those conveniences out from `NIO` as well since people can just wrap in their initialisers themselves and it avoids a bunch of APIs for us.

rnro (approved):

glbrntt (approved):

PR #413, opened 2023-07-26
Fix parent channel `read()` call on `HTTP2StreamChannel` initialization causing incorrect order of inbound `HTTP2Frame`s

### LINKED ISSUE #410: Race condition causes header frame to be fired on HTTP2StreamChannel AFTER first data frame(s)
Hi, I have a pretty weird bug going on, not sure if I'm doing something wrong but I thought I'd share this. I'm on tag 1.17.0. It's fairly easy for me to reproduce but I have quite a lot of custom code involved that I cannot easily share. I'll do my best though to describe what I think is happening.

So whenever we get a new decoded header frame in `HTTP2CommonInboundStreamMultiplexer::receivedFrame(_:, context:, multiplexer:)`
https://github.com/apple/swift-nio-http2/blob/044339de7bdffddb81cb77fc57fb77ff2e743088/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift#L92 
we create a new stream channel and trigger its initializer:
https://github.com/apple/swift-nio-http2/blob/044339de7bdffddb81cb77fc57fb77ff2e743088/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift#L112
Then, we fire the header we got through the pipeline of this new channel as a first frame
https://github.com/apple/swift-nio-http2/blob/044339de7bdffddb81cb77fc57fb77ff2e743088/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift#L113

The **problem** seems to be, that within the initialization of the new stream channel, eventually `HTTP2StreamChannel::performActivation()` will be called
https://github.com/apple/swift-nio-http2/blob/044339de7bdffddb81cb77fc57fb77ff2e743088/Sources/NIOHTTP2/HTTP2StreamChannel.swift#L305
then
https://github.com/apple/swift-nio-http2/blob/044339de7bdffddb81cb77fc57fb77ff2e743088/Sources/NIOHTTP2/HTTP2StreamChannel.swift#L707
which will trigger another read on the pipeline:
https://github.com/apple/swift-nio-http2/blob/044339de7bdffddb81cb77fc57fb77ff2e743088/Sources/NIOHTTP2/HTTP2StreamChannel.swift#L712
That in turn might cause another frame to be processed in 
https://github.com/apple/swift-nio-http2/blob/044339de7bdffddb81cb77fc57fb77ff2e743088/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift#L70
recursively and fire that second frame through the pipeline of the now existent (in `HTTP2CommonInboundStreamMultiplexer::streams`) stream channel! The first (header) frame payload is still to be delivered since  `HTTP2CommonInboundStreamMultiplexer::receivedFrame(...)` is still on the call stack, waiting for the initializer of the new stream channel to return here:
https://github.com/apple/swift-nio-http2/blob/044339de7bdffddb81cb77fc57fb77ff2e743088/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift#L112

No idea why this bug wouldn't have appeared for anyone else earlier. Seems pretty fundamental. What could I be doing wrong here? I'm using a custom pipeline underneath the HTTP2 handler which implements a proxy connection / virtual socket.

Wrapping `self.tryToRead()` and `self.deliverPendingWrites()` in `eventLoop.execute()` like so
```
self.eventLoop.execute {
  self.tryToAutoRead()
  self.deliverPendingWrites()
}
```
in https://github.com/apple/swift-nio-http2/blob/044339de7bdffddb81cb77fc57fb77ff2e743088/Sources/NIOHTTP2/HTTP2StreamChannel.swift#L305
so that new frames are only read after the header has finished processing seems to solve the issue for me.

qusc:
Motivation:

Make sure the order of `HTTP2Frame`s that are fired through the pipeline of a `HTTP2StreamChannel` by `HTTP2CommonInboundStreamMultiplexer` is correct when a `read()` call on the parent channel synchronously causes further frames to be processed.

Modifications:

Reorder calls in `HTTP2CommonInboundStreamMultiplexer.receivedFrame(frame:context:multiplexer)` so that initial header frame is buffered and processed first.

Result:

Resolves #410 and makes newly added test case `HTTP2StreamMultiplexerTests.testMultiplexerFiresInitialFramesInCorrectOrder()` pass.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
@@ -109,8 +109,11 @@ extension HTTP2CommonInboundStreamMultiplexer {
             // does no actual work.
             self.streamChannelContinuation?.yield(channel: channel.baseChannel)
 
-            channel.configureInboundStream(initializer: self.inboundStreamStateInitializer)
+            // Note: Firing the initial (header) frame before calling `HTTP2StreamChannel.configureInboundStream(initializer:)`
+            // is crucial to preserve frame order, since the initialization process might trigger another read on the parent
+            // channel which in turn might cause further frames to be processed synchronously.
             channel.receiveInboundFrame(frame)
+            channel.configureInboundStream(initializer: self.inboundStreamStateInitializer)
 
             if !channel.inList {
                 self.didReadChannels.append(channel)
--- a/Tests/NIOHTTP2Tests/HTTP2StreamMultiplexerTests.swift
+++ b/Tests/NIOHTTP2Tests/HTTP2StreamMultiplexerTests.swift
@@ -1927,4 +1927,50 @@ final class HTTP2StreamMultiplexerTests: XCTestCase {
         XCTAssertNoThrow(try channel.finish(acceptAlreadyClosed: false))
     }
 
+    func testMultiplexerFiresInitialFramesInCorrectOrder() throws {
+        final class BufferingChannelHandler<Element>: ChannelDuplexHandler {
+            typealias InboundIn = Element
+            typealias InboundOut = Element
+            typealias OutboundIn = Element
+            typealias OutboundOut = Element
+            
+            let elementsToBufferCount: Int = 1
+            var bufferedElements: CircularBuffer<Element> = []
+
+            func read(context: ChannelHandlerContext) {
+                while let bufferedElement = bufferedElements.popFirst() {
+                    context.fireChannelRead(wrapInboundOut(bufferedElement))
+                }
+                
+                context.read()
+            }
+            
+            func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+                bufferedElements.append(unwrapInboundIn(data))
+                
+                if bufferedElements.count > elementsToBufferCount {
+                    context.fireChannelRead(wrapInboundOut(bufferedElements.removeFirst()))
+                }
+            }
+        }
+        
+        XCTAssertNoThrow(try self.channel.connect(to: SocketAddress(unixDomainSocketPath: "/whatever"), promise: nil))
+        let streamID = HTTP2StreamID(1)
+        let headerFrame = HTTP2Frame(streamID: streamID, payload: .headers(.init(headers: HPACKHeaders())))
+        let payloadBuffer = channel.allocator.buffer(capacity: 0)
+        let dataFrame = HTTP2Frame(streamID: streamID, payload: .data(.init(data: .byteBuffer(payloadBuffer))))
+        
+        let multiplexer = HTTP2StreamMultiplexer(mode: .server, channel: self.channel) { channel in
+            channel.pipeline.addHandler(FramePayloadExpecter(expectedPayload: [headerFrame.payload, dataFrame.payload]))
+        }
+        
+        XCTAssertNoThrow(try self.channel.pipeline.addHandler(BufferingChannelHandler<HTTP2Frame>()).wait())
+        XCTAssertNoThrow(try self.channel.pipeline.addHandler(multiplexer).wait())
+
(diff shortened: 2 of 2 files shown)


Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -109,8 +109,8 @@ extension HTTP2CommonInboundStreamMultiplexer {
             // does no actual work.
             self.streamChannelContinuation?.yield(channel: channel.baseChannel)
 
-            channel.configureInboundStream(initializer: self.inboundStreamStateInitializer)
             channel.receiveInboundFrame(frame)
+            channel.configureInboundStream(initializer: self.inboundStreamStateInitializer)
```
FranzBusch, line 113: This is quite subtle. Can we add a comment here that this order is important.

qusc: Absolutely; done

Tests/NIOHTTP2Tests/HTTP2StreamMultiplexerTests.swift
```diff
@@ -1927,4 +1927,51 @@ final class HTTP2StreamMultiplexerTests: XCTestCase {
         XCTAssertNoThrow(try channel.finish(acceptAlreadyClosed: false))
     }
 
+    @available(*, deprecated, message: "Deprecated so deprecated functionality can be tested without warnings")
```
Lukasa, line 1930: Why do we need this to be deprecated? What deprecated function are we using?

Lukasa (changes requested):

Lukasa:

Tests/NIOHTTP2Tests/HTTP2StreamMultiplexerTests.swift
qusc: Not using any deprecated function to my knowledge; I added this since every other test in this file has it and I thought it's used to avoid warnings caused by potential future deprecations. Will remove it

Lukasa:

Lukasa:
Locally your test crashes for me: 

```
Test Case '-[NIOHTTP2Tests.HTTP2StreamMultiplexerTests testMultiplexerFiresInitialFramesInCorrectOrder]' started.
NIOCore/NIOAny.swift:200: Fatal error: tried to decode as type HTTP2Frame but found FramePayload with contents other(NIOHTTP2.HTTP2Frame.FramePayload.headers(NIOHTTP2.HTTP2Frame.FramePayload.Headers(headers: [], priorityData: nil, endStream: false, _paddingBytes: nil)))
```

qusc:
> Locally your test crashes for me:
> 
> ```
> Test Case '-[NIOHTTP2Tests.HTTP2StreamMultiplexerTests testMultiplexerFiresInitialFramesInCorrectOrder]' started.
> NIOCore/NIOAny.swift:200: Fatal error: tried to decode as type HTTP2Frame but found FramePayload with contents other(NIOHTTP2.HTTP2Frame.FramePayload.headers(NIOHTTP2.HTTP2Frame.FramePayload.Headers(headers: [], priorityData: nil, endStream: false, _paddingBytes: nil)))
> ```

@Lukasa looking into this...

qusc:
@Lukasa Apparently caused by the switch to the non-deprecated `init` of `HTTP2StreamMultiplexer`. That implicitly switches the elements being sent down the pipeline to `HTTP2Frame.Payload` to exclude the stream id? But the `InboundOut` type of `HTTP2StreamMultiplexer` is technically still `HTTP2Frame`?

qusc:
@swift-server-bot test this please if I'm allowed to trigger that

Lukasa:

Lukasa (approved):

PR #415, opened 2023-08-11
Async stream initializers order matches existing

rnro:
## Motivation:

When adding in the SPI methods for exposing async sequences of HTTP/2 streams we moved the stream initialization to a subtly different location so that it was easier to exfiltrate the outputs of those initialization functions (such as protocol negotiation outputs).

In some cases this broke ordering expectations.

## Modifications:

Yield the (optionally wrapped) channels as initializer outputs to the async stream when the initialization and activation steps succeed.

## Result:

Changes only exist within SPI. Async inbound stream channel initialization now matches previous behavior.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
@@ -23,20 +23,7 @@ internal struct InlineStreamMultiplexer {
     /// The delegate to be notified upon stream creation and close.
     private var streamDelegate: NIOHTTP2StreamDelegate?
 
-    internal var streamInitializerProductContinuation: (any AnyContinuation)? {
-        self.commonStreamMultiplexer.streamInitializerProductContinuation
-    }
-
-    init(
-        context: ChannelHandlerContext,
-        outboundView: NIOHTTP2Handler.OutboundView,
-        mode: NIOHTTP2Handler.ParserMode,
-        inboundStreamStateInitializer: MultiplexerAbstractChannel.InboundStreamStateInitializer,
-        targetWindowSize: Int, streamChannelOutboundBytesHighWatermark: Int,
-        streamChannelOutboundBytesLowWatermark: Int,
-        streamDelegate: NIOHTTP2StreamDelegate?,
-        streamInitializerProductContinuation: (any AnyContinuation)?
-    ) {
+    init(context: ChannelHandlerContext, outboundView: NIOHTTP2Handler.OutboundView, mode: NIOHTTP2Handler.ParserMode, inboundStreamStateInitializer: MultiplexerAbstractChannel.InboundStreamStateInitializer, targetWindowSize: Int, streamChannelOutboundBytesHighWatermark: Int, streamChannelOutboundBytesLowWatermark: Int, streamDelegate: NIOHTTP2StreamDelegate?) {
         self.context = context
         self.commonStreamMultiplexer = HTTP2CommonInboundStreamMultiplexer(
             mode: mode,
@@ -44,8 +31,7 @@ internal struct InlineStreamMultiplexer {
             inboundStreamStateInitializer: inboundStreamStateInitializer,
             targetWindowSize: targetWindowSize,
             streamChannelOutboundBytesHighWatermark: streamChannelOutboundBytesHighWatermark,
-            streamChannelOutboundBytesLowWatermark: streamChannelOutboundBytesLowWatermark,
-            streamInitializerProductContinuation: streamInitializerProductContinuation
+            streamChannelOutboundBytesLowWatermark: streamChannelOutboundBytesLowWatermark
         )
         self.outboundView = outboundView
         self.streamDelegate = streamDelegate
@@ -220,6 +206,12 @@ extension NIOHTTP2Handler {
     }
 }
 
+extension InlineStreamMultiplexer {
+    func setChannelContinuation(_ streamChannels: any ChannelContinuation) {
+        self.commonStreamMultiplexer.setChannelContinuation(streamChannels)
+    }
+}
+
 extension NIOHTTP2Handler {
     /// A variant of `NIOHTTP2Handler.StreamMultiplexer` which creates a child channel for each HTTP/2 stream and
     /// provides access to inbound HTTP/2 streams.
@@ -240,8 +232,9 @@ extension NIOHTTP2Handler {
         public let inbound: NIOHTTP2InboundStreamChannels<InboundStreamOutput>
 
         // Cannot be created by users.
-        internal init(_ inlineStreamMultiplexer: InlineStreamMultiplexer, inboundStreamChannels: NIOHTTP2InboundStreamChannels<InboundStreamOutput>) {
+        internal init(_ inlineStreamMultiplexer: InlineStreamMultiplexer, continuation: any ChannelContinuation, inboundStreamChannels: NIOHTTP2InboundStreamChannels<InboundStreamOutput>) {
             self.inlineStreamMultiplexer = inlineStreamMultiplexer
+            self.inlineStreamMultiplexer.setChannelContinuation(continuation)
             self.inbound = inboundStreamChannels
         }
 
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
@@ -242,27 +242,5 @@ extension NIOHTTP2Handler {
         public func createStreamChannel<OutboundStreamOutput>(_ initializer: @escaping NIOChannelInitializerWithOutput<OutboundStreamOutput>) async throws -> OutboundStreamOutput {
             return try await self.inlineStreamMultiplexer.createStreamChannel(initializer).get()
         }
-
-        /// Create a stream channel initialized with the provided closure and return it wrapped within a `NIOAsyncChannel`.
-        ///
-        /// - Parameters:
-        ///   - configuration: Configuration for the ``NIOAsyncChannel`` wrapping the HTTP/2 stream channel.
-        ///   - initializer: A callback that will be invoked to allow you to configure the
-        ///         `ChannelPipeline` for the newly created channel.
-        @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
-        @_spi(AsyncChannel)
-        public func createStreamChannel<Inbound, Outbound>(
-            asyncChannelConfiguration: NIOAsyncChannel<Inbound, Outbound>.Configuration = .init(),
-            initializer: @escaping NIOChannelInitializer
-        ) async throws -> NIOAsyncChannel<Inbound, Outbound> {
-            return try await self.createStreamChannel { channel in
-                initializer(channel).flatMapThrowing { _ in
-                    return try NIOAsyncChannel(
-                        synchronouslyWrapping: channel,
-                        configuration: asyncChannelConfiguration
-                    )
-                }
-            }
-        }
     }
 }
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
@@ -140,11 +140,11 @@ extension InlineStreamMultiplexer {
 }
 
 extension InlineStreamMultiplexer {
-    internal func createStreamChannel(promise: EventLoopPromise<Channel>?, _ streamStateInitializer: @escaping (Channel) -> EventLoopFuture<Void>) {
+    internal func createStreamChannel(promise: EventLoopPromise<Channel>?, _ streamStateInitializer: @escaping NIOChannelInitializer) {
         self.commonStreamMultiplexer.createStreamChannel(multiplexer: .inline(self), promise: promise, streamStateInitializer)
     }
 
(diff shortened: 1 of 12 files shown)


rnro:
The API breakage the checker is picking up is under SPI and I believe it is fine.

Sources/NIOHTTP2/MultiplexerAbstractChannel.swift
```diff
@@ -89,24 +90,36 @@ extension MultiplexerAbstractChannel {
...
+            let channelPromise: EventLoopPromise<Channel>?
+            if let promise = promise {
+                channelPromise = self.baseChannel.eventLoop.makePromise(of: Channel.self)
+                channelPromise!.completeWith(promise.futureResult.map { value in value as! Channel })
+            } else {
+                channelPromise = nil
+            }
+            self.baseChannel.configure(initializer: initializer, userPromise: channelPromise)
+        case .returnsAny(let initializer):
+            self.baseChannel.configure(initializer: initializer, userPromise: promise)
```
FranzBusch, line 107: I am a bit confused here but might just lack the context. Are we using both cases because one is the inbound stream and the other is creating a new outbound stream case?

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -243,7 +243,6 @@ extension NIOHTTP2Handler {
             return try await self.inlineStreamMultiplexer.createStreamChannel(initializer).get()
         }
 
-
         /// Create a stream channel initialized with the provided closure and return it wrapped within a `NIOAsyncChannel`.
```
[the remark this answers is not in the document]
rnro: So am I but I didn't think tackling the issue of its continued existence was in-scope for this PR

Sources/NIOHTTP2/MultiplexerAbstractChannel.swift
rnro: Sorry, I don't follow the question. Both cases of what? I'll take a punt at clarifying some things.

For initializer type:
- `.includesStreamID` is the legacy case
- `.excludesStreamID` is the non-async case
- `.returnsAny` is the async case

We optionally remap the promise type if we have one from `Any` to `Channel` but only if we have one. In the past we didn't supply a promise on the inbound path so if we aren't in async world then we don't supply one now.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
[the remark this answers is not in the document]
FranzBusch: Yeah sure, we should just make sure to remove this API here as well. I am fine if we do it in a follow up but maybe you could quickly check if this is just internal API right now and maybe RIP it out here as well?

Sources/NIOHTTP2/MultiplexerAbstractChannel.swift
FranzBusch: The thing I am confused then is why did something in the implementation for `.excludesStreamID` change now. If this is the non-async case can we actually hit this with a non-nil promise?

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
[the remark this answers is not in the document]
rnro: After thinking about it, I just removed the API and fixed up the test

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -304,7 +308,20 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
-        channel.configure(initializer: streamStateInitializer, userPromise: promise)
+
+        let anyInitializer: NIOChannelInitializerWithOutput<Any> = { channel in
+            streamStateInitializer(channel).map { $0 }
+        }
+
+        let anyPromise: EventLoopPromise<Any>?
+        if let promise = promise {
+            anyPromise = channel.baseChannel.eventLoop.makePromise(of: Any.self)
+            promise.completeWith(anyPromise!.futureResult.map { value in value as! Output })
```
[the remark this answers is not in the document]
glbrntt: I think we can save an allocation here by avoiding a future and completing `promise` in the `whenComplete` block of `anyPromise`:

```swift
anyPromise.futureResult.whenComplete {
    switch $0 {
    // ...
    }
}
```

Sources/NIOHTTP2/MultiplexerAbstractChannel.swift
```diff
@@ -89,24 +90,29 @@ extension MultiplexerAbstractChannel {
...
 
-    func configureInboundStream(initializer: InboundStreamStateInitializer) {
+    func configureInboundStream(initializer: InboundStreamStateInitializer, promise: EventLoopPromise<Any>?) {
         switch initializer {
         case .includesStreamID(let initializer):
             self.baseChannel.configure(initializer: initializer, userPromise: nil)
         case .excludesStreamID(let initializer):
             self.baseChannel.configure(initializer: initializer, userPromise: nil)
+        case .returnsAny(let initializer):
+            self.baseChannel.configure(initializer: initializer, userPromise: promise)
```
glbrntt, line 100: If `promise` isn't `nil` then it will be leaked if the initialiser is one of the other two cases.

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -989,9 +1004,13 @@ extension NIOHTTP2Handler {
 #if swift(>=5.7)
     /// The type of all `inboundStreamInitializer` callbacks which do not need to return data.
     public typealias StreamInitializer = NIOChannelInitializer
+    /// The type of NIO Channel initializer callbacks which need to return untyped data.
+    internal typealias StreamInitializerWithAnyOutput = @Sendable (Channel) -> EventLoopFuture<Any>
```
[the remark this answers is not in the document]
FranzBusch: Yeah that makes sense to me

FranzBusch (commented):
Overall LGTM! If we just fix up the last comments from @glbrntt

Sources/NIOHTTP2/MultiplexerAbstractChannel.swift
rnro: I've separated out the configuration function with a precondition to ensure that the promise-supplying case only accepts `.returnsAny`. I _could_ wire it up with a type transformation on the other promises but in practice they are never used at the moment.

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -304,10 +311,30 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+            streamStateInitializer(channel).map { return $0 }
+        }
+
+        let anyPromise: EventLoopPromise<Any>?
+        if let promise = promise {
+            anyPromise = channel.baseChannel.eventLoop.makePromise(of: Any.self)
+            anyPromise?.futureResult.whenComplete { result in
+                switch result {
+                case .success(let any):
+                    promise.succeed(any as! Output)
```
FranzBusch, line 325: Can we add this comment here
```suggestion
                    // The cast through any here is unfortunate but the only way to make this work right now
                    // since the HTTP2ChildChannel and the multiplexer is not generic over the output of the initializer.
                    promise.succeed(any as! Output)
```

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -439,6 +439,7 @@ extension Channel {
     ///   - configuration: The settings that will be used when establishing the connection and new streams.
     ///   - position: The position in the pipeline into which to insert this handler.
     ///   - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    ///     It must output the stream `Channel` (either directly or wrapped in some form) to allow users to iterate over inbound streams.
```
FranzBusch, line 442: This comment confuses me on all the methods here. The users can return whatever they want inside the initializer right. They might multiplex another `HTTP2` connection side a single stream and the return will be a `NIOHTTP2Handler.AsyncStreamMultiplexer<NIOHTTP2Handler.AsyncStreamMultiplexer<Output>>>`.

Additionally, we should make `Output: Sendable` here as well.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -537,6 +538,7 @@ extension Channel {
     ///   - http2ConnectionInitializer: An optional callback that will be invoked only when the negotiated protocol
     ///     is HTTP/2 to configure the connection channel.
     ///   - http2InboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    ///     It must output the stream `Channel` (either directly or wrapped in some form) to allow users to iterate over inbound streams.
```
FranzBusch, line 541: We should require all generic types to be `Sendable` here.

FranzBusch (commented):
Just a couple of minor nits left otherwise LGTM!

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
```diff
@@ -239,7 +239,7 @@ extension NIOHTTP2Handler {
         }
 
         /// Create a stream channel initialized with the provided closure
-        public func createStreamChannel<OutboundStreamOutput>(_ initializer: @escaping NIOChannelInitializerWithOutput<OutboundStreamOutput>) async throws -> OutboundStreamOutput {
+        public func createStreamChannel<OutboundStreamOutput: Sendable>(_ initializer: @escaping NIOChannelInitializerWithOutput<OutboundStreamOutput>) async throws -> OutboundStreamOutput {
```
glbrntt, line 242: I think `OutboundStreamOutput` is unnecessarily verbose here: `Output` is fine.

glbrntt (commented):
One nit, looks good otherwise.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
FranzBusch: Agreed. `Output` should be enough.

FranzBusch: I personally still think we should call this method `createOutboundStream` so that it aligns with the `inboundStreamInitializer`.

FranzBusch (approved):
LGTM! Thanks @rnro

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
rnro: You're right, this isn't correct, I was attempting to give guidance on the intent of the `Output` type but it's too narrow a description. I've reworded this to reflect the broader use.

Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
rnro: This name aligns with the equivalent method on the non-async multiplexer, I think that alignment is more important.

FranzBusch: I personally don't think that is more important because this here is the chance for a new name. I recently saw somebody giving feedback on exactly this API that they are confused what initializer runs when.

glbrntt: I think `createOutboundStream` is misleading: it sounds like it's a unidirectional stream to me, which it isn't. (FWIW I have the same issue with our inbound stream initializers.)

glbrntt:
These are all SPI changes so are safe to ignore:

```
14:05:45 5 breaking changes detected in NIOHTTP2:
14:05:45   💔 API breakage: func NIOHTTP2Handler.AsyncStreamMultiplexer.createStreamChannel(_:) has generic signature change from <InboundStreamOutput, OutboundStreamOutput> to <InboundStreamOutput, Output where Output : Swift.Sendable>
14:05:45   💔 API breakage: func Channel.configureAsyncHTTP2Pipeline(mode:configuration:position:inboundStreamInitializer:) has generic signature change from <Self, Output where Self : NIOCore.Channel> to <Self, Output where Self : NIOCore.Channel, Output : Swift.Sendable>
14:05:45   💔 API breakage: func Channel.configureAsyncHTTPServerPipeline(http2Configuration:http1ConnectionInitializer:http2ConnectionInitializer:http2InboundStreamInitializer:) has generic signature change from <Self, HTTP1ConnectionOutput, HTTP2ConnectionOutput, HTTP2StreamOutput where Self : NIOCore.Channel, HTTP1ConnectionOutput : Swift.Sendable> to <Self, HTTP1ConnectionOutput, HTTP2ConnectionOutput, HTTP2StreamOutput where Self : NIOCore.Channel, HTTP1ConnectionOutput : Swift.Sendable, HTTP2ConnectionOutput : Swift.Sendable, HTTP2StreamOutput : Swift.Sendable>
14:05:45   💔 API breakage: func SynchronousOperations.configureAsyncHTTP2Pipeline(mode:configuration:position:inboundStreamInitializer:) has generic signature change from <Output> to <Output where Output : Swift.Sendable>
14:05:45   💔 API breakage: func NIOHTTP2Handler.AsyncStreamMultiplexer.createStreamChannel(configuration:initializer:) has been removed
```

PR #419, opened 2023-10-04
Remove dependency on SPI

rnro:
Motivation:

Remove dependency on SPI so that we can cut a release without introducing dependency on code subject to change.

Modifications:

Remove all code which introduces a dependency on the NIOAsyncChannel SPI

Result:

Code inside SPI is removed


FranzBusch (approved):

glbrntt:
Not actually breaking as it's unreleased SPI.

PR #422, opened 2023-10-10
DOSHeuristics DeadlineClock more explicit on 5.6

rnro:
Motivation:

The type inference system isn't able to work out the default clock type in the DOSHeuristics init on Swift 5.6 so this change makes it more explicit

Modifications:

Add more explicit inits

Result:

The code should build on 5.6 and later


Lukasa (approved):

Lukasa:
API breakage is failing compiling the old `main`.

PR #423, opened 2023-10-11
Reinstate Async Channel API

rnro:
### Motivation:

We previously removed Async Channel SPI to allow a clean release without
adding dependency on NIO code which was subject to change, however that
code is now ready to be promoted to API.

### Modifications:

* Revert the commit which removed the Async Channel code
* Remove `AsyncChannel` SPI guards from imports and code which uses it

### Result:

* Async API is exposed as public

### DIFF
--- a/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
@@ -451,3 +451,124 @@ internal protocol AnyContinuation {
     func finish()
     func finish(throwing error: Error)
 }
+
+
+/// `NIOHTTP2InboundStreamChannels` provides access to inbound stream channels as a generic `AsyncSequence`.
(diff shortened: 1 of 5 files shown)


Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -451,3 +451,123 @@ internal protocol AnyContinuation {
...
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension NIOHTTP2InboundStreamChannels: @unchecked Sendable where Output: Sendable {}
+#endif
+
+
+#if swift(<5.9)
+// this should be available in the std lib from 5.9 onwards
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension AsyncThrowingStream {
+    public static func makeStream(
```
FranzBusch, line 563: This shouldn't be public

rnro:
https://github.com/apple/swift-nio-http2/pull/424

PR #417, opened 2023-09-04
Sendability warnings

rnro:
Motivation:

To address warnings revealed with
`.enableExperimentalFeature("StrictConcurrency=complete"),` in Swift 5.8

Modifications:

* Numerous protections of state in tests
* Make the multiplexer instances explicitly loop bound
* Update channel initializers to be marked as Sendable
* Mark error backing storage types as unchecked sendable

Result:

Better thread safety

PR #427, opened 2023-10-27
Add `Sendable` annotations to new async APIs

FranzBusch:
# Motivation
We just released our new async APIs but forgot to mark the new types as `Sendable`. This is causing warnings for users and we should properly mark them as `Sendable` since we expect users to use them from concurrent contexts.

# Modification
Adding `Sendable` to the `NIONegotiatedHTTPVersion` was pretty straight forward; however, adding `Sendable` to `AsyncStreamMultiplexer` was a bit more complicated since it stored the `InlineStreamMultiplexer` which we cannot easily make `Sendable`. Hence, we are now holding the `InlineStreamMultiplexer` in a `NIOLoopBound` and are submitting the work to the correct `EventLoop` before we call the underlying multiplexer.

# Result
No more `Sendable` warnings for our users.


Lukasa (changes requested):
@rnro has been working on changes in this area in #417. I talked him out of using `NIOLoopBound` for the use-case you used it, and I'd like to do the same here. Please align with those changes.

FranzBusch:
Let's close this in favour of @rnro PR then and work on the lense type a bit more.

PR #428, opened 2023-11-16
Make HEADERS frame payload non-indirect

Lukasa:
Motivation:

In previous patches we shrank the size of HTTP2Frame by making various data types indirect in the frame payload. This included HEADERS, which is unfortunte as HEADERS frames are quite common.

This patch changes the layout of the HEADERS frame to remove the indirect case.

Modifications:

- Move END_STREAM into an OptionSet.
- Turn the two optional bits into flags in the aforementioned OptionSet
- Replace the properties with computed properties.
- Remove the indirect case

Result:

HEADERS frames are cheaper.


dnadoba:
Can we add a unit test that checks that the size of `FramePayload`/`HTTP2Frame` doesn't exceed the maximum size of the inline existential storage? IIRC that would be 3 words / 24 Bytes on 64-bit systems.

Lukasa:
Done.

Sources/NIOHTTP2/HTTP2Frame.swift
```diff
@@ -27,9 +27,15 @@ public struct HTTP2Frame: Sendable {
...
-        public var exclusive: Bool
         public var dependency: HTTP2StreamID
+        public var exclusive: Bool
         public var weight: UInt8
+
+        internal init(exclusive: Bool, dependency: HTTP2StreamID, weight: UInt8) {
+            self.exclusive = exclusive
+            self.dependency = dependency
+            self.weight = weight
+        }
```
glbrntt, line 38: Was the `init` intentionally `internal` before? 

It's a parameter in `Headers.init` but is defaulted to `nil` so it looks like it not being public was an oversight?

dnadoba (approved):
Great performance/allocation improvement!

glbrntt (approved):
Nice! ⛳️

dnadoba:
The CI failure is interesting:
```
13:31:41 /code/Sources/NIOHTTP2/HTTP2Frame.swift:257:17: error: stored property 'booleans' of 'Sendable'-conforming struct 'Headers' has non-sendable type 'HTTP2Frame.FramePayload.Headers.Booleans'
13:31:41             var booleans: Booleans
13:31:41                 ^
13:31:41 /code/Sources/NIOHTTP2/HTTP2Frame.swift:226:20: note: consider making struct 'Booleans' conform to the 'Sendable' protocol
13:31:41             struct Booleans: OptionSet {
13:31:41                    ^
13:31:41                                       , Sendable
```
`Booleans` is an internal struct which is just made up of a single stored property which is `Sendable`. The compiler should be able to automatically synthesise `Sendable` conformance for us. Something is wrong here. Maybe `@usableFromInline` or the nesting is screwing with the compiler inference.

Sources/NIOHTTP2/HTTP2Frame.swift
dnadoba: Seems like an oversight to me as well

Lukasa:
Updated.

dnadoba:
Now we are hitting depreciation warnings because of the new NIO release:
```
15:38:56 /code/Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift:174:25: error: 'init(synchronouslyWrapping:configuration:)' is deprecated: This method has been deprecated since it defaults to deinit based resource teardown
15:38:56                     try NIOAsyncChannel(
15:38:56                         ^
15:38:56 /code/Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift:174:25: note: use 'init(wrappingChannelSynchronously:configuration:)' instead
15:38:56                     try NIOAsyncChannel(
15:38:56                         ^
15:38:56 /code/Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift:192:66: error: 'inbound' is deprecated: Use the executeThenClose scoped method instead.
15:38:56                     for try await receivedFrame in streamChannel.inbound {
15:38:56                                                                  ^
15:38:56 /code/Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift:195:49: error: 'outbound' is deprecated: Use the executeThenClose scoped method instead.
15:38:56                         try await streamChannel.outbound.write(ConfiguringPipelineAsyncMultiplexerTests.responseFramePayload)
```

dnadoba:
I have reported the `Sendable` issue: https://github.com/apple/swift/issues/70019

PR #429, opened 2023-12-15
Add missing availability guards in tests

simonjbeaumont:
## Motivation

Some of the test code was missing availability guards for Apple platforms, resulting in build failures for these platforms, e.g.

```
error: 'NIOAsyncTestingChannel' is only available in iOS 13.0 or newer
         var clientChannel: NIOAsyncTestingChannel!
```

## Modifications

Add missing availability guards. I've tried to keep them as scoped as possible. Some test suites used these as class properties, so these got a guard at suite level.

## Result

Tests can now build and run on iOS and other Apple platforms.

### DIFF
--- a/Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
+++ b/Tests/NIOHTTP2Tests/ConfiguringPipelineAsyncMultiplexerTests.swift
@@ -22,6 +22,7 @@ import NIOHTTP1
 import NIOHTTP2
 import NIOTLS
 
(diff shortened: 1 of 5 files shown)


glbrntt (approved):
This looks good but there's a heap of failures from using deprecated API (as warnings as errors is enabled in CI), we should fix that up separately.

simonjbeaumont:
> This looks good but there's a heap of failures from using deprecated API (as warnings as errors is enabled in CI), we should fix that up separately.

Looks like they all might be coming from one test. Added a commit which might fix it.

PR #430, opened 2023-12-18
Use explicit 64-bit UInt64 to support 32-bit platforms

simonjbeaumont:
## Motivation

This project doesn't currently build for watchOS because some tests make use of 64-bit integer literals.

```
IntegerCodingTests.swift:161:115: Integer literal '9151314442816848127' overflows when stored into 'Int'
```

Beyond the test code, the integer encoding logic makes use of `UInt`, which will be `UInt64` on some platforms and `UInt32` on others, e.g. some Apple Watch versions.

## Modifications

- Compile out the tests that use 64-bit integer literals with `#if !os(watchOS)`.
- Use explicit `UInt64` instead of `UInt` for encoding methods.

## Result

Code can now build for watchOS.

### DIFF
--- a/Sources/NIOHPACK/HPACKEncoder.swift
+++ b/Sources/NIOHPACK/HPACKEncoder.swift
@@ -133,12 +133,12 @@ public struct HPACKEncoder {
             self.state = .encoding
         case .resized(nil):
             // one resize
-            self.buffer.write(encodedInteger: UInt(self.allowedDynamicTableSize), prefix: 5, prefixBits: 0x20)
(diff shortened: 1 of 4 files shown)


glbrntt (approved):

FranzBusch (approved):

simonjbeaumont:
@FranzBusch @glbrntt dunno what's up with nightly pipeline but unlikely related to this. Should we merge over it?

glbrntt:
@simonjbeaumont I'll merge this given https://github.com/apple/swift-nio-http2/pull/431 has the same issue.

PR #432, opened 2023-12-20
Remove precondition: yielding to terminated AsyncSequence

rnro:
### Motivation:

If a task iterating over inbound streams using the new async API is cancelled then it can lead to the process crashing as we hit a precondition which guards against yielding to a terminated NIOHTTP2AsyncSequence.

### Modifications:

Remove the precondition.

### Result:

The code no longer crashes, instead it silently drops the yielded value.

Doing more than this such as closing the connection requires rethinking more of the API.


FranzBusch (approved):

Lukasa (changes requested):
Can you please add a test to cover this?

Lukasa (approved):

PR #433, opened 2023-12-22
public HTTP2StreamMultiplexer.SendableView

rnro:
### Motivation:

`Bench1Conn10kRequests` performance tester incorrectly used a `NIOLoopBound` to access the `HTTP2StreamMultiplexer` to create streams.

### Modifications:

Make `HTTP2StreamMultiplexer.SendableView` public and use it to pass the multiplexer into the event loop execute.

### Result:

- `HTTP2StreamMultiplexer.SendableView` is public
- `Bench1Conn10kRequests` are not broken

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
@@ -282,13 +282,18 @@ extension HTTP2StreamMultiplexer {
 extension HTTP2StreamMultiplexer {
     /// HTTP2StreamMultiplexer.SendableView exposes only the thread-safe API of HTTP2StreamMultiplexer
     ///
-    /// We use unckecked Sendable here because we always make sure we are on the right event loop
+    /// We use unchecked Sendable here because we always make sure we are on the right event loop
     /// from this code on.
-    struct SendableView: @unchecked Sendable {
+    public struct SendableView: @unchecked Sendable {
         let http2StreamMultiplexer: HTTP2StreamMultiplexer
         let eventLoop: EventLoop
 
-        func createStreamChannel(promise: EventLoopPromise<Channel>?, _ streamStateInitializer: @escaping NIOChannelInitializer) {
+        public init(http2StreamMultiplexer: HTTP2StreamMultiplexer, eventLoop: EventLoop) {
+            self.http2StreamMultiplexer = http2StreamMultiplexer
+            self.eventLoop = eventLoop
+        }
+
+        public func createStreamChannel(promise: EventLoopPromise<Channel>?, _ streamStateInitializer: @escaping NIOChannelInitializer) {
             // Always create streams channels on the next event loop tick. This avoids re-entrancy
             // issues where handlers interposed between the two HTTP/2 handlers could create streams
             // in channel active which become activated twice.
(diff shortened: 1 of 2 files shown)


Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
```diff
@@ -282,13 +282,18 @@ extension HTTP2StreamMultiplexer {
...
-    /// We use unckecked Sendable here because we always make sure we are on the right event loop
+    /// We use unchecked Sendable here because we always make sure we are on the right event loop
     /// from this code on.
-    struct SendableView: @unchecked Sendable {
+    public struct SendableView: @unchecked Sendable {
         let http2StreamMultiplexer: HTTP2StreamMultiplexer
         let eventLoop: EventLoop
 
-        func createStreamChannel(promise: EventLoopPromise<Channel>?, _ streamStateInitializer: @escaping NIOChannelInitializer) {
+        public init(http2StreamMultiplexer: HTTP2StreamMultiplexer, eventLoop: EventLoop) {
```
glbrntt, line 291: Do we want the `init` to be `public`? Normally the sendable view would be created by the `HTTP2StreamMultuplexer`.

Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
```diff
@@ -298,7 +303,7 @@ extension HTTP2StreamMultiplexer {
         }
 
         @available(*, deprecated, message: "The signature of 'streamStateInitializer' has changed to '(Channel) -> EventLoopFuture<Void>'")
-        func createStreamChannel(promise: EventLoopPromise<Channel>?, _ streamStateInitializer: @escaping NIOChannelInitializerWithStreamID) {
+        public func createStreamChannel(promise: EventLoopPromise<Channel>?, _ streamStateInitializer: @escaping NIOChannelInitializerWithStreamID) {
```
glbrntt, line 306: Why do we need to make this one `public`?

Sources/NIOHTTP2PerformanceTester/Bench1Conn10kRequests.swift
```diff
@@ -77,18 +77,26 @@ func setupServer(group: EventLoopGroup) throws -> Channel {
...
+            ],
+            position: .last
+        )
     }
-    let loopBoundMultiplexer = NIOLoopBound(multiplexer, eventLoop: channel.eventLoop)
+    let multiplexerSendableView = HTTP2StreamMultiplexer.SendableView(http2StreamMultiplexer: multiplexer, eventLoop: channel.eventLoop)
     channel.eventLoop.execute {
-        loopBoundMultiplexer.value.createStreamChannel(promise: nil, requestStreamInitializer)
+        multiplexerSendableView.createStreamChannel(promise: nil, requestStreamInitializer)
     }
```
glbrntt, line 100: I don't think we need a sendable view here:

```swift
channel.pipeline.handler(type: HTTP2StreamMultiplexer.self).whenSuccess { multiplexer in 
  multiplexer.createStreamChannel(promise: nil, requestStreamInitializer)
}
```

glbrntt (changes requested):

Sources/NIOHTTP2PerformanceTester/Bench1Conn10kRequests.swift
rnro: Yep, you're right. In which case I don't think we need this PR at all and I'll open one with your suggestion :D

rnro:
https://github.com/apple/swift-nio-http2/pull/434

PR #438, opened 2024-04-11
Avoid CoW while processing the next state when HTTP2FrameDecoder decodes

stefanadranca:
Motivation:

CoWs appear when switching over the current state of the parser and holding it while also modifying it - inside `processNExtState()`. Following `append(bytes: ByteBuffer)`'s pattern, we should use a function that sets the state to an intermediary one with no associated data, before making the transformations.

Modifications:
- updated the `avoidParserCoW()` helper function to take into consideration throwing functions
- used it in the switch cases over the parser state inside `processNextState()` 

Result:

CoW will be avoided when changing the state of the HTTP2FrameDecoder while decoding, so we won't have unnecessary heap allocations.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2FrameParser.swift
+++ b/Sources/NIOHTTP2/HTTP2FrameParser.swift
@@ -1404,7 +1404,7 @@ extension HTTP2FrameDecoder {
             throw error
         }
     }
-    
+
     private var isAppending: Bool {
         if case .appending = self.state {
             return true
--- a/Sources/NIOHTTP2/HTTP2FrameParser.swift
+++ b/Sources/NIOHTTP2/HTTP2FrameParser.swift
@@ -793,38 +793,38 @@ struct HTTP2FrameDecoder {
             return try self.avoidingParserCoW { newState in
                 guard let processResult = try state.process() else {
                     newState = .awaitingClientMagic(state)
-                    return ParseResult.needMoreData
+                    return .needMoreData
                 }
                 newState = .accumulatingFrameHeader(processResult)
-                return ParseResult.continue
+                return .continue
             }
             
         case .initialized:
             // no bytes, no frame
             return .needMoreData
-            
+
         case .accumulatingFrameHeader(var state):
             let maxFrameSize = self.maxFrameSize
             let maxHeaderListSize = self.headerDecoder.maxHeaderListSize
             return try self.avoidingParserCoW { newState in
                 guard let targetState = try state.process(maxFrameSize: maxFrameSize, maxHeaderListSize: maxHeaderListSize) else {
                     newState = .accumulatingFrameHeader(state)
-                    return ParseResult.needMoreData
+                    return .needMoreData
                 }
                 
                 newState = .init(targetState)
-                return ParseResult.continue
+                return .continue
             }
             
         case .awaitingPaddingLengthByte(var state):
             return try self.avoidingParserCoW { newState in
                 guard let targetState = try state.process() else {
                     newState = .awaitingPaddingLengthByte(state)
-                    return ParseResult.needMoreData
+                    return .needMoreData
                 }
                 
                 newState = .init(targetState)
-                return ParseResult.continue
+                return .continue
             }
             
         case .accumulatingPayload(var state):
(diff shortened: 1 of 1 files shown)


Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -1355,6 +1394,26 @@ extension HTTP2FrameDecoder {
         return body(&self.state)
     }
 
+    /// This is the implementation for a throwing function. If the state doesn't get updated because an error
+    /// was thrown, this function sets the parser state to the initial one.
+    ///
+    /// The `body` function must assign a value to the parser state, otherwise it will be the intermediary
+    /// `appending` state, which is not wanted and it results in the this function's assertion failure.
+    @inline(__always)
+    private mutating func avoidingParserCoW<ReturnType>(_ body: (inout ParserState) throws -> ReturnType) throws -> ReturnType {
```
glbrntt, line 1403: Rather than adding a complete new function can we just modify the existing one to accept a throwing closing and use `rethrows` to propagate the throwing-ness of the closure?

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -781,60 +781,79 @@ struct HTTP2FrameDecoder {
...
         case .awaitingClientMagic(var state):
-            guard let newState = try state.process() else {
-                return .needMoreData
+            return try self.avoidingParserCoW { newState in
+                guard let processResult = try state.process() else {
+                    newState = .awaitingClientMagic(state)
+                    return ParseResult.needMoreData
+                }
+                newState = .accumulatingFrameHeader(processResult)
+                return ParseResult.continue
```
glbrntt, line 799: We should be able to use type inference here (and elsewhere in this PR)

```suggestion
                    return .needMoreData
                }
                newState = .accumulatingFrameHeader(processResult)
                return .continue
```

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -781,60 +781,79 @@ struct HTTP2FrameDecoder {
...
-
-            self.state = .accumulatingFrameHeader(newState)
-            return .continue
-
+            
         case .initialized:
             // no bytes, no frame
             return .needMoreData
-
+
```
glbrntt, line 805: nit: trailing whitespace

```suggestion

```

glbrntt (changes requested):

glbrntt:

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -1345,14 +1384,25 @@ extension HTTP2FrameDecoder {
...
     /// not ideal.
+    ///
+    /// The `body` function must assign a value to the parser state, otherwise it will remain the intermediary
+    /// `appending` state, which is not wanted and it would result in the failure of the final assertion.
+    /// If the state doesn't get updated because of an error that was thrown, this function sets the parser state
+    /// to the initial one.
     @inline(__always)
-    private mutating func avoidingParserCoW<ReturnType>(_ body: (inout ParserState) -> ReturnType) -> ReturnType {
+    private mutating func avoidingParserCoW<ReturnType>(_ body: (inout ParserState) throws -> ReturnType) rethrows -> ReturnType {
+        let initialState = self.state
```
glbrntt, line 1394: Sorry I didn't catch this before, but I think by doing this we're undoing the effectiveness of this function. The reason is that the closure will be capturing the associated payload held by `self.state`, now we're grabbing it here again so we'll CoW if we mutate the one we captured in the closure.

I don't think restoring the state in case of an error is actually necessary either: the previous code didn't do it.

glbrntt (changes requested):

Sources/NIOHTTP2/HTTP2FrameParser.swift
stefanadranca: You are right, I'm trying to think how to solve this. The problem is that in the case of an error that assertion that checks the final state fails - the previous implementation didn't receive a throwing function as a parameter so the state was always updated to a valid one (in the body of that function). In the `append()` function where `avoidingParserCoW` was already used there are no throwing function calls.

stefanadranca: should we get rid of that assertion if the function is throwing?

glbrntt: > should we get rid of that assertion if an error is thrown?

No the assertion is there for a reason: we must never leave state in the `appending` state, otherwise the next call which switches over the state won't have any idea what it can do!

stefanadranca: Thanks!

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -781,60 +781,117 @@ struct HTTP2FrameDecoder {
...
-            return .continue
-
+            switch result {
+            case let .success(parseResult):
+                return parseResult
+            case let .failure(error):
+                newState = .awaitingClientMagic(state)
+                throw error
+            }
+        }
```
glbrntt, line 809: Can you fix the indentation here?

glbrntt (changes requested):

glbrntt (approved):

PR #439, opened 2024-05-01
Add a variant of configureAsyncHTTP2Pipeline which takes a stream delegate

glbrntt:
Motivation:

`configureAsyncHTTP2Pipeline` doesn't allow a stream delegate to be specified. As the async pipeline uses the "inline" stream multiplexer there's no way to account for streams within the connection channel.

Modifications:

- Add a sync and async variants of `configureAsyncHTTP2Pipeline` which accepts an optional stream delegate
- Rewrite the existing helpers in terms of the new one

Result:

Users can configure an async http pipeline with a stream delegate.


Lukasa (changes requested):
Looks fine, but can we add a test for this?

Lukasa (approved):

PR #443, opened 2024-06-24
Reject long sequences of CONTINUATION frames

clintonpi:
Motivation:

Long sequences of CONTINUATION frames can be used to mount attacks by attempting to get a remote peer to consume large amounts of memory.

Modifications:

- Add a limit to the number of sequential CONTINUATION frames that can be received. This limit is configurable by users at the NIOHTTP2Handler level and has a default value of 5. When this limit is exceeded, the recipient responds with a GOAWAY frame and an "Enhance Your Calm" error of a newly created type `ExcessiveContinuationFrames`.

Result:

Long sequences of CONTINUATION frames are now rejected by the recipient.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
@@ -1129,7 +1129,7 @@ extension NIOHTTP2Handler {
         public var contentLengthValidation: ValidationState = .enabled
         public var maximumSequentialEmptyDataFrames: Int = 1
         public var maximumBufferedControlFrames: Int = 10000
-        public let maximumSequentialContinuationFrames: Int = 5
+        public var maximumSequentialContinuationFrames: Int = 5
         public init() {}
     }
 
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
@@ -232,14 +232,12 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
     ///   - maximumBufferedControlFrames: Controls the maximum buffer size of buffered outbound control frames. If we are unable to send control frames as
     ///         fast as we produce them we risk building up an unbounded buffer and exhausting our memory. To protect against this DoS vector, we put an
     ///         upper limit on the depth of this queue. Defaults to 10,000.
-    ///   - maximumSequentialContinuationFrames: The maximum number of sequential CONTINUATION frames.
     public convenience init(mode: ParserMode,
                             initialSettings: HTTP2Settings = nioDefaultSettings,
                             headerBlockValidation: ValidationState = .enabled,
                             contentLengthValidation: ValidationState = .enabled,
                             maximumSequentialEmptyDataFrames: Int = 1,
-                            maximumBufferedControlFrames: Int = 10000,
-                            maximumSequentialContinuationFrames: Int = 5) {
+                            maximumBufferedControlFrames: Int = 10000) {
         self.init(mode: mode,
                   eventLoop: nil,
                   initialSettings: initialSettings,
@@ -247,7 +245,7 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
                   contentLengthValidation: contentLengthValidation,
                   maximumSequentialEmptyDataFrames: maximumSequentialEmptyDataFrames,
                   maximumBufferedControlFrames: maximumBufferedControlFrames,
-                  maximumSequentialContinuationFrames: maximumSequentialContinuationFrames,
+                  maximumSequentialContinuationFrames: 5,
                   maximumResetFrameCount: 200,
                   resetFrameCounterWindow: .seconds(30))
 
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
@@ -104,6 +104,9 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
     /// The delegate for (de)multiplexing inbound streams.
     private var inboundStreamMultiplexerState: InboundStreamMultiplexerState
 
+    /// The maximum number of sequential CONTINUATION frames.
+    private var maximumSequentialContinuationFrames: Int
+
     @usableFromInline
     internal var inboundStreamMultiplexer: InboundStreamMultiplexer? {
         return self.inboundStreamMultiplexerState.multiplexer
@@ -212,6 +215,7 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
                   contentLengthValidation: contentLengthValidation,
                   maximumSequentialEmptyDataFrames: 1,
                   maximumBufferedControlFrames: 10000,
+                  maximumSequentialContinuationFrames: 5,
                   maximumResetFrameCount: 200,
                   resetFrameCounterWindow: .seconds(30))
     }
@@ -228,12 +232,14 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
     ///   - maximumBufferedControlFrames: Controls the maximum buffer size of buffered outbound control frames. If we are unable to send control frames as
     ///         fast as we produce them we risk building up an unbounded buffer and exhausting our memory. To protect against this DoS vector, we put an
     ///         upper limit on the depth of this queue. Defaults to 10,000.
+    ///   - maximumSequentialContinuationFrames: The maximum number of sequential CONTINUATION frames.
     public convenience init(mode: ParserMode,
                             initialSettings: HTTP2Settings = nioDefaultSettings,
                             headerBlockValidation: ValidationState = .enabled,
                             contentLengthValidation: ValidationState = .enabled,
                             maximumSequentialEmptyDataFrames: Int = 1,
-                            maximumBufferedControlFrames: Int = 10000) {
+                            maximumBufferedControlFrames: Int = 10000,
+                            maximumSequentialContinuationFrames: Int = 5) {
         self.init(mode: mode,
                   eventLoop: nil,
                   initialSettings: initialSettings,
@@ -241,6 +247,7 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
                   contentLengthValidation: contentLengthValidation,
                   maximumSequentialEmptyDataFrames: maximumSequentialEmptyDataFrames,
                   maximumBufferedControlFrames: maximumBufferedControlFrames,
+                  maximumSequentialContinuationFrames: maximumSequentialContinuationFrames,
                   maximumResetFrameCount: 200,
                   resetFrameCounterWindow: .seconds(30))
 
@@ -262,6 +269,7 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
                   contentLengthValidation: connectionConfiguration.contentLengthValidation,
                   maximumSequentialEmptyDataFrames: connectionConfiguration.maximumSequentialEmptyDataFrames,
                   maximumBufferedControlFrames: connectionConfiguration.maximumBufferedControlFrames,
+                  maximumSequentialContinuationFrames: connectionConfiguration.maximumSequentialContinuationFrames,
                   maximumResetFrameCount: streamConfiguration.streamResetFrameRateLimit.maximumCount,
                   resetFrameCounterWindow: streamConfiguration.streamResetFrameRateLimit.windowLength)
     }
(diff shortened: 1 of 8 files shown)


Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -31,9 +31,11 @@ struct HTTP2FrameDecoder {
     /// The state for a parser that is waiting for the client magic.
     private struct ClientMagicState {
         private var pendingBytes: ByteBuffer?
+        private var maximumSequentialContinuationFrames: Int
```
glbrntt, line 34: I don't think this value should ever change, it should be a `let`

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -58,14 +63,16 @@ struct HTTP2FrameDecoder {
     /// The state for a parser that is currently accumulating the bytes of a frame header.
     private struct AccumulatingFrameHeaderParserState {
         private(set) var unusedBytes: ByteBuffer
+        private(set) var maximumSequentialContinuationFrames: Int
```
glbrntt, line 66: Same here, can this ever change?

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -105,11 +112,19 @@ struct HTTP2FrameDecoder {
                         throw InternalError.codecError(code: .protocolError)
                     }
                     return .awaitingPaddingLengthByte(
-                        AwaitingPaddingLengthByteParserState(fromAccumulatingFrameHeader: self, frameHeader: header)
+                        AwaitingPaddingLengthByteParserState(
+                            fromAccumulatingFrameHeader: self, frameHeader: header,
+                            maximumSequentialContinuationFrames: self.maximumSequentialContinuationFrames
```
glbrntt, line 117: If you're spreading arguments across multiple lines do one per line

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -146,14 +177,20 @@ struct HTTP2FrameDecoder {
     private struct AwaitingPaddingLengthByteParserState {
         private var header: FrameHeader
         private var accumulatedBytes: ByteBuffer
+        private var maximumSequentialContinuationFrames: Int
```
glbrntt, line 180: same here re: `var`/`let`

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -146,14 +177,20 @@ struct HTTP2FrameDecoder {
...
 
-        init(fromAccumulatingFrameHeader state: AccumulatingFrameHeaderParserState, frameHeader: FrameHeader) {
```
glbrntt, line 150: The `AccumulatingFrameHeaderParserState` stores `maximumSequentialContinuationFrames` so we don't need to pass it in again, we can just copy it from `state`.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -212,14 +251,25 @@ struct HTTP2FrameDecoder {
         private(set) var header: FrameHeader
         private var accumulatedBytes: ByteBuffer
         private var expectedPadding: Int?
+        private var maximumSequentialContinuationFrames: Int
 
-        init(fromIdle state: AccumulatingFrameHeaderParserState, header: FrameHeader) {
+        init(
+            fromIdle state: AccumulatingFrameHeaderParserState,
+            header: FrameHeader,
+            maximumSequentialContinuationFrames: Int
```
glbrntt, line 259: Same here, we can just copy it from `state` rather than pasing it in separately

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -323,16 +384,28 @@ struct HTTP2FrameDecoder {
...
+            header: FrameHeader,
+            remainingBytes: Int,
+            maximumSequentialContinuationFrames: Int
+        ) {
             self.header = header
             self.payload = state.unusedBytes
             self.expectedPadding = nil
             self.remainingByteCount = remainingBytes
             self.flowControlledLength = header.length
+            self.maximumSequentialContinuationFrames = maximumSequentialContinuationFrames
```
glbrntt, line 400: same here re: copying from `state`

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -450,21 +530,34 @@ struct HTTP2FrameDecoder {
...
+            excessBytes: ByteBuffer,
+            expectedPadding: Int,
+            maximumSequentialContinuationFrames: Int
+        ) {
             precondition(expectedPadding > 0)
 
             self.header = state.header
             self.excessBytes = excessBytes
             self.expectedPadding = expectedPadding
+            self.maximumSequentialContinuationFrames = maximumSequentialContinuationFrames
```
glbrntt, line 546: same here, let's just copy from `state`

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -450,21 +530,34 @@ struct HTTP2FrameDecoder {
...
```
glbrntt, line 560: same here, let's just copy from `state`

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -504,24 +602,37 @@ struct HTTP2FrameDecoder {
         private var currentFrameBytes: ByteBuffer
         private var continuationPayload: ByteBuffer
         private var originalPaddingBytes: Int?
-
-        init(fromAccumulatingHeaderBlockFragments acc: AccumulatingHeaderBlockFragmentsParserState,
-             continuationHeader: FrameHeader) {
+        private var sequentialContinuationFramesSize: Int
```
glbrntt, line 605: nit: prefer `count` to `size` when counting things

glbrntt (changes requested):

Sources/NIOHTTP2/HTTP2FrameParser.swift
clintonpi: No, it can't.

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -685,6 +775,7 @@ struct HTTP2FrameDecoder {
 
     internal var headerDecoder: HPACKDecoder
     private var state: ParserState
+    private var maximumSequentialContinuationFrames: Int
```
glbrntt, line 778: This should be a `let`. Also we don't need to pass `maximumSequentialContinuationFrames` through all of the states, I think we only need to pass it to the `process` func of the appropriate state.

glbrntt (changes requested):

Sources/NIOHTTP2/HTTP2FrameParser.swift
clintonpi: > Also we don't need to pass maximumSequentialContinuationFrames through all of the states, I think we only need to pass it to the process func of the appropriate state.

Yes. I completely missed that. Thanks!

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -1109,6 +1129,7 @@ extension NIOHTTP2Handler {
         public var contentLengthValidation: ValidationState = .enabled
         public var maximumSequentialEmptyDataFrames: Int = 1
         public var maximumBufferedControlFrames: Int = 10000
+        public let maximumSequentialContinuationFrames: Int = 5
```
glbrntt, line 1132: This should be `var` as its a config that users can adjust before passing in

```suggestion
        public var maximumSequentialContinuationFrames: Int = 5
```

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -601,6 +619,11 @@ struct HTTP2FrameDecoder {
                 throw NIOHTTP2Errors.excessivelyLargeHeaderBlock()
             }
 
+            // The sequence of CONTINUATION frames received is not longer than the configured limit
+            guard self.sequentialContinuationFramesCount < maximumSequentialContinuationFrames else {
```
glbrntt, line 623: Shouldn't this be `<=`?

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -769,7 +799,11 @@ struct HTTP2FrameDecoder {
     mutating func nextFrame() throws -> (HTTP2Frame, flowControlledLength: Int)? {
         // Start running through our state machine until we run out of bytes or we emit a frame.
         while true {
-            switch (try self.processNextState()) {
+            switch (
+                try self.processNextState(
+                    maximumSequentialContinuationFrames: self.maximumSequentialContinuationFrames
+                )
```
glbrntt, line 805: We don't need other pass this in here: `maximumSequentialContinuationFrames` is already stored on `self` so we can just read it from `self` in `processNextState`

glbrntt (changes requested):

Sources/NIOHTTP2/HTTP2FrameParser.swift
clintonpi: No, it shouldn't be. But I have made some adjustments to make the condition clearer.

Sources/NIOHTTP2/HTTP2FrameParser.swift
glbrntt: This should still be a `let`

Sources/NIOHTTP2/HTTP2FrameParser.swift
```diff
@@ -769,7 +796,9 @@ struct HTTP2FrameDecoder {
     mutating func nextFrame() throws -> (HTTP2Frame, flowControlledLength: Int)? {
         // Start running through our state machine until we run out of bytes or we emit a frame.
         while true {
-            switch (try self.processNextState()) {
+            switch (
+                try self.processNextState()
+            ) {
```
glbrntt, line 801:
```suggestion
            switch (try self.processNextState()) {
```

Tests/NIOHTTP2Tests/SimpleClientServerFramePayloadStreamTests.swift
```diff
@@ -1789,6 +1794,65 @@ class SimpleClientServerFramePayloadStreamTests: XCTestCase {
...
+        )
+
+        let headersFrame: [UInt8] = [
+            0x00, 0x00, 0x01,           // 3-byte payload length (1 byte)
+            0x01,                       // 1-byte frame type (HEADERS)
+            0x00,                       // 1-byte flags (none)
+            0x00, 0x00, 0x00, 0x03,     // 4-byte stream identifier
+            0x00,                       // payload
+
+        ]
```
glbrntt, line 1812:
```suggestion
            0x00,                       // payload
        ]
```

Tests/NIOHTTP2Tests/SimpleClientServerFramePayloadStreamTests.swift
```diff
@@ -1789,6 +1794,65 @@ class SimpleClientServerFramePayloadStreamTests: XCTestCase {
...
+    func testForbidsExceedingMaximumSequentialContinuationFrames() throws {
+        let maximumSequentialContinuationFrames = 5
+
+        // Begin by getting the connection up.
+        try self.basicHTTP2Connection(
+            maximumSequentialContinuationFrames: maximumSequentialContinuationFrames
+        )
+
+        let headersFrame: [UInt8] = [
+            0x00, 0x00, 0x01,           // 3-byte payload length (1 byte)
```
glbrntt, line 1806: Why do we have a 1-byte payload?

Tests/NIOHTTP2Tests/HTTP2InlineStreamMultiplexerTests.swift
```diff
@@ -35,8 +35,15 @@ private struct MyError: Error { }
 typealias IODataWriteRecorder = WriteRecorder<IOData>
 
 extension IODataWriteRecorder {
-    func drainConnectionSetupWrites(mode: NIOHTTP2Handler.ParserMode = .server) throws {
-        var frameDecoder = HTTP2FrameDecoder(allocator: ByteBufferAllocator(), expectClientMagic: mode == .client)
+    func drainConnectionSetupWrites(
+        mode: NIOHTTP2Handler.ParserMode = .server,
+        maximumSequentialContinuationFrames: Int
```
glbrntt, line 40: Can we just default this to 5? Doing that means we can avoid all the other changes in this file.

glbrntt (changes requested):

glbrntt:

glbrntt (approved):

FranzBusch (commented):
LGTM just one nit

FranzBusch (approved):
Awesome! LGTM

PR #444, opened 2024-06-28
Discard read bytes when accumulating continuation frames

glbrntt:
Motivation:

When accumulating sequences of CONTINUATION frames, each frame is parsed from a buffer. These bytes are read when the CONTINUATION frame is parsed, but if more CONTINUATION frames follow then the buffer isn't reset. This means that long sequences of CONTINUATION frames can result in a larger than necessary buffer where most of the contents have already been read.

Modifications:

- Discard the bytes of the accumulation buffer when transitioning back to AccumulatingHeaderBlockFragmentsParserState if more than half of the buffer has been read.

Result:

Lower memory footprint when parsing sequences of CONTINUATION frames.


Lukasa (approved):
Nice one.

PR #445, opened 2024-07-04
Fix bug where client ignores HEADERS frames on open stream when locally quiescing

aryan-25:
### Motivation:

Currently, HEADERS frames are ignored by the client on a client-initiated stream after the client sends a GOAWAY.

As per [Section 6.8 (GOAWAY) in RFC 9113](https://httpwg.org/specs/rfc9113#GOAWAY):
_"Once the GOAWAY is sent, the sender will ignore frames sent on streams **initiated by the receiver** if the stream has an identifier higher than the included last stream identifier."_

In this case, the client (sender) should **not** ignore the HEADERS frame as the stream is initiated by the _client_, not the receiver.

### Modifications:

Fixed the condition in the `receiveHeaders` function (given `LocallyQuiescingState`) in `ReceivingHeadersState.swift` to work for _both_ the client and server roles -- before this change, the condition is only correct when `self.role == ConnectionRole.server`.

### Result:

HEADERS frames are not ignored by the client on a client-initiated stream when locally quiescing.

### DIFF
--- a/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
+++ b/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
@@ -1753,19 +1753,6 @@ extension HTTP2StreamID {
     }
 }
 
-extension HTTP2ConnectionStateMachine.ConnectionRole {
-    /// Obtain the inverse role to self
-    /// - Returns: The inverse role
-    fileprivate func inverse() -> Self {
-        switch self {
-        case .server:
-            return .client
-        case .client:
-            return .server
-        }
-    }
-}
-
 /// A simple protocol that provides helpers that apply to all connection states that keep track of a role.
 protocol ConnectionStateWithRole {
     var role: HTTP2ConnectionStateMachine.ConnectionRole { get }
@@ -1773,7 +1760,12 @@ protocol ConnectionStateWithRole {
 
 extension ConnectionStateWithRole {
     var peerRole: HTTP2ConnectionStateMachine.ConnectionRole {
-        return self.role.inverse()
+        switch self.role {
+        case .client:
+            return .server
+        default:
+            return .client
+        }
     }
 }
 
--- a/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
+++ b/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
@@ -1764,7 +1764,7 @@ extension ConnectionStateWithRole {
         switch self.role {
         case .client:
             return .server
-        default:
+        case .server:
             return .client
         }
     }
--- a/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
+++ b/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
@@ -1753,20 +1753,27 @@ extension HTTP2StreamID {
     }
 }
 
+extension HTTP2ConnectionStateMachine.ConnectionRole {
+    /// Obtain the inverse role to self
+    /// - Returns: The inverse role
+    fileprivate func inverse() -> Self {
+        switch self {
+        case .server:
+            return .client
+        case .client:
+            return .server
+        }
+    }
+}
 
 /// A simple protocol that provides helpers that apply to all connection states that keep track of a role.
-private protocol ConnectionStateWithRole {
+protocol ConnectionStateWithRole {
     var role: HTTP2ConnectionStateMachine.ConnectionRole { get }
 }
 
 extension ConnectionStateWithRole {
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP2/ConnectionStateMachine/FrameReceivingStates/ReceivingHeadersState.swift
```diff
@@ -74,7 +86,10 @@ extension ReceivingHeadersState where Self: LocallyQuiescingState {
         let localSupportsExtendedConnect = self.localSupportsExtendedConnect
         let remoteSupportsExtendedConnect = self.remoteSupportsExtendedConnect
 
-        if streamID.mayBeInitiatedBy(.client) && streamID > self.lastRemoteStreamID {
+        // We are in `LocallyQuiescingState`. The sender of the GOAWAY is `self.role`, so the receiver is the inverse of `self.role`.
+        let receiver = self.role.inverse()
```
glbrntt, line 90: I find `receiver` really misleading because the state machine is dealing with receiving headers and `receiver` doesn't refer to the state machine but rather the remote peer. I think `remotePeer` would be clearer here.

Tests/NIOHTTP2Tests/ConnectionStateMachineTests.swift
```diff
@@ -985,6 +985,29 @@ class ConnectionStateMachineTests: XCTestCase {
         XCTAssertTrue(self.client.fullyQuiesced)
     }
 
+    func testHeadersOnOpenStreamLocallyQuiescing() {
+        let streamOne = HTTP2StreamID(1)
+
+        exchangePreamble()
```
glbrntt, line 991: nit: explicit self

```suggestion
        self.exchangePreamble()
```

Sources/NIOHTTP2/ConnectionStateMachine/FrameReceivingStates/ReceivingHeadersState.swift
```diff
@@ -61,6 +61,18 @@ extension ReceivingHeadersState {
...
+    /// - Returns: The inverse role
+    fileprivate func inverse() -> Self {
+        switch self {
+        case .server:
+            return .client
+        case .client:
+            return .server
+        }
+    }
+}
```
glbrntt, line 75: nit: can we move this to where `ConnectionRole` is defined?

glbrntt (changes requested):
This looks great, I only style/naming comments.

glbrntt:

Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
```diff
@@ -1753,20 +1753,27 @@ extension HTTP2StreamID {
...
+    /// - Returns: The inverse role
+    fileprivate func inverse() -> Self {
+        switch self {
+        case .server:
+            return .client
+        case .client:
+            return .server
+        }
+    }
+}
```
Lukasa, line 1767: This refactor can probably be backed out: we aren't calling this anywhere other than where it was originally defined, so out-of-lining it hasn't brought us any utility. Let's remove the churn in the code instead.

Sources/NIOHTTP2/ConnectionStateMachine/LocallyQuiescingState.swift
```diff
@@ -16,6 +16,6 @@
 /// the remote peer may not initiate new streams.
 ///
 /// This protocol should only be conformed to by states for the HTTP/2 connection state machine.
-protocol LocallyQuiescingState {
+protocol LocallyQuiescingState: ConnectionStateWithRole {
```
Lukasa, line 19: I don't think the refinement logically applies here: it belongs on `ReceivingHeadersState`.

Sources/NIOHTTP2/ConnectionStateMachine/FrameReceivingStates/ReceivingHeadersState.swift
```diff
@@ -17,7 +17,7 @@ import NIOHPACK
 /// can validly accept headers.
 ///
 /// This protocol should only be conformed to by states for the HTTP/2 connection state machine.
-protocol ReceivingHeadersState: HasFlowControlWindows, HasLocalExtendedConnectSettings, HasRemoteExtendedConnectSettings {
+protocol ReceivingHeadersState: HasFlowControlWindows, HasLocalExtendedConnectSettings, HasRemoteExtendedConnectSettings, ConnectionStateWithRole {
```
Lukasa, line 20: Do we need the protocol constraint here at all? We already have the `role`.

Lukasa (approved):
Great, I'm happy if George is happy.

glbrntt (approved):
Awesome, thanks @aryan-25!

PR #447, opened 2024-07-05
Keep capacity when dropping pending reads in the stream channel

glbrntt:
Motivation:

When the stream channel has closed it drops all of its pending reads. It does this by calling 'removeAll()' on its CircularBuffer. The buffers underlying storage is emptied and then a `nil` is appended (internally CircularBuffer is never empty) which triggers an allocation.

Modifications:

- Remove all but keep capacity

Result:

Fewer allocations

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamChannel.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamChannel.swift
@@ -728,7 +728,7 @@ private extension HTTP2StreamChannel {
     /// Drop all pending reads.
     private func dropPendingReads() {
         /// We don't need to report the dropped reads, just remove them all.
-        self.pendingReads.removeAll()
+        self.pendingReads.removeAll(keepingCapacity: true)
     }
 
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -728,7 +728,7 @@ private extension HTTP2StreamChannel {
     /// Drop all pending reads.
     private func dropPendingReads() {
         /// We don't need to report the dropped reads, just remove them all.
-        self.pendingReads.removeAll()
+        self.pendingReads.removeAll(keepingCapacity: true)
```
gjcairo, line 731: Should we worry about this potentially keeping an unneeded/too-big capacity for the duration of the stream?

glbrntt: No, this is only called when the channel is closing.

gjcairo (approved):

PR #449, opened 2024-07-12
Reduce allocs in HTTP2StreamChannel

glbrntt:
Motivation:

The HTTP2StreamChannel stores two atomic booleans for whether the channel is active and writable. These are only accessed externally and access is relatively infrequent. Each atomic requires an allocation, instead we can just protect each with a single lock.

Modifications:

- Combine the two atomics into a single 'flags' struct protected by a locked value box

Result:

Fewer alloations

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamChannel.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamChannel.swift
@@ -925,7 +925,7 @@ internal extension HTTP2StreamChannel {
             // Do nothing here.
             return
         case .remoteActive, .active, .closing, .closingNeverActivated, .closed:
-            self.flags.withLockedValue { $0.isActive = localValue }
+            self.flags.withLockedValue { $0.isWritable = localValue }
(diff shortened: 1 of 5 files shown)


Lukasa:
It would be useful to get an idea of whether there's any significant performance impact to this change. I _suspect_ that for most applications it comes out about neutral, but I'm not sure.

glbrntt:
> It would be useful to get an idea of whether there's any significant performance impact to this change. I _suspect_ that for most applications it comes out about neutral, but I'm not sure.

This will obviously depend on how frequently `isActive`/`isWritable` are called. My gut feeling is that most users rarely call these. 

A concrete example is AHC which uses it a few times per request (in channel active and when writability changes). I used it to benchmark against httpbin (running locally) and the difference was negligible.

gjcairo:
You'll need to update the allocations for main too

Lukasa (approved):
As the change is low risk and easily reverted, I'm fine with it.

PR #450, opened 2024-07-12
Reduce allocations on `InlineStreamMultiplexer/createStreamChannel`

gjcairo:
## Motivation
When creating stream channels on the `InlineStreamMultiplexer`, we are currently performing a `flatSubmit` on two different codepaths. This allocates two ELFs: one on `submit`, and one on `flatMap`. This is unnecessary, as we could do with just one allocation.

## Modifications
Create a single ELP and call `execute` on the EL instead of using `flatSubmit`.

## Result
Fewer allocations

### DIFF
--- a/docker/docker-compose.2204.main.yaml
+++ b/docker/docker-compose.2204.main.yaml
@@ -39,7 +39,7 @@ services:
       - MAX_ALLOCS_ALLOWED_client_server_request_response_many_inline=889050
       - MAX_ALLOCS_ALLOWED_create_client_stream_channel=37050
       - MAX_ALLOCS_ALLOWED_create_client_stream_channel_inline=37050
-      - MAX_ALLOCS_ALLOWED_create_client_stream_channel_inline_no_promise_based_API=44050
+      - MAX_ALLOCS_ALLOWED_create_client_stream_channel_inline_no_promise_based_API=37050
       - MAX_ALLOCS_ALLOWED_create_client_stream_channel_no_promise_based_API=37050
       - MAX_ALLOCS_ALLOWED_get_100000_headers_canonical_form=200050
       - MAX_ALLOCS_ALLOWED_get_100000_headers_canonical_form_trimming_whitespace=200050
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler+InlineStreamMultiplexer.swift
@@ -152,6 +152,11 @@ extension InlineStreamMultiplexer {
     }
 
     @inlinable
+    internal func createStreamChannel<Output: Sendable>(promise: EventLoopPromise<Output>?, _ initializer: @escaping NIOChannelInitializerWithOutput<Output>) {
+        self._commonStreamMultiplexer.createStreamChannel(multiplexer: .inline(self), promise: promise, initializer)
+    }
+
+    @inlinable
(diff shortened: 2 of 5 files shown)


glbrntt:

gjcairo:

gjcairo:

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -173,12 +173,13 @@ extension Channel {
                 )
             }
         } else {
+            let unsafePosition = UnsafeTransfer(position)
```
glbrntt, line 176: I don't think we should be doing this: `position` isn't sendable because it can contain a `ChannelHandler` which might not be `Sendable`. All we're doing here is suppressing a correct warning.

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
```diff
@@ -219,13 +220,14 @@ extension Channel {
...
```
glbrntt, line 223: Same here: I think we should live with the warning

glbrntt (changes requested):
Looks good apart from fixing the warnings!

Sources/NIOHTTP2/HTTP2PipelineHelpers.swift
gjcairo: Oh yeah this is not supposed to stay :) that's why this was just opened as draft. I just wanted to get the nightly pipeline to not fatal error (it's got warnings as errors enabled so it's been failing because of this for a while now) to double check the tests passed. But all of the pipelines seem broken right now anyways because of some Jenkins error I believe so I'll just revert this and trust in my local builds passing. 

We need to fix this somehow though, or remove the warns as errors flag temporarily.

gjcairo:
I updated the allocations for nightly because they had been outdated for a while.

However this change doesn't affect any allocation tests: this is because allocation tests that call `createStreamChannel`, call the `createStreamChannel(promise:_:)` version (instead of `createStreamChannel(_:)`), which is unaffected by this change. We could duplicate the allocation tests to use `createStreamChannel(_:)`, or modify them to stop using the version with the promise, but I don't know if that's worth it/desired?

glbrntt:
> However this change doesn't affect any allocation tests: this is because allocation tests that call `createStreamChannel`, call the `createStreamChannel(promise:_:)` version (instead of `createStreamChannel(_:)`), which is unaffected by this change. We could duplicate the allocation tests to use `createStreamChannel(_:)`, or modify them to stop using the version with the promise, but I don't know if that's worth it/desired?

I think it's probably worth adding a version of the test which doesn't pass in the promise in a separate PR and verifying that we do save allocations here.

docker/docker-compose.2204.main.yaml
```diff
@@ -39,7 +39,7 @@ services:
       - MAX_ALLOCS_ALLOWED_client_server_request_response_many_inline=889050
       - MAX_ALLOCS_ALLOWED_create_client_stream_channel=37050
       - MAX_ALLOCS_ALLOWED_create_client_stream_channel_inline=37050
-      - MAX_ALLOCS_ALLOWED_create_client_stream_channel_inline_no_promise_based_API=44050
```
gjcairo, line 42: Updated the allocations of the newly added test in the latest commit

glbrntt (approved):
Nice!

PR #451, opened 2024-07-15
Add alloc tests for alternative stream creation API

gjcairo:
Allocation tests currently use a func that takes a promise to create the stream multiplexer.
However, we also provide a different API that doesn't require a promise and instead returns an `EventLoopFuture<Channel>`.
We don't currently have allocation tests for the latter, so this PR duplicates the current H2 tests but uses the non-promise API to create the multiplexer.
I also had to add a new method on the old H2 multiplexer to conform to a test protocol.


docker/docker-compose.2204.59.yaml
```diff
@@ -38,6 +38,10 @@ services:
       - MAX_ALLOCS_ALLOWED_client_server_request_response_inline=244050
       - MAX_ALLOCS_ALLOWED_client_server_request_response_many=1198050
       - MAX_ALLOCS_ALLOWED_client_server_request_response_many_inline=889050
+      - MAX_ALLOCS_ALLOWED_client_server_request_response_no_promise_on_channel_creation=253050
+      - MAX_ALLOCS_ALLOWED_client_server_request_response_no_promise_on_channel_creation_inline=251050
+      - MAX_ALLOCS_ALLOWED_client_server_request_response_no_promise_on_channel_creation_many=1198050
+      - MAX_ALLOCS_ALLOWED_client_server_request_response_no_promise_on_channel_creation_many_inline=896050
```
glbrntt, line 44: Rather than duping all of these tests wouldn't it make more sense to have a variant of one of 'create_client_stream_channel_*'?

gjcairo: Yeah you're right that makes more sense. Just changed it.

glbrntt (approved):

glbrntt:
> Instead of adding more of those "old" tests can we please adopt the new benchmarking. We have to migrate over anyways so we should use chances like this to start the move.

I generally agree but in this instance we're just modifying an existing test so the overhead is low. 

I think it makes more sense to get the scaffolding for new benchmarks done across all of our repos in one go and then migrate the tests over time. Bootstrapping the setup is always more painful than you think it's going to be...

PR #455, opened 2024-08-21
Accept PING frames even when fully quiesced

clintonpi:
Motivation:

When the connection is fully quiesced, an endpoint responds with a connection error when it receives or tries to respond to a PING frame. This behaviour is unnecessary.

Modifications:

- Adjust `HTTP2ConnectionStateMachine` to successfully receive and send PING frames even when the connection is fully quiesced.

Result:

When the connection is fully quiesced, an endpoint will treat receipt and sending of a PING frame as it would when the connection is active.

### DIFF
--- a/Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
+++ b/Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
@@ -864,6 +864,42 @@ class SimpleClientServerTests: XCTestCase {
             XCTAssertTrue(error is NIOHTTP2Errors.UnableToParseFrame)
         }
     }
+
+    @available(*, deprecated, message: "Deprecated so deprecated functionality can be tested without warnings")
+    func testSuccessfullyReceiveAndSendPingEvenWhenConnectionIsFullyQuiesced() throws {
+        let serverHandler = FrameRecorderHandler()
+        try self.basicHTTP2Connection() { channel, streamID in
+            return channel.pipeline.addHandler(serverHandler)
+        }
+
+        let goAwayFrame = HTTP2Frame(streamID: .rootStream, payload: .goAway(lastStreamID: .rootStream, errorCode: .noError, opaqueData: nil))
+
+        // Fully quiesce the connection on the server.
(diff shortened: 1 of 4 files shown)


clintonpi:
Oops. Yes, of course.

Tests/NIOHTTP2Tests/SimpleClientServerTests.swift
```diff
@@ -864,6 +864,42 @@ class SimpleClientServerTests: XCTestCase {
             XCTAssertTrue(error is NIOHTTP2Errors.UnableToParseFrame)
         }
     }
+
+    @available(*, deprecated, message: "Deprecated so deprecated functionality can be tested without warnings")
+    func testSuccessfullyReceiveAndSendPingEvenWhenConnectionIsFullyQuiesced() throws {
```
glbrntt, line 869: I'm afraid you've fallen into one of the many deprecated-API traps 🪤

This whole test class is deprecated because it's testing deprecated APIs. There's an equivalent set of tests in  `SimpleClientServerInlineStreamMultiplexerTests` where this test should live. I think in this instance your test should just work without modifications.

--- 

An aside on the deprecated APIs. The original APIs for multiplexing created child channels for streams where you'd read and write `HTTP2Frame`s. 

Each `HTTP2Frame` holds a stream ID and its payload (headers, data etc.) However, when you open a stream the connection state machine wouldn't know about the stream until you did the first write on that stream. 

This means that a client could create stream 1 and then stream 3 but write on stream 3 first. A subsequent write on stream 1 would be a protocol violation (new streams must have higher IDs than previously created streams).

To fix this we added new stream creation APIs where the child channels operate on `HTTP2Frame.FramePayload`s and only when they do their first write do they get assigned a stream ID.

clintonpi: Ah, that makes sense. Thanks.

glbrntt (approved):
This LGTM, thanks @clintonpi!

Lukasa:

PR #457, opened 2024-08-29
Handle race condition between sending GOAWAY and HEADERS

clintonpi:
Motivation:

A potential race exists between a server sending a GOAWAY frame and a client opening a new stream (i.e., sending a HEADERS frame before receiving an already sent GOAWAY frame). If there are no open streams when the server sends the GOAWAY frame, the connection state on the server transitions to being fully quiesced, which throws a connection error on receipt of a HEADERS frame.

Modifications:

- Adjust the connection state machine on the server to throw a stream error (instead of a connection error), and consequently, send a RST_STREAM frame sent to the client.

Result:

This condition will be treated as a stream-level error instead.

### DIFF
--- a/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
+++ b/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
@@ -704,7 +704,7 @@ extension HTTP2ConnectionStateMachine {
             return .init(result: .connectionError(underlyingError: NIOHTTP2Errors.missingPreface(), type: .protocolError), effect: nil)
 
         case .fullyQuiesced:
-            return .init(result: .connectionError(underlyingError: NIOHTTP2Errors.ioOnClosedConnection(), type: .protocolError), effect: nil)
+            return .init(result: .streamError(streamID: streamID, underlyingError: NIOHTTP2Errors.ioOnClosedConnection(), type: .refusedStream), effect: nil)
 
         case .modifying:
             preconditionFailure("Must not be left in modifying state")
@@ -1042,7 +1042,7 @@ extension HTTP2ConnectionStateMachine {
             return .init(result: .connectionError(underlyingError: NIOHTTP2Errors.missingPreface(), type: .protocolError), effect: nil)
 
         case .fullyQuiesced:
-            return .init(result: .connectionError(underlyingError: NIOHTTP2Errors.ioOnClosedConnection(), type: .protocolError), effect: nil)
+            return .init(result: .succeed, effect: nil)
 
         case .modifying:
             preconditionFailure("Must not be left in modifying state")
--- a/Tests/NIOHTTP2Tests/SimpleClientServerInlineStreamMultiplexerTests.swift
+++ b/Tests/NIOHTTP2Tests/SimpleClientServerInlineStreamMultiplexerTests.swift
@@ -328,4 +328,48 @@ class SimpleClientServerInlineStreamMultiplexerTests: XCTestCase {
         promise.succeed(())
         stream.writeAndFlush(headerPayload, promise: promise)
     }
+
+    func testOpenStreamBeforeReceivingAlreadySentGoAway() throws {
+        let serverHandler = InboundFramePayloadRecorder()
+        try self.basicHTTP2Connection() { channel in
+            return channel.pipeline.addHandler(serverHandler)
+        }
+
+        let clientHandler = InboundFramePayloadRecorder()
+        let childChannelPromise = self.clientChannel.eventLoop.makePromise(of: Channel.self)
+        let multiplexer = try (self.clientChannel.pipeline.context(handlerType: NIOHTTP2Handler.self).wait().handler as! NIOHTTP2Handler).multiplexer.wait()
+        multiplexer.createStreamChannel(promise: childChannelPromise) { channel in
+            return channel.pipeline.addHandler(clientHandler)
+        }
+        (self.clientChannel.eventLoop as! EmbeddedEventLoop).run()
+        let childChannel = try childChannelPromise.futureResult.wait()
+
+        // Server sends GOAWAY frame.
+        let goAwayFrame = HTTP2Frame(streamID: .rootStream, payload: .goAway(lastStreamID: .maxID, errorCode: .noError, opaqueData: nil))
+        serverChannel.writeAndFlush(goAwayFrame, promise: nil)
+
+        // Client sends headers.
+        let headers = HPACKHeaders([(":path", "/"), (":method", "POST"), (":scheme", "https"), (":authority", "localhost")])
+        let reqFramePayload = HTTP2Frame.FramePayload.headers(.init(headers: headers))
+        childChannel.writeAndFlush(reqFramePayload, promise: nil)
+
+        self.interactInMemoryExpectingErrors(self.clientChannel, self.serverChannel) { error in
+            XCTAssert(error is NIOHTTP2Errors.IOOnClosedConnection)
+        }
+
+        // Client receives GOAWAY and RST_STREAM frames.
+        try self.clientChannel.assertReceivedFrame().assertGoAwayFrame(lastStreamID: .maxID, errorCode: 0, opaqueData: nil)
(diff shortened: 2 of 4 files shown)


Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStateMachine.swift
```diff
@@ -1042,7 +1042,7 @@ extension HTTP2ConnectionStateMachine {
             return .init(result: .connectionError(underlyingError: NIOHTTP2Errors.missingPreface(), type: .protocolError), effect: nil)
 
         case .fullyQuiesced:
-            return .init(result: .connectionError(underlyingError: NIOHTTP2Errors.ioOnClosedConnection(), type: .protocolError), effect: nil)
+            return .init(result: .succeed, effect: nil)
```
glbrntt, line 1045: Can you add a brief code comment on why we allow RST_STREAM frames to be sent when in this state?

Tests/NIOHTTP2Tests/SimpleClientServerInlineStreamMultiplexerTests.swift
```diff
@@ -329,12 +329,55 @@ class SimpleClientServerInlineStreamMultiplexerTests: XCTestCase {
...
-    func testSuccessfullyReceiveAndSendPingEvenWhenConnectionIsFullyQuiesced() throws {
+    func testOpenStreamBeforeReceivingAlreadySentGoAway() throws {
         let serverHandler = InboundFramePayloadRecorder()
         try self.basicHTTP2Connection() { channel in
             return channel.pipeline.addHandler(serverHandler)
         }
 
+        let clientHandler = InboundFramePayloadRecorder()
+        let childChannelPromise = self.clientChannel.eventLoop.makePromise(of: Channel.self)
+        let multiplexer = try (self.clientChannel.pipeline.context(handlerType: NIOHTTP2Handler.self).wait().handler as! NIOHTTP2Handler).multiplexer.wait()
```
glbrntt, line 340: This can be simplified a lot (I'm guessing the tests were written before this API was added) by using the `handler(type:)` API (see https://swiftpackageindex.com/apple/swift-nio/2.70.0/documentation/niocore/channelpipeline/handler(type:))

Tests/NIOHTTP2Tests/SimpleClientServerInlineStreamMultiplexerTests.swift
```diff
@@ -329,12 +329,55 @@ class SimpleClientServerInlineStreamMultiplexerTests: XCTestCase {
...
+        multiplexer.createStreamChannel(promise: childChannelPromise) { channel in
+            return channel.pipeline.addHandler(clientHandler)
+        }
+        (self.clientChannel.eventLoop as! EmbeddedEventLoop).run()
```
glbrntt, line 344: This can also be simplified:

```suggestion
        self.clientChannel.embeddedEventLoop.run()
```

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -63,6 +63,60 @@ extension XCTestCase {
...
+
+        repeat {
+            operated = false
+
+            if let data = readBytesFromChannel(first) {
+                operated = true
+
+                do {
+                    try second.writeInbound(data)
+                } catch(let error) {
```
glbrntt, line 98: No need for the `let error`, it's implicitly captured.

```suggestion
                } catch {
```

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -63,6 +63,60 @@ extension XCTestCase {
...
+                    encounteredError = true
+                    errorHandler(error)
+                }
+            }
+            if let data = readBytesFromChannel(second) {
+                operated = true
+
+                do {
+                    try first.writeInbound(data)
+                } catch(let error) {
```
glbrntt, line 108: Same here

Tests/NIOHTTP2Tests/TestUtilities.swift
```diff
@@ -63,6 +63,60 @@ extension XCTestCase {
         } while operated
     }
 
+    /// Have two `EmbeddedChannel` objects send and receive data from each and throw errors other until
+    /// they make no forward progress.
+    func interactInMemoryExpectingErrors(
```
glbrntt, line 68: There's a lot of duplication here. Can we remove some of it?

We could (probably) have a single function:

```swift
func interactInMemory(
  _ first: EmbeddedChannel,
  _ second: EmbeddedChannel,
  expectError: Bool = false,
  file: StaticString = #filePath,
  line: Uint = #line,
  _ errorHandler: (_ error: any Error) -> Void = { _ in }
) {
  // ...
}
```

glbrntt (changes requested):
This looks good aside from a few nits!

Tests/NIOHTTP2Tests/SimpleClientServerInlineStreamMultiplexerTests.swift
```diff
@@ -329,12 +329,55 @@ class SimpleClientServerInlineStreamMultiplexerTests: XCTestCase {
...
+        let goAwayFrame = HTTP2Frame(streamID: .rootStream, payload: .goAway(lastStreamID: .maxID, errorCode: .noError, opaqueData: nil))
+        serverChannel.writeAndFlush(goAwayFrame, promise: nil)
+
+        // Client sends headers.
+        let headers = HPACKHeaders([(":path", "/"), (":method", "POST"), (":scheme", "https"), (":authority", "localhost")])
+        let reqFramePayload = HTTP2Frame.FramePayload.headers(.init(headers: headers))
+        childChannel.writeAndFlush(reqFramePayload, promise: nil)
+
+        self.interactInMemory(self.clientChannel, self.serverChannel, expectError: true) { error in
+            XCTAssert(error is NIOHTTP2Errors.IOOnClosedConnection)
```
glbrntt, line 357: The original problem was that the race between the GOAWAY and HEADERS resulted in a connection error on the server. 

While we're now sending a RST_STREAM frame to the client, we're still propagating a connection-level error on the server. 

I think this the server propagate a `StreamError` which uses `CreatedStreamAfterGoaway` as its `baseError` and we should be catching that here.

glbrntt (changes requested):

Tests/NIOHTTP2Tests/SimpleClientServerInlineStreamMultiplexerTests.swift
clintonpi: Oh yeah, that makes sense. I didn't want to assert that a GOAWAY had been sent, due to this statement: "[A GOAWAY frame might not immediately precede closing of the connection](https://www.rfc-editor.org/rfc/rfc9113.html#section-6.8-7)". But I can see that I mistook a fully quiesced connection for a closed one.

glbrntt (changes requested):

Tests/NIOHTTP2Tests/SimpleClientServerInlineStreamMultiplexerTests.swift
```diff
@@ -329,12 +329,59 @@ class SimpleClientServerInlineStreamMultiplexerTests: XCTestCase {
...
+        // Client sends headers.
+        let headers = HPACKHeaders([(":path", "/"), (":method", "POST"), (":scheme", "https"), (":authority", "localhost")])
+        let reqFramePayload = HTTP2Frame.FramePayload.headers(.init(headers: headers))
+        childChannel.writeAndFlush(reqFramePayload, promise: nil)
+
+        self.interactInMemory(self.clientChannel, self.serverChannel, expectError: true) { error in
+            if let error = error as? NIOHTTP2Errors.StreamError {
+                XCTAssert(error.baseError is NIOHTTP2Errors.CreatedStreamAfterGoaway)
+            } else {
+                XCTFail("Expected error to be of type StreamError, got error of type \(type(of: error)).", file: #file, line: #line)
```
glbrntt, line 360: You don't need to set `#file` or `#line`, they're defaulted

```suggestion
                XCTFail("Expected error to be of type StreamError, got error of type \(type(of: error)).")
```

glbrntt (changes requested):
One nit but looks good otherwise!

glbrntt (approved):
Great, thank you!

PR #474, opened 2024-10-29
Enable python lint in CI

PeterAdams-A:
Motivation:

More checking is better

Modifications:

Enable and configure flake8

Result:

More CI checks

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,12 @@
+[flake8]
+
+ignore =
+    # These are needed to make our license headers pass the linting
+    E265,
+    E266,
+
+# 10% larger than the standard 80 character limit. Conforms to the black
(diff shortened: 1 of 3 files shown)


.flake8
```diff
@@ -0,0 +1,12 @@
...
+ignore =
+    # These are needed to make our license headers pass the linting
+    E265,
+    E266,
+
+# 10% larger than the standard 80 character limit. Conforms to the black
+# standard and Bugbear's B950.
+max-line-length = 88
+exclude =
+    Tests/hpack-test-case
```
PeterAdams-A, line 12: This code appears to have been cloned from another GitHub repo which we don't own - feels valuable not to diverge.

FranzBusch (approved):

PR #448, opened 2024-07-12
Use Bionic module from new Android overlay in Swift 6 instead

finagolfin:
Swift 6 has switched to this new Android overlay, swiftlang/swift#74758, which I've been testing on this repo without a problem using my daily Android CI for weeks now, finagolfin/swift-android-sdk#151.

PR #479, opened 2024-10-30
Import new Android overlay instead in the tests

finagolfin:
I added the `Bionic` import this summer in #448 and the compiler accepted it then, but it started erroring a couple months later without this full Android overlay, so add that instead.


Lukasa (approved):

Lukasa:
Thanks @finagolfin!

PR #480, opened 2024-11-01
Make configureHTTP2AsyncSecureUpgrade public

adam-fowler:
Minor PR to make `configureHTTP2AsyncSecureUpgrade ` helper function public


adam-fowler:
> @adam-fowler did you still want / need this change? (Apologies for missing this until now.)

Yes please. I currently have a copy of this function in Hummingbird. It would be preferable to use the original from swift-nio-http2.

glbrntt (approved):

PR #481, opened 2024-11-06
Better handle new streams when server is quiescing

glbrntt:
Motivation:

When the server is in the locally quiescing state and a client attempts to open a new stream, the connection state machine treats receiving the headers as a connection error. This results in the connection being closed and in-flight requests being failed.

This can happen because of the inherent race between the server sending a GOAWAY frame and the client receiving it, during which time the client can open a stream not knowing it's doomed to failure.

Rather than the server treating this as a connection error, it should reject the stream with a RST_STREAM frame and emit a stream error.

Modifications:

- Emit a stream error when receiving HEADERS in the locally quiesced state

Result:

Better handling of new streams when the server is quiescing

### DIFF
--- a/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
+++ b/Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
@@ -181,6 +181,7 @@ struct ConnectionStreamState {
         streamID: HTTP2StreamID,
         ignoreRecentlyReset: Bool,
         ignoreClosed: Bool = false,
+        isQuiescing: Bool = false,
         _ modifier: (inout HTTP2StreamStateMachine) -> StateMachineResultWithStreamEffect
     ) -> StateMachineResultWithStreamEffect {
         guard let result = self.activeStreams.autoClosingTransform(streamID: streamID, modifier) else {
@@ -188,7 +189,8 @@ struct ConnectionStreamState {
                 result: self.streamMissing(
                     streamID: streamID,
                     ignoreRecentlyReset: ignoreRecentlyReset,
-                    ignoreClosed: ignoreClosed
+                    ignoreClosed: ignoreClosed,
+                    isQuiescing: isQuiescing
                 ),
                 effect: nil
             )
@@ -218,11 +220,28 @@ struct ConnectionStreamState {
         _ modifier: (inout HTTP2StreamStateMachine) -> StateMachineResultWithStreamEffect
     ) -> StateMachineResultWithStreamEffect {
         guard let result = self.activeStreams.autoClosingTransform(streamID: streamID, modifier) else {
-            // We never ignore recently reset streams here, as this should only ever be used when *sending* frames.
-            return StateMachineResultWithStreamEffect(
-                result: self.streamMissing(streamID: streamID, ignoreRecentlyReset: false, ignoreClosed: false),
-                effect: nil
-            )
+            // This state can be reached when a RST_STREAM frame is sent by this peer but the stream
+            // doesn't exist. This can happen for a few reasons, including:
+            //
+            // 1. The RST_STREAM frame is being sent because the stream wasn't accepted (i.e. the
+            //    client opening the stream raced with the server sending a GOAWAY frame). In this
+            //    case the state machine doesn't know about the stream but does need to send
+            //    a RST_STREAM frame.
+            // 2. An implementation or user error where the stream genuinely doesn't exist and
+            //    attempting to send a RST_STREAM frame is an error.
+            if streamID.isClientInitiated, streamID > self.lastClientStreamID {
+                return StateMachineResultWithStreamEffect(result: .succeed, effect: nil)
+            } else {
+                return StateMachineResultWithStreamEffect(
+                    result: self.streamMissing(
+                        streamID: streamID,
+                        ignoreRecentlyReset: false,
+                        ignoreClosed: false,
+                        isQuiescing: true
+                    ),
+                    effect: nil
+                )
+            }
         }
 
         guard let effect = result.effect, effect.closedStream else {
@@ -324,7 +343,8 @@ struct ConnectionStreamState {
     private func streamMissing(
         streamID: HTTP2StreamID,
         ignoreRecentlyReset: Bool,
-        ignoreClosed: Bool
+        ignoreClosed: Bool,
+        isQuiescing: Bool
     ) -> StateMachineResult {
         if ignoreRecentlyReset && self.recentlyResetStreams.contains(streamID) {
             return .ignoreFrame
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
```diff
@@ -218,11 +220,28 @@ struct ConnectionStreamState {
...
+            // This state can be reached when a RST_STREAM frame is sent by this peer but the stream
+            // doesn't exist. This can happen for a few reasons, including:
+            //
+            // 1. The RST_STREAM frame is being sent because the stream wasn't accepted (i.e. the
+            //    client opening the stream raced with the server sending a GOAWAY frame). In this
+            //    case the state machine doesn't know about the stream but does need to send
+            //    a RST_STREAM frame.
+            // 2. An implementation or user error where the stream genuinely doesn't exist and
+            //    attempting to send a RST_STREAM frame is an error.
+            if streamID.isClientInitiated, streamID > self.lastClientStreamID {
```
Lukasa, line 232: This check seems overly specific. Do we think it's not possible to get to this via PUSH_PROMISE?

Lukasa: Oh, isn't this also asking the wrong question? This would also satisfy if _we_ were the client, which seems wrong.

Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
```diff
@@ -218,11 +220,28 @@ struct ConnectionStreamState {
...
+                return StateMachineResultWithStreamEffect(result: .succeed, effect: nil)
+            } else {
+                return StateMachineResultWithStreamEffect(
+                    result: self.streamMissing(
+                        streamID: streamID,
+                        ignoreRecentlyReset: false,
+                        ignoreClosed: false,
+                        isQuiescing: true
```
Lukasa, line 240: I'm also a bit suspicious about this. We can't always be quiescing.

Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
```diff
@@ -333,11 +353,20 @@ struct ConnectionStreamState {
         switch streamID.mayBeInitiatedBy(.client) {
         case true where streamID > self.lastClientStreamID,
             false where streamID > self.lastServerStreamID:
-            // The stream in question is idle.
-            return .connectionError(
-                underlyingError: NIOHTTP2Errors.noSuchStream(streamID: streamID),
-                type: .protocolError
-            )
+            if isQuiescing {
```
Lukasa, line 356: I think this flag isn't quite specific enough. I'm not even entirely sure we want to sink this logic into this function: we really only want this flow to happen on receipt of stream creation frames, which might suggest a better name for the parameter at least.

Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
```diff
@@ -324,7 +343,8 @@ struct ConnectionStreamState {
     private func streamMissing(
         streamID: HTTP2StreamID,
         ignoreRecentlyReset: Bool,
-        ignoreClosed: Bool
+        ignoreClosed: Bool,
+        isQuiescing: Bool
```
Lukasa, line 347: Can we rename this to be `isLocallyQuiescing`?

Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
glbrntt: To be honest I wasn't really happy with this whole approach. It's all a bit awkward.

The issue is that we receive HEADERS in a state where it's valid to receive them if the stream already exists but is a stream error if it doesn't. The problem with is that we return the stream error out of the state machine and back into the channel handler, which will then call back into the state machine to send the RST_STREAM frame as a result of the stream error. From the POV of the state machine, that stream doesn't exist, and that's the awkward bit.

It makes me wonder whether we should modify `StateMachineResult` so that we can represent the notion of rejecting a stream (as opposed to a stream error, for which we assume the stream already exists). WDYT?

Lukasa: Yeah, that's an interesting issue. Perhaps the actual fix is that we should tolerate sending RST_STREAM frames on streams that don't exist?

glbrntt: I initially ruled that out because there are tests which validate that you can't reset a stream twice (or something similar). This would certainly be the simplest fix assuming we're okay with that small regression.

Lukasa: I'm open to that small regression, yeah.

Sources/NIOHTTP2/ConnectionStateMachine/ConnectionStreamsState.swift
```diff
@@ -333,11 +336,22 @@ struct ConnectionStreamState {
         switch streamID.mayBeInitiatedBy(.client) {
         case true where streamID > self.lastClientStreamID,
             false where streamID > self.lastServerStreamID:
-            // The stream in question is idle.
-            return .connectionError(
-                underlyingError: NIOHTTP2Errors.noSuchStream(streamID: streamID),
-                type: .protocolError
-            )
+            if isLocallyQuiescing {
```
Lukasa, line 339: Ok so this is looking a lot better. I wonder now only if we need some amount of DoS protection. Presumably we shouldn't tolerate this happening too often. Can we record some counter or something that lets us produce a rate limit elsewhere?

glbrntt: This gets ... interesting. 

One approach is to re-use the rate limiter from `DOSHeuristics` within the state machine and check the rate for this specific scenario. This would mean the state machine would need to understand time and would likely become generic over `NIODeadlineClock`. I don't like that at all!

Another approach would be to widen the scope of the DOS scenario we're considering and apply rate limiting in the channel handler as we currently do. I think the appropriate thing to rate limit here is "inbound stream errors". WDYT?

Sources/NIOHTTP2/DOSHeuristics.swift
```diff
@@ -28,12 +28,15 @@ struct DOSHeuristics<DeadlineClock: NIODeadlineClock> {
     /// The maximum number of "empty" data frames we're willing to tolerate.
     private let maximumSequentialEmptyDataFrames: Int
 
-    private var resetFrameRateControlStateMachine: HTTP2ResetFrameRateControlStateMachine
+    private var resetFrameRateControlStateMachine: RateLimitStateMachine
+    private var streamErrorRateControlStateMachine: RateLimitStateMachine
```
glbrntt, line 32: Because the state machine reserves capacity on `init` we end up regressing allocations per-connection here; given it's not per-stream I think that's okay.

PR #483, opened 2024-11-14
Finish multiplexer's inbound streams in more cases

gjcairo:
## Motivation
Currently, the multiplexer's inbound streams stream is finished only when the channel becomes inactive.
There are some scenarios in which the channel may be closed before it has a chance to become active, and the stream will never be finished. This can cause any users iterating over the stream to hang.

## Modifications
This PR finishes the inbound streams stream when a connection error is fired, and when the handler is removed.

## Result
Fewer bugs.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
@@ -214,26 +214,10 @@ extension HTTP2CommonInboundStreamMultiplexer {
     }
 
     internal func propagateErrorCaught(_ error: any Error) {
-        for channel in self.streams.values {
-            channel.receiveStreamClosed(nil)
-        }
-
-        for channel in self._pendingStreams.values {
-            channel.receiveStreamClosed(nil)
-        }
-
         self.streamChannelContinuation?.finish(throwing: error)
     }
 
     internal func propagateHandlerRemoved() {
-        for channel in self.streams.values {
-            channel.receiveStreamClosed(nil)
-        }
-
-        for channel in self._pendingStreams.values {
-            channel.receiveStreamClosed(nil)
-        }
-
         self.streamChannelContinuation?.finish()
     }
 
--- a/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
@@ -213,6 +213,30 @@ extension HTTP2CommonInboundStreamMultiplexer {
         self.streamChannelContinuation?.finish()
     }
 
+    internal func propagateErrorCaught(_ error: any Error) {
+        for channel in self.streams.values {
+            channel.receiveStreamClosed(nil)
+        }
+
+        for channel in self._pendingStreams.values {
+            channel.receiveStreamClosed(nil)
+        }
+
+        self.streamChannelContinuation?.finish(throwing: error)
+    }
+
+    internal func propagateHandlerRemoved() {
+        for channel in self.streams.values {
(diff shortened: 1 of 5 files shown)


Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
```diff
@@ -213,6 +213,30 @@ extension HTTP2CommonInboundStreamMultiplexer {
...
+        for channel in self.streams.values {
+            channel.receiveStreamClosed(nil)
+        }
+
+        for channel in self._pendingStreams.values {
+            channel.receiveStreamClosed(nil)
+        }
+
+        self.streamChannelContinuation?.finish()
+    }
```
glbrntt, line 238: I don't think we should be closing the streams in these cases, we should just finish the continuation

glbrntt (changes requested):
Can we add a test for this too?

Sources/NIOHTTP2/HTTP2CommonInboundStreamMultiplexer.swift
gjcairo: For `errorCaught` I think you're right (although I'm not sure what it would mean if there was a connection error fired down the pipeline but we keep the streams opened?); however if the handler was removed, wouldn't it make sense to close the streams? Is there anything we can do with them if the connection has died?

glbrntt: In the `HTTP2StreamMultiplexer`, handler removed calls `clearDidReadChannels` in the common (i.e. this) multiplexer: https://github.com/apple/swift-nio-http2/blob/91136a6604e58e4f1e6cab24086b876fe01bc24c/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift#L43-L46

I don't think we should have divergent behaviour between the two. When we reach `handlerRemoved` the streams will have already been closed.

---

For error caught, the `HTTP2StreamMultiplexer` only propagates stream errors to the child channels, they then get to decide how to act on those errors. Any connection level errors aren't propagated to streams and go down the channel pipeline and another handler decides how to act on it (usually by closing the connection, which will result in the streams being closed), but we should avoid short circuiting that because not all errors result in the connection being closed.

gjcairo: Thanks George, that's clear. I'll make the change.

Tests/NIOHTTP2Tests/SimpleClientServerInlineStreamMultiplexerTests.swift
```diff
@@ -503,4 +503,67 @@ class SimpleClientServerInlineStreamMultiplexerTests: XCTestCase {
...
+        stream.writeAndFlush(frame, promise: nil)
+        try await self.interactInMemory(asyncClientChannel, asyncServerChannel)
+
+        // Fire an error down the server pipeline: it should cause the inbound stream to finish with error
+        asyncServerChannel.pipeline.fireErrorCaught(ErrorCaughtPropagated.error)
+        do {
+            for try await _ in serverMultiplexer.inbound {}
+            XCTFail("Expected error to be thrown")
+        } catch {
+            XCTAssertEqual(error as! ErrorCaughtPropagated, .error)
```
glbrntt, line 566: FWIW in this case you can just use `is` to test whether it's the expected type. This also avoids trapping in case the error isn't as expected:

```suggestion
            XCTAssert(error is ErrorCaughtPropagated)
```

 `XCTAssert(error is ErrorCaughtPropagated)`

glbrntt (approved):
One nit, but looks good otherwise, thanks @gjcairo

PR #487, opened 2024-12-13
Don't fail close promise when an error occurs on closing

gjcairo:
Strictly speaking, a `close` operation can't really fail (except when the closing mode isn't supported or the channel is already closed). Even if an error is encountered while closing a channel, or if an error caused the channel to be closed, the end result is the same: the channel will be closed. 

In this case in particular, when streams are closed from the client (i.e. by sending a RST_STREAM frame) the `HTTP2StreamChannel` would fail the close promise. This isn't appropriate however, as the channel is successfully closed: the promise failure is only giving additional information that can actually be exposed via other means.

This change stops failing both the stream channel's `closePromise` and the promise passed to the `close` method in the `HTTP2StreamChannel`. Instead, if the close happens uncleanly, the error will be fired down the pipeline. Close promises will now only be succeeded.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamChannel.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamChannel.swift
@@ -676,7 +676,7 @@ final class HTTP2StreamChannel: Channel, ChannelCore, @unchecked Sendable {
 
         self.eventLoop.execute {
             self.removeHandlers(pipeline: self.pipeline)
-            self.closePromise.succeed(())
+            self.closePromise.succeed()
             if let streamID = self.streamID {
                 self.multiplexer.streamClosed(id: streamID)
             } else {
@@ -694,14 +694,14 @@ final class HTTP2StreamChannel: Channel, ChannelCore, @unchecked Sendable {
         self.failPendingWrites(error: error)
         if let promise = self.pendingClosePromise {
             self.pendingClosePromise = nil
-            promise.fail(error)
+            promise.succeed()
         }
         self.pipeline.fireErrorCaught(error)
         self.pipeline.fireChannelInactive()
 
         self.eventLoop.execute {
             self.removeHandlers(pipeline: self.pipeline)
-            self.closePromise.fail(error)
+            self.closePromise.succeed()
             if let streamID = self.streamID {
                 self.multiplexer.streamClosed(id: streamID)
             } else {
--- a/Tests/NIOHTTP2Tests/ConfiguringPipelineInlineMultiplexerTests.swift
+++ b/Tests/NIOHTTP2Tests/ConfiguringPipelineInlineMultiplexerTests.swift
@@ -71,7 +71,10 @@ class ConfiguringPipelineInlineMultiplexerTests: XCTestCase {
             )
         )
 
-        clientMultiplexer.createStreamChannel(promise: nil) { channel in
+        let errorHandler = ErrorEncounteredHandler()
+        let streamChannelPromise = self.clientChannel.eventLoop.makePromise(of: Channel.self)
+        clientMultiplexer.createStreamChannel(promise: streamChannelPromise) { channel in
+            try? channel.pipeline.addHandler(errorHandler).wait()
             channel.writeAndFlush(reqFrame.payload).whenComplete { _ in channel.close(promise: requestPromise) }
             return channel.eventLoop.makeSucceededFuture(())
         }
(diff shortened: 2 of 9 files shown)


Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -694,14 +694,14 @@ final class HTTP2StreamChannel: Channel, ChannelCore, @unchecked Sendable {
         self.failPendingWrites(error: error)
         if let promise = self.pendingClosePromise {
             self.pendingClosePromise = nil
-            promise.fail(error)
+            promise.succeed()
```
glbrntt, line 697: I don't think this is what was agreed: IIRC the promise from `close(mode:promise:)` (i.e. `pendingClosePromise`)  _can_ fail to provide diagnostic information. It's the `closeFuture` (i.e. part of the `Channel` API) which mustn't fail.

Tests/NIOHTTP2Tests/ConfiguringPipelineInlineMultiplexerTests.swift
```diff
@@ -71,7 +71,10 @@ class ConfiguringPipelineInlineMultiplexerTests: XCTestCase {
             )
         )
 
-        clientMultiplexer.createStreamChannel(promise: nil) { channel in
+        let errorHandler = ErrorEncounteredHandler()
+        let streamChannelPromise = self.clientChannel.eventLoop.makePromise(of: Channel.self)
+        clientMultiplexer.createStreamChannel(promise: streamChannelPromise) { channel in
+            try? channel.pipeline.addHandler(errorHandler).wait()
```
glbrntt, line 77: This stream channel initializer will be executed on the event-loop; you can't `wait()` while on an EL. You need to use an alternative API, like the sync operations.

glbrntt (changes requested):

Sources/NIOHTTP2/HTTP2StreamChannel.swift
FranzBusch: Agree with what @glbrntt said here. This means we need to change the implementation of `executeThenClose` to discard the error from `close` and instead add a call to `closeFuture.get()`

glbrntt: And update [the docs](https://github.com/apple/swift-nio/blob/ecfaa2c6ff3234bdf7909eba0da01c245922e471/Sources/NIOCore/Channel.swift#L109) for `Channel.closeFuture`

gjcairo: Oh okay, I misunderstood this. The one thing I found weird when implementing this (and why I decided to not fail this promise either) is that calling `channel.close()` will return a failed promise (since it creates one and passes it to `channel.close(promise: newPromise)` under the hood), which I personally find slightly confusing. But if you're okay with this behaviour I can just be clear in the docs about it.

glbrntt: That's expected. Roughly speaking:
- `channel.close(promise:)` (and variants of, i.e. `channel.close()`) are "close the channel and tell me _how_ it closed" (cleanly/uncleanly) 
- `channel.closeFuture` is "tell me _when_ the channel is closed"

glbrntt (changes requested):

Tests/NIOHTTP2Tests/ConfiguringPipelineInlineMultiplexerTests.swift
gjcairo: > This one got marked as resolved but wasn't actually resolved.
Ah, sorry I fixed it in a bunch of places but missed a few others.

> FWIW I also don't think it's necessary for all of these tests to check this behaviour [..] I'd prefer we had a separate test which asserts the specific behaviour.
We were already performing assertions on the `close` promise everywhere - you could use the same argument there. I thought that it was more consistent if we are going to assert something on the `close` promise that we also check the right behaviour is seen for the `closeFuture` and the errors being thrown downstream. But I can revert those changes if you disagree and add a single test.

glbrntt: No worries, I saw the first one and assumed they'd all been missed.

I'm okay not reverting it, the work's been done now. I get the following the existing pattern argument; I just think we should be mindful when we make changes like this to keep tests maintainable. This package is bad enough as it is as most tests have about three different versions because of the different multiplexing APIs...

glbrntt (approved):

PR #488, opened 2024-12-16
support stream resets through user events

ehaydenr:
Resetting streams with specific error codes is required by some applications such as those implementing the CONNECT method (see https://datatracker.ietf.org/doc/html/rfc9113#section-8.5-8). `ChannelOutboundInvoker.close` doesn't allow for closing channels while specifying an error code. To get around this, this change defines an event `HTTP2StreamChannelEvent` which allows applications to close the channel with additional detail.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamChannel.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamChannel.swift
@@ -605,6 +605,10 @@ final class HTTP2StreamChannel: Channel, ChannelCore, @unchecked Sendable {
     }
 
     public func close0(error: Error, mode: CloseMode, promise: EventLoopPromise<Void>?) {
+        self.closeWithCode(code: .cancel, promise: promise)
+    }
+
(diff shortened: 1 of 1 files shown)


Sources/NIOHTTP2/HTTP2StreamChannel.swift
```diff
@@ -998,3 +1007,9 @@ extension HTTP2StreamChannel {
         SynchronousOptions(channel: self)
     }
 }
+
+/// Events that can be sent by the application to be handled by the `HTTP2StreamChannel`
+public enum HTTP2StreamChannelEvent {
```
Lukasa, line 1012: So I think we'll probably end up regressing making this a public enum, as we can't easily evolve that as an API. I think we should store the "real" value in an enum, but have this wrapped in a struct that has a few `static func` constructors..

We'll also need tests for this.

ehaydenr:
closing in favor of https://github.com/apple/swift-nio-extras/pull/241

PR #501, opened 2025-03-19
Add Sendable conformance to HTTP2StreamMultiplexer to support Swift concurrency patterns

djmango:
This is required to use NIOHTTPClient in Swift 6 w/ Connect for gRPC.

For example

```swift
        let protocolClient = ProtocolClient(
            httpClient: NIOHTTPClient(host: host, port: port),
            config: ProtocolClientConfig(
                host: host,
                networkProtocol: .grpc,
                codec: ProtoCodec()
            )
        )
```

Co-Authored-By: Grok <noreply@grok.com>

### DIFF
--- a/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
+++ b/Sources/NIOHTTP2/HTTP2StreamMultiplexer.swift
@@ -22,7 +22,7 @@ import NIOCore
 /// on ``HTTP2Frame`` or ``HTTP2Frame/FramePayload`` objects as their base communication
 /// atom, as opposed to the regular NIO `SelectableChannel` objects which use `ByteBuffer`
 /// and `IOData`.
-public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboundHandler {
+public final class HTTP2StreamMultiplexer: ChannelInboundHandler, ChannelOutboundHandler, Sendable {
     public typealias InboundIn = HTTP2Frame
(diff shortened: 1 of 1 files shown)


djmango:
> Thanks for the PR! Unfortuantely this type isn't actually `Sendable`. I'm currently working on a patch to make this package compile cleanly with strict concurrency enabled which will include appropriately annotating all public types so you're probably better off waiting for that to land instead.

That's great, will subscribe, looking forward to that patch!

glbrntt:
It hasn't been released _yet_. We'll try to get a release out soon.

---

This error will remain though, the `HTTP2StreamMultiplexer` _isn't_ `Sendable`.

```
Type 'HTTP2StreamMultiplexer' does not conform to the 'Sendable' protocol
```

This is the relevant change https://github.com/apple/swift-nio-http2/pull/506 -- once released you'll need to use the `SendableView` on the multiplexer.

djmango:
> @djmango / @glbrntt is this fixed? Is there any workaround since I am facing above error while using `Connect` package.
> 
> Error Message: Type 'HTTP2StreamMultiplexer' does not conform to the 'Sendable' protocol

I've been using it in Swift 6 by just forcing it Sendable. Technically not, but so far hasn't caused any problems. Here are the a relevant forks https://github.com/djmango/swift-nio-http2 https://github.com/djmango/connect-swift

Lukasa:
What are you trying to actually do with the multiplexer that triggers the error?

PR #502, opened 2025-03-20
Strict concurrency for NIOHPACK

glbrntt:
Motivation:

To increase concurrency safety the NIOPHACK module should compile without warning under strict concurrency checking.

Modifications:

- Add explicit Sendable annotations

Result:

No concurrency warnings.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,13 +15,31 @@
 
 import PackageDescription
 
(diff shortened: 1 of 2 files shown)


Lukasa (approved):
Easy peasy.

PR #518, opened 2025-06-29
Add GlitchesMonitor

gjcairo:
Add a `GlitchesMonitor` that will terminate a connection if too many stream resets are triggered.

### DIFF
--- a/Sources/NIOHTTP2/GlitchesMonitor.swift
+++ b/Sources/NIOHTTP2/GlitchesMonitor.swift
@@ -1,0 +1,70 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2025 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+struct GlitchesMonitor {
+    static let defaultMaxGlitches: UInt = 200
+    private var stateMachine: GlitchesMonitorStateMachine
+
+    init(maxGlitches: UInt = GlitchesMonitor.defaultMaxGlitches) {
+        self.stateMachine = GlitchesMonitorStateMachine(maxGlitches: maxGlitches)
+    }
+
+    mutating func processStreamError() throws {
+        switch self.stateMachine.recordEvent() {
+        case .belowLimit:
+            ()
+
+        case .exceededLimit:
+            throw NIOHTTP2Errors.excessiveNumberOfGlitches()
+        }
+    }
+}
+
+extension GlitchesMonitor {
+    private struct GlitchesMonitorStateMachine {
+        enum State {
+            case monitoring(numberOfGlitches: UInt)
+            case glitchesExceeded
+        }
+
+        private var state: State
+        private let maxGlitches: UInt
+
+        init(maxGlitches: UInt) {
+            self.state = .monitoring(numberOfGlitches: 0)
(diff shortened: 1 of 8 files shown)


Sources/NIOHTTP2/GlitchesMonitor.swift
```diff
@@ -0,0 +1,70 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+struct GlitchesMonitor {
+    static let defaultMaxGlitches: UInt = 200
```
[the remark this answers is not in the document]
gjcairo: > and do we really want a UInt here?

Well my thought was this limit won't be negative so `UInt` made sense; do we always prefer `Int` generally? Or why are you asking?

gjcairo:
The failing allocation test seems to be a regression elsewhere in NIO (see https://github.com/apple/swift-nio-http2/pull/5190)

glbrntt (approved):

Sources/NIOHTTP2/GlitchesMonitor.swift
```diff
@@ -0,0 +1,71 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+package struct GlitchesMonitor {
```
Lukasa, line 15: Are we sure `package` visibility is a good idea here? It enables release mode unit tests, sure, but it also causes the optimizer to generate suboptimal code. Given that swift-nio-http2 uses `@testable import` extensively _already_, I don't think we gain much by avoiding it here.

Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -1386,6 +1445,7 @@ extension NIOHTTP2Handler {
         public var maximumBufferedControlFrames: Int = 10000
         public var maximumSequentialContinuationFrames: Int = NIOHTTP2Handler.defaultMaximumSequentialContinuationFrames
         public var maximumRecentlyResetStreams: Int = NIOHTTP2Handler.defaultMaximumRecentlyResetFrames
+        public var maximumConnectionGlitches: Int = GlitchesMonitor.defaultMaximumGlitches
```
Lukasa, line 1448: It would be good, I think, to include narrative docs that explain this field and what it does.

Sources/NIOHTTP2/HTTP2Error.swift
```diff
@@ -1832,6 +1844,31 @@ public enum NIOHTTP2Errors {
         }
     }
 
+    /// The remote peer has triggered too many stream errors on this connection.
```
Lukasa, line 1847: We might broaden the definition here in future, so let's just call it `glitches` instead of `stream errors`.

Sources/NIOHTTP2/GlitchesMonitor.swift
Lukasa: It's suboptimal because it inhibits some optimizations that the compiler can make by virtue of being able to observe all usage sites. An `internal` symbol or type, except when compiled with `-enable-testability`, has all usage sites visible in the code. This can allow strong effects to be inferred, some types to be deleted, etc.

`package` visibility doesn't do that: it's essentially the same visibility as `-enable-testability` would trigger, but all the time. Its saving grace is that it works equivalently well in both modes.

I think `package` should be preferred in contexts where running the unit tests in release mode is meaningful. I personally don't care much about those, but users who do should go for it. Otherwise, I'd recommend sticking with `@testable import`.

PR #522, opened 2025-07-17
Add a frame delegate

glbrntt:
Motivation:

Frames written from a child channel may not be written immediately by the connection channel. For example, a DATA frame written on a stream may be larger than the max frame size imposed on the connection and so the connection channel will have to slice it into multiple frames. Or the connection may not be writable and may delay the transmission of the frame. This behaviour isn't currently observable but is useful to know about.

Modifications:

- Add a frame delegate which is notified when certain frames are written by the connection channel.

Result:

Users can observer when headers and data frames are written into the connection channel.

### DIFF
--- a/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
+++ b/Sources/NIOHTTP2/HTTP2ChannelHandler.swift
@@ -113,6 +113,9 @@ public final class NIOHTTP2Handler: ChannelDuplexHandler {
     /// The maximum number of sequential CONTINUATION frames.
     private let maximumSequentialContinuationFrames: Int
 
+    /// A delegate which is told about frames which have eebn written.
+    private let frameDelegate: NIOHTTP2FrameDelegate?
+
     @usableFromInline
     internal var inboundStreamMultiplexer: InboundStreamMultiplexer? {
         self.inboundStreamMultiplexerState.multiplexer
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP2/HTTP2ChannelHandler.swift
```diff
@@ -1067,6 +1098,24 @@ extension NIOHTTP2Handler {
...
+            switch frame.payload {
+            case .headers(let headers):
+                delegate.wroteHeaders(headers.headers, endStream: headers.endStream, streamID: frame.streamID)
+
+            case .data(let data):
+                switch data.data {
+                case .byteBuffer(let buffer):
+                    delegate.wroteData(buffer, endStream: data.endStream, streamID: frame.streamID)
+                case .fileRegion:
+                    ()
```
rnro, line 1112: Why do we take no action here? If it's intended should it be documented that we don't notify for file regions?

Tests/NIOHTTP2Tests/SimpleClientServerFramePayloadStreamTests.swift
```diff
@@ -2844,6 +2848,72 @@ class SimpleClientServerFramePayloadStreamTests: XCTestCase {
         XCTAssertNoThrow(try self.clientChannel.finish())
         XCTAssertNoThrow(try self.serverChannel.finish())
     }
+
+    func testFrameDelegateIsCalledForDATAFrames() throws {
```
rnro, line 2852: This seems to also assert on header frames, is this name appropriate?

rnro (approved):

Lukasa (approved):

PR #551, opened 2026-06-30
HPACKDecoder: reject more than two table size updates per header block

sour-exploit:
RFC 7541 § 4.2 specifies that an encoder emits at most two dynamic table size updates between two header blocks (the smallest size reached, then the final size). The decoder enforced placement - updates must precede the first header - but put no upper bound on their count, so a block of arbitrarily many 0x20 bytes followed by a header was accepted. This is a conformance gap. It is not a DoS: setting an already-empty table to 0 is O(1).

Modifications:

- Count consecutive table size updates in decodeHeaders(from:) and throw IllegalDynamicTableSizeChange once a third is seen.
- Extend the existing § 4.2 comment to cite the at-most-two rule.
- Add tests covering the two-accepted and three-rejected cases.

Result:

Header blocks with more than two dynamic table size updates are rejected as malformed, matching RFC 7541 § 4.2.


glbrntt (approved):
Thanks, this looks great. The formatter is failing for an unrelated reason which will be fixed in https://github.com/apple/swift-nio-http2/pull/553

PR #552, opened 2026-06-30
NIOHTTP2: forbid keep-alive and upgrade connection-specific headers

### LINKED ISSUE #549: HPACKHeaders+Validation: add keep-alive and upgrade to connection-specific header blacklist (RFC 9113 §8.2.2)
### Summary
 
`HPACKHeaders+Validation.swift`'s `legalHeaderField(value:)` rejects connection-specific header fields per RFC 9113 §8.2.2, but the blacklist is incomplete.
 
Current code (`HPACKHeaders+Validation.swift:290-292`):
 
```swift
switch self.baseName {
case "connection", "transfer-encoding", "proxy-connection":
    throw NIOHTTP2Errors.forbiddenHeaderField(name: String(self.baseName), value: value)
default:
    return
}
```
 
RFC 9113 §8.2.2 names five connection-specific fields that a conformant HTTP/2 endpoint MUST treat as malformed:
 
> The `Connection`, `Proxy-Connection`, `Keep-Alive`, `Transfer-Encoding`, and `Upgrade` header fields
 
The current set blocks three of five. `keep-alive` and `upgrade` pass validation and are copied verbatim by `appendRegularHeaders` (HTTP2ToHTTP1Codec.swift:867-874) into the translated HTTP/1.1 message.
 
The existing comment at lines 276-278 indicates the set was intended to cover connection-specific fields ("A connection-specific header field technically is one that is listed in the Connection header, but could also be proxy-connection & transfer-encoding"), so this appears to be an oversight rather than a deliberate allowance. The comment still references RFC 7540; RFC 9113 §8.2.2 makes the enumeration explicit.
 
Note: `te` is already correctly handled separately at line 171 (restricted to `trailers`-only per §8.2.2).
 
### Proposed fix
 
```swift
case "connection", "transfer-encoding", "proxy-connection", "keep-alive", "upgrade":
    throw NIOHTTP2Errors.forbiddenHeaderField(name: String(self.baseName), value: value)
```
 
### Reproducer
 
An HTTP/2 request with `upgrade: websocket` and `keep-alive: timeout=5` headers is accepted by `HTTP2FramePayloadToHTTP1ServerCodec` and translated into an HTTP/1.1 request carrying both headers unmodified.
 
```swift
import NIOCore
import NIOEmbedded
import NIOHPACK
import NIOHTTP1
import XCTest
@testable import NIOHTTP2
 
final class ConnectionHeaderLeakTests: XCTestCase {
    func testUpgradeAndKeepAliveLeakThroughCodec() throws {
        let channel = EmbeddedChannel()
        try channel.pipeline.syncOperations.addHandler(HTTP2FramePayloadToHTTP1ServerCodec())
 
        let headers = HPACKHeaders([
            (":method", "GET"),
            (":path", "/"),
            (":scheme", "https"),
            (":authority", "example.com"),
            ("upgrade", "websocket"),
            ("keep-alive", "timeout=5, max=1000"),
        ])
 
        try channel.writeInbound(
            HTTP2Frame.FramePayload.headers(.init(headers: headers, endStream: true))
        )
 
        guard let part = try channel.readInbound(as: HTTPServerRequestPart.self),
              case .head(let reqHead) = part else {
            XCTFail("no request head emitted")
            return
        }
 
        // Both headers survive H2→H1 translation (should be rejected):
        XCTAssertNotNil(reqHead.headers["upgrade"].first)
        XCTAssertNotNil(reqHead.headers["keep-alive"].first)
    }
}
```
 
### Impact
 
Conformance / defense-in-depth. `Connection` is already blocked, so a bare `Upgrade` without `Connection: upgrade` is inert to most HTTP/1.1 parsers. Completing the §8.2.2 set prevents potential connection-state desync in downstream parsers that handle `Upgrade` or `Keep-Alive` independently.
 
### References
 
- RFC 9113 §8.2.2: https://www.rfc-editor.org/rfc/rfc9113#section-8.2.2

sour-exploit:
RFC 9113 § 8.2.2 enumerates five connection-specific header fields that a conformant HTTP/2 endpoint must treat as malformed: Connection, Proxy-Connection, Keep-Alive, Transfer-Encoding, and Upgrade. HeaderFieldName .legalHeaderField(value:) only rejected three of them, so keep-alive and upgrade passed validation and were translated verbatim into HTTP/1.1 messages by the H2->H1 codecs. This is a conformance gap and a defence-in-depth issue, since downstream HTTP/1.1 parsers may act on those fields independently.

Modifications:

- Add keep-alive and upgrade to the forbidden connection-specific header switch in legalHeaderField(value:).
- Update the surrounding comment to cite RFC 9113 § 8.2.2 and its explicit enumeration in place of the obsoleted RFC 7540 § 8.1.2.2 reference.
- Add request- and response-block validation tests covering all five connection-specific fields.

Result:

All five connection-specific header fields named by RFC 9113 § 8.2.2 are rejected during header-block validation.

Resolves #549


glbrntt (approved):
Thank you!

PR #554, opened 2026-07-17
NIOHTTP2: treat CR, LF and NUL in regular header field values as malformed

abhinav9-coder:
## Motivation

RFC 9113 § 8.2.1 requires that a field value must not contain NUL, LF or CR at any position, and that a message carrying one is treated as malformed. That applies to every field value, not just pseudo-headers.

`legalHeaderField(value:)` only enforces it for pseudo-headers. The regular-header branch checks the forbidden-name list and then returns without looking at the value:

```swift
switch self.baseName {
case "connection", "transfer-encoding", "proxy-connection", "keep-alive", "upgrade":
    throw NIOHTTP2Errors.forbiddenHeaderField(name: String(self.baseName), value: value)
default:
    return
}
```

Nothing else on the inbound path picks it up either. HPACK values are length-prefixed octet strings, and I couldn't find any charset validation of values anywhere in NIOHPACK.

Where it bites is `HTTP2ToHTTP1Codec`. Serializing `name: value CRLF` with an embedded CR or LF terminates the field early and injects header lines into the HTTP/1.1 message downstream. Same sink as the `:path` validation added for GHSA-4px2-pw77-vc85, just reached through a regular field instead.

To be clear about severity: I don't think this is exploitable on a default pipeline and I don't have a PoC. NIOHTTP1's `NIOHTTPRequestHeadersValidator` rejects 0x00...0x1F and 0x7F on write, and the pipeline helpers insert it by default (`enableOutboundHeaderValidation` defaults to true). So this is hardening.

I still think it's worth doing here. It's a MUST that NIOHTTP2 is in the right place to enforce; `HTTPEncoder` itself doesn't validate, so a hand-built pipeline has no check at all; and anyone reading `HTTPRequestHead.headers` and passing values on elsewhere is inheriting a check they'd probably assume the HTTP/2 layer had already done.

## Modifications

`isValidFieldValue(_:)` rejects 0x00, 0x0A and 0x0D, and `legalHeaderField(value:)` now calls it for regular headers, throwing a new `NIOHTTP2Errors.invalidHTTP2HeaderFieldValue`. The state machine already turns a throw here into the right stream/connection error, so nothing else needed changing.

SP, HTAB and obs-text stay legal (RFC 9110 § 5.5). It sits behind the existing `headerBlockValidation` flag. The new error type mirrors `ForbiddenHeaderField`'s storage pattern and is additive public API.

Two things I left out on purpose, both easy to add if you'd rather have them:

- § 8.2.1 also forbids a value starting or ending with SP/HTAB. Not smuggling-relevant, and rejecting it seemed likely to break peers sending harmless padding.
- No check in the codec itself. `HTTPRequestHead(http2HeaderBlock:isEndStream:)` validates pseudo-headers inline via `peekPseudoHeader`, but `appendRegularHeaders(from:)` just copies. I didn't add one there because it's per-request and you have allocation thresholds. It does leave the two asymmetric if `headerBlockValidation` is `.disabled`.

## Result

Regular field values containing NUL, CR or LF are treated as malformed. Well-formed values are unaffected.

## Testing

Added to `HTTP2FramePayloadToHTTP1CodecCRLFTests.swift`, the regression file from GHSA-4px2-pw77-vc85, which currently has no coverage for regular field values.

- `isValidFieldValue` against CRLF, lone CR, lone LF, NUL, a full smuggling payload, a TE desync payload
- values that must still pass: `with spaces inside`, `with\ttab`, `text/html; charset=utf-8`, bearer tokens, non-ASCII
- request, response and trailers validation

Built and ran on Linux, Swift 6.1 aarch64. `swift test -Xswiftc -warnings-as-errors` is clean with no new warnings, the new tests pass, and `swift format lint --strict Sources/ Tests/` is clean.


glbrntt (approved):

abhinav9-coder:
Can you please file a CVE for this , I am happy to accept for my contribution

PR #555, opened 2026-07-17
NIOHTTP2: reject SP and control characters in pseudo-header values

abhinav9-coder:
Stacks on #554 — the diff will show that commit too until it merges.

## Motivation

`isValidPseudoHeaderValue` rejects CR, LF and NUL, which is narrower than the HTTP/1.1 request line actually needs.

`:path` ends up as the request-target in:

```
METHOD SP request-target SP HTTP-version CRLF
```

so `:path = "/a HTTP/1.1"` serializes to `GET /a HTTP/1.1 HTTP/1.1`. Whether a recipient reads the first or the last SP-delimited token as the version decides what the request-target actually is, which is the usual parser-differential setup. RFC 9112 § 3.2 wants SP in a request-target percent-encoded, and bare HTAB and the other CTLs aren't valid there either.

None of the HTTP/2 pseudo-headers have a grammar that allows SP or a CTL: `:method` and `:protocol` are tokens, `:scheme` is a URI scheme, `:authority` is host[:port], `:path` is a request-target, `:status` is 3DIGIT. So rejecting them doesn't reject anything well-formed.

Same severity caveat as #554: NIOHTTP1's `uriOnlyContainsAllowedCharacters` already rejects SP in a request-target by default, so this is hardening rather than something exploitable on a stock pipeline.

## Modifications

`isValidPseudoHeaderValue` goes from `{0x00, 0x0A, 0x0D}` to `<= 0x20 || == 0x7F`, so the whole CTL range plus SP and DEL.

Bytes >= 0x80 stay allowed on purpose. They aren't delimiters in a request line or header block, and rejecting them would break anyone sending unencoded UTF-8 in `:path`.

### One behaviour change worth flagging

`validateRequestBlock` runs on send as well as receive, so a client passing a URI with a raw space in it will now throw where it didn't before. I'd argue that's the right outcome, since NIOHTTP1's validator already rejects that URI by default and this makes the two layers agree, but it is a behaviour change and it's the bit most likely to need discussion. That's why I split it out of #554 instead of bundling them.

## Result

Pseudo-header values containing SP or any control character are rejected.

## Testing

Added to `HTTP2FramePayloadToHTTP1CodecCRLFTests.swift`:

- `:path` with SP (`/a HTTP/1.1`), HTAB, VT, FF, DEL, SOH
- SP in `:authority`, `:method`, `:scheme`
- covered at the predicate, request-validation and server-codec levels

The existing positive cases in that file still pass untouched — `/`, `/foo/bar`, `/foo?q=1&r=2`, `/foo#fragment`, `*`, `/foo%20bar`, `200` — which is the check that this doesn't reject anything legitimate.

Built and ran on Linux, Swift 6.1 aarch64. `swift test -Xswiftc -warnings-as-errors` is clean with no new warnings, the new tests pass, and `swift format lint --strict Sources/ Tests/` is clean.


glbrntt (approved):
Thanks, this is a nice correctness improvement.

Sources/NIOHTTP2/HTTP2ToHTTP1Codec.swift
```diff
@@ -796,7 +796,52 @@ extension HPACKHeaders {
...
+        // downstream HTTP/1.1 message.
+        //
+        // Unlike pseudo-header values, SP, HTAB, and obs-text (0x80-0xFF) are permitted here,
+        // because RFC 9110 § 5.5 admits all of them inside a field value.
+        //
+        // Note: RFC 9113 § 8.2.1 additionally forbids a field value from *starting or ending*
+        // with SP or HTAB. That constraint is not enforced here, as it does not enable
+        // smuggling and rejecting it risks breaking peers that emit sloppy-but-harmless
+        // padding.
+        !value.utf8.contains(where: { $0 == 0x00 || $0 == 0x0A || $0 == 0x0D })
```
glbrntt: Hmm merge dupe'd this. Removing the dupe.

```suggestion
```

abhinav9-coder:
Can you  please file a CVE for this , I am happy to accept for my research

